PR #64, opened 2025-04-16
chore(nimbus): change `Alert` roots

jaikamat:
## Summary

While migrating, I encountered `"<div/p> cannot be descendants of <p>"` errors. So, best to use a `div` as the container for possibly highly styled children. This should cause no regression.

In the end, you were right @misama-ct 😆 🚀


tylermorrisford (approved):

misama-ct (approved):
> In the end, you were right @misama-ct 😆 🚀

😏

PR #73, opened 2025-04-23
docs(alert): design guidelines

stephsprinkle:
#### Summary

- Created `Alert` guidelines
- Cleaned up `Button` guidelines (just a little bit)

### DIFF
--- a/apps/docs/src/components/document-renderer/components/live-code-editor/live-code-editor.tsx
+++ b/apps/docs/src/components/document-renderer/components/live-code-editor/live-code-editor.tsx
@@ -36,7 +36,7 @@ export const LiveCodeEditor = (props: LiveCodeEditorProps) => {
   }, [props.children]);
 
   return (
-    <Box fontFamily="body" mb="1600">
+    <Box fontFamily="body" mb="1600" whiteSpace="pre-wrap">
       <Flex>
         <Box
           borderBottom="solid-25"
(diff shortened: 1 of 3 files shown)


apps/docs/src/components/document-renderer/components/live-code-editor/live-code-editor.tsx
```diff
@@ -36,7 +36,7 @@ export const LiveCodeEditor = (props: LiveCodeEditorProps) => {
   }, [props.children]);
 
   return (
-    <Box fontFamily="body" mb="1600">
+    <Box fontFamily="body" mb="1600" whiteSpace="pre-wrap">
```
stephsprinkle, line 39: This makes it so that the content in the preview isn't busting up out of the container.

misama-ct: The issue is caused by a `<pre>` tag and it's styles, wrapping the whole construct (preview/editor), but you can't just fix it with `whiteSpace="pre-wrap"` for both the editor and the preview.

You will have to set `whiteSpace="normal"` on the `Box` that is wrapping the preview and `whiteSpace="pre-wrap"` on the Box wrapping the editor. Only this way components will display as intended in the preview-tab and long lines can be wrapped in the editor-tab.

misama-ct (approved):
Approved (if you do the one tiny change I requested) :)

ddouglasz (approved):
Thank you! 🙌🏽

apps/docs/src/components/document-renderer/components/live-code-editor/live-code-editor.tsx
stephsprinkle: I moved the `whiteSpace` prop to the `Box` wrapping the `LivePreview` component. FWIW, I didn't see any issues with having it on the main `Box`

PR #81, opened 2025-04-24
chore(docs): create `IconButton` docs

jaikamat:
## Summary

This PR creates the IconButton docs. Reviewers, please pull this down and run locally it to view it in all its glory


packages/nimbus/src/components/icon-button/icon-button.mdx
```diff
@@ -14,85 +14,644 @@ figmaLink: >-
...
+| ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------- |
+| <Icons.Check color="hsla(239, 58%, 56%, 1)" width="24px" height="24px" /><Icons.CheckCircle color="hsla(239, 58%, 56%, 1)" width="24px" height="24px" />                                                                                                                                                                            | Use a button with “Save” instead of an icon button.                     |
```
jaikamat, line 221: Will eventually add <Box> here before merge, just ended the day in this state. It works as is

packages/nimbus/src/components/icon-button/icon-button.mdx
```diff
@@ -14,85 +14,644 @@ figmaLink: >-
...
-        </Stack>
-      ))}
+    <Stack width="185px">
+      <TooltipTrigger delay={0} closeDelay={0} isOpen>
+        <IconButton variant="ghost" tone="primary" alignSelf="end">
+          <Icons.ViewColumn />
+        </IconButton>
+        <Tooltip placement="top" offset={-37} crossOffset={-100}>
+          Table settings
+        </Tooltip>
```
jaikamat, line 530: These tooltip examples here and below deviate a bit from the spec, but the spirit of the design is intact. `placement="left"` is buggy (tooltip appeared 20,000px at the bottom of the page) so I had to manually alter these using the offset API. I wasted enough time here trying to figure this out, so if anyone has any ideas, I'm all ears.

packages/nimbus/src/components/icon-button/icon-button.mdx
misama-ct: Since we import all icons anyway into the docs app, you might as well provide an `Icons` component to the mdx renderer (Line `82-85`):

image

Usage in the MDX file would be like this:

```
<Icon id="Check" color="use a token value" />
<Icon id="CheckCircle" color="use a token value" />
```

Oh, and if both `height` and `width` share the same value, you can use `boxSize` as shorthand.

packages/nimbus/src/components/icon-button/icon-button.mdx
```diff
@@ -14,85 +14,655 @@ figmaLink: >-
...
+
+These specifications define the standard styles and variations for icon buttons
+within our design system, covering different visual treatments for diverse use
+cases.
+
+### Resources
+
+Deep dive on details and access design library.
+
+[TECH RELATED LINK](#) [TECH RELATED LINK](#)
```
ddouglasz, line 34: Going by a feedback I got, we may not need these empty links for now
```suggestion
```

ddouglasz (approved):
🙌🏽

packages/nimbus/src/components/icon-button/icon-button.mdx
jaikamat: Worked like a charm. see 5a03613468a8e53044db0c619124ae0271b922dd

packages/nimbus/src/components/icon-button/icon-button.mdx
jaikamat: Thanks for this, Douglas! Resolved in ee9f6db442f326a9f8846c3d75905c1b2c28cc38

misama-ct (approved):

packages/nimbus/src/components/icon-button/icon-button.mdx
jaikamat: Sounds good! https://commercetools.atlassian.net/browse/FCT-1506

PR #82, opened 2025-04-25
chore(docs): create `Card` docs

jaikamat:
## Summary

This PR creates the documentation for `Card`.

Do note that I've not added the illustration images from the spec yet as they were screenshots. I'll need @FilPob to export the originals

## TODO:
- [ ] Use real image assets


packages/nimbus/src/components/card/card.mdx
```diff
@@ -10,201 +10,394 @@ menu:
...
+          </Box>
+          <Stack direction="horizontal" gap="400" alignItems="center">
+            <Stack gap="200">
+              <Flex gap="100" alignItems="center">
+                <Text fontWeight="bold" color="primary.11">
+                  See your products
+                </Text>
+                <Box color="primary.11" boxSize="600">
+                  <Icons.ArrowForward width="full" height="full" />
+                </Box>
```
jaikamat, line 271: @ByronDWall @misama-ct I really do feel we need to discuss how we are asking consumers to style Icons in Nimbus. Wrapping a box, applying styles there assuming the Icon inherits said styling, and then having to match height and width internally is clumsy IMO, unless I am still missing something. I feel icons should take these properties directly 🤔

packages/nimbus/src/components/card/card.mdx
```diff
@@ -10,201 +10,394 @@ menu:
...
+    </Stack>
   </Stack>
-)
+);
 ```
 
-### Flexible Content
+## Properties
+
+<PropTable id="Card" />
```
jaikamat, line 367: @misama-ct This isn't pulling in the props. Related to the current work you'd mentioned on Friday?

packages/nimbus/src/components/card/card.mdx
```diff
@@ -10,201 +10,394 @@ menu:
...
+### Actions
+
+Actions within a card should be limited and focused, prioritizing the primary action related to the card's content to avoid overwhelming the user. Ensure actions are clearly labeled and visually prominent to facilitate quick interaction.
 
-Cards support different background styles to help with visual hierarchy. Options include `default` (white/high contrast) and `muted` (subtle background color).
+> [!TIP]\
+> **Do**
+>
+> - If interaction is needed, provide obvious and limited actions.
+> - Users often assume cards are clickable, especially when they contain visual cues that suggest interactivity, like buttons, links, or hover effects.
```
ddouglasz, line 204:
```suggestion
> - Users often assume cards are clickable, especially when they contain visual cues that suggest interactivity, like buttons, links, or hover effects. However, this assumption isn't universal and can vary depending on the context and user experience.
```

ddouglasz (approved):
Only one tiny change and the rest, LGTM!

packages/nimbus/src/components/card/card.mdx
ByronDWall: many design systems have an `Icon` component, where you pass the name of the icon you want to display as a prop. eg [what this article suggests](https://dev.to/seanyasno/handling-icons-in-react-best-practices-22c5)  We will probably end up implementing either something like that, or some other component/icon helper.

Im not as clear on how much of a priority that is, however

ByronDWall (approved):
Looks good aside from issues already mentioned by other commenters, but nothing here is large enough to hold back an approval

packages/nimbus/src/components/card/card.mdx
```diff
@@ -10,201 +10,394 @@ menu:
...
+    </Stack>
   </Stack>
-)
+);
 ```
 
-### Flexible Content
+## Properties
+
+<PropTables ids="CardRoot, CardHeader, CardContent" />
```
jaikamat, line 367: @ddouglasz then I had to use it here

PR #85, opened 2025-04-28
docs(select): enhance select input documentation with usage examples …

misama-ct:
The updated docs for the `Select` Input.

[Figma-Spec](https://www.figma.com/design/gHbAJGfcrCv7f2bgzUQgHq/NIMBUS-Guidelines?node-id=1699-55466&m=dev)


packages/nimbus/src/components/select/select.mdx
```diff
@@ -12,159 +12,443 @@ menu:
...
+  </Select.Root>  
 )
 ```
 
-### Sizes
+### Fixed width
 
-Available sizes.
+#### Solid-styled
+Max width when not in a page wide form should be 268px. If the solid-styled
```
ByronDWall, line 74: if we are going to specify a px value, should we also list the token name?  Also, `268px` is not a defined `size` token, the closest are `size.6400`(256px) and `size.7200`(288px) - we could either use one of these or add a `size.6700`(268px) to the token set.

```suggestion
Max width when not in a page wide form should be 268px, tokens.size.<number>. If the solid-styled
```

packages/nimbus/src/components/select/select.mdx
```diff
@@ -12,159 +12,443 @@ menu:
...
+  </Select.Root>  
+)
 ```
 
-### Variants
+#### Ghost-styled
+Hug the select-input field, this styling needs closer placement to the drop
+down icon than a solid-styled version. The popover width has two options with
+this placement based on content. Short content should match the hug width of
+the field. Longer content should have a max width of 268px.
```
ByronDWall, line 95: same thing about listing the token and having a token value as mentioned above

packages/nimbus/src/components/select/select.mdx
```diff
@@ -12,159 +12,443 @@ menu:
...
+
+### Usage
+
+Select inputs are common and users are familiar with their uses. Keeping
+consistent with the guidelines will help our products intuitive and easy to
+understand, even through complicated areas. 
+
+### Solid-styled widths
+
+These placements are more regular, they have either a maximum width of 268px
```
ByronDWall, line 237: same thing about also mentioning the token and that `268px` is not a tokenized value

packages/nimbus/src/components/select/select.mdx
```diff
@@ -12,159 +12,443 @@ menu:
...
+> **Don't**
+> - Do not have different widths for select inputs.
+> - Do not change the width of the field to match the content once selected.
+
+![](/images/select/solid-styled-widths-dont.png)
+
+### Ghost-styled widths
+There are a few differences that ghost-styled select inputs have that aren't
+needed with solid-style placements. Ghost-styled fields should hug width, with
+popovers having a min width matching the field, and max width being 268px unless
```
ByronDWall, line 259: token/value/etc

packages/nimbus/src/components/select/select.mdx
```diff
@@ -12,159 +12,443 @@ menu:
...
+      <Select.Option>Oranges</Select.Option>
+      <Select.Option>Cherries</Select.Option>
+    </Select.Options>
+  </Select.Root>  
+)
 ```
 
-### Variants
+#### Ghost-styled
+Hug the select-input field, this styling needs closer placement to the drop
```
ByronDWall, line 92: as much as inputs need a Hug, what does "Hug the select-input field" mean in this context?

```suggestion
Hugs the select-input field, this styling needs closer placement to the drop
```

ByronDWall (approved):
A few small nits, but nothing that would stop me from approving overall

packages/nimbus/src/components/select/select.mdx
misama-ct: I don't know where the 268px value came from, the component itself does not have this limitation. The select is as wide as the label requires it to be or the surrounding space allows it to be.

packages/nimbus/src/components/select/select.mdx
misama-ct: I don't know.

packages/nimbus/src/components/select/select.mdx
misama-ct: I mean I do know where it comes from, but those specs were added to the figma file after the implementation was done and I don't know the reason for that.

packages/nimbus/src/components/select/select.mdx
```diff
@@ -12,159 +12,443 @@ menu:
...
+TODO: Image
+
+### Accessibility Standards
 
-<PropTables
-  root="Select"
-  ids="SelectRoot, SelectOptions, SelectOptionGroup, SelectOption"
-/>
+#### 1. Perceivable:
+- **1.3.1 Info and Relationships (Level A):
```
jaikamat, line 419: needs the closing `**`

packages/nimbus/src/components/select/select.mdx
```diff
@@ -12,159 +12,443 @@ menu:
...
+> [!CAUTION]\
+> **Don't**
+> - Don’t show both helper text and information text simultaneously. 
+
+![](/images/select/help-and-error-text-inclusion-dont.png)
+
+
+## Specs
+
+<PropTable id="Select" />
```
jaikamat, line 405: Specs not working again, but not gonna let it block this

jaikamat (approved):
Other than the existing comments, I didn't notice anything out of order! ✨

ddouglasz (approved):

packages/nimbus/src/components/select/select.mdx
misama-ct: [Fixed](https://github.com/commercetools/nimbus/pull/85/commits/4f5507b8fd502f84df4b86a91f9aff28aa85b424)

packages/nimbus/src/components/select/select.mdx
```diff
@@ -12,159 +12,443 @@ menu:
...
+## Overview
+Select inputs have two main appearances, two sizes, and several states that will
+respond to validation status. **Currently, this is a single value selector,
+this will be reworked to bring in multi-select in the future.**
 
-## Basic Usage
+
+### Resources
+
+ [Figma Library](https://www.figma.com/design/AvtPX6g7OGGCRvNlatGOIY/NIMBUS-design-system?node-id=2304-22847&m=dev)
```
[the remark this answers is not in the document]
misama-ct: [Fixed](https://github.com/commercetools/nimbus/pull/85/commits/9c163af485eddac017ad059457d900e6a3c54723)

PR #87, opened 2025-04-28
FCT-1450- Publishing again...for fun let's stop tracking the phases.

valoriecarli:
After further review, the following adjustments have been made to the `Nimbus` package:

- Nimbus' `peerDependencies` align with those in the `vite.config` external dependencies array.
- Dependencies that were declared in peer/external were removed & or relocated from `dependencies`

⬇️ Note that that though specified, certain peerDeps/externals were not required for the mc-fe to work. These will be tracked in [1502](https://commercetools.atlassian.net/browse/FCT-1502) for now for future review.

### DIFF
--- a/packages/nimbus/package.json
+++ b/packages/nimbus/package.json
@@ -35,7 +35,6 @@
     }
   },
   "dependencies": {
-    "@react-aria/optimize-locales-plugin": "^1.1.4",
     "react-aria": "catalog:react",
     "react-aria-components": "catalog:react",
     "react-stately": "catalog:react"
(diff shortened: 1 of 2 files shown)


packages/nimbus/package.json
```diff
@@ -34,8 +34,18 @@
       ]
     }
   },
+  "dependencies": {
+    "@react-aria/optimize-locales-plugin": "^1.1.4",
```
ByronDWall, line 38: this can probably be a dev dependency as well, since it's only used during build time w/vite

ByronDWall (approved):
Looks real good, saw it install with my own eyes, very approved

misama-ct (approved):

ddouglasz (approved):
🙌🏽

PR #90, opened 2025-04-30
chore(gitignore): add Storybook log and static files to .gitignore

misama-ct:
The output folder for the static-build of storybook was not in the .gitignore file yet.


tylermorrisford (approved):
[standing ovation]

PR #95, opened 2025-05-05
chore(dependabot): add configuration for dependabot updates

misama-ct:
This PR is an attempt to provide a bit of context to dependabot, as it currently does not seem to grasp how dependencies are managed (monorepo, pnpm, workspaces, catalogs) and thus provides [unmergeable PR's](https://github.com/commercetools/nimbus/pull/92) (a change in the lock-file is not enough, the exact versions are specified in the `pnpm-workspace.yaml` file).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+version: 2
+updates:
+  # Configuration for pnpm dependencies in the monorepo root
+  - package-ecosystem: "npm" # "npm" = for pnpm, yarn, and npm projects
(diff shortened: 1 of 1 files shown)


ddouglasz (approved):

misama-ct:
I also found [this link](https://github.com/dependabot/dependabot-core/issues/11953), claiming something is currently broken (no updates to the `pnpm-workspace.yaml` file), but in any case, an explicit config can't hurt.

tylermorrisford (approved):

PR #102, opened 2025-05-06
FCT-1469: Nimbus - TextInput - Resolve layouting issues with style-props

misama-ct:
After ditching the idea of changing the`onChange` handler back to propagating a regular `ChangeEvent` instead of the value itself, the only functional change is ditching react-arias `TextField` component in favor of the `useTextField` hook.

I did this, cause the `TextField` component was adding an additional, unnecessary div around the input which caused some issues when attempting to apply additional style-props to the `TextInput`.

The component is now wiring only the needed props (`inputProps`) from the hook to `Input`.

### DIFF
--- a/packages/nimbus/src/components/form-field/form-field.stories.tsx
+++ b/packages/nimbus/src/components/form-field/form-field.stories.tsx
@@ -246,7 +246,7 @@ export const Required: Story = {
     });
 
     await step("Input has required attribute", async () => {
-      await expect(input).toHaveAttribute("required");
+      await expect(input).toHaveAttribute("aria-required");
     });
   },
 };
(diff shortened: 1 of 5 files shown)


ByronDWall:
> I did this, cause the TextField component was adding an additional, unnecessary div around the input which caused some issues when attempting to apply additional style-props to the TextInput.

I think this is because the `TextField` is supposed to be a wrapper around an `Input`, `Label`, etc, etc.  React-aria does export an `Input` [component](https://github.com/adobe/react-spectrum/blob/main/packages/react-aria-components/src/Input.tsx) that might be more suitable for use here, since the `TextField` sets up a lot of contexts for the non-input slots that aren't really intended for use in this component

misama-ct:
> > I did this, cause the TextField component was adding an additional, unnecessary div around the input which caused some issues when attempting to apply additional style-props to the TextInput.
> 
> I think this is because the `TextField` is supposed to be a wrapper around an `Input`, `Label`, etc, etc. React-aria does export an `Input` [component](https://github.com/adobe/react-spectrum/blob/main/packages/react-aria-components/src/Input.tsx) that might be more suitable for use here, since the `TextField` sets up a lot of contexts for the non-input slots that aren't really intended for use in this component

Actually `TextInput` is using `Input`, I even tried using it without the hook first, but `Input` does not have any of the `is*` props by default and `TextInput` would also not correctly work with `FormField` anymore.

ByronDWall (approved):
lgtm, thanks!

ddouglasz (approved):

PR #139, opened 2025-05-14
chore(nimbus): fix duplicate onChange event handling in text input

jaikamat:
## Summary

Our `TextInput` component's `onChange` `value` would seemingly switch between an event and a string value. In reality, it was called twice.

It causes validation errors on the consumer end because `.trim()` is not a valid function for the event's type of `object`

image

### DIFF
--- a/packages/nimbus/src/components/text-input/text-input.tsx
+++ b/packages/nimbus/src/components/text-input/text-input.tsx
@@ -32,12 +32,7 @@ export const TextInput = forwardRef<HTMLInputElement, TextInputProps>(
     const { inputProps } = useTextField(otherProps, ref);
 
     return (
-      <TextInputRootSlot
-        {...recipeProps}
-        {...styleProps}
-        {...otherProps}
-        asChild
-      >
+      <TextInputRootSlot {...recipeProps} {...styleProps} asChild>
         <Input ref={ref} {...inputProps} />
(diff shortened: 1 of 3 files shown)


packages/nimbus/src/components/text-input/text-input.tsx
```diff
@@ -32,12 +32,7 @@ export const TextInput = forwardRef<HTMLInputElement, TextInputProps>(
     const { inputProps } = useTextField(otherProps, ref);
 
     return (
-      <TextInputRootSlot
-        {...recipeProps}
-        {...styleProps}
-        {...otherProps}
```
jaikamat, line 38: Do we even need `otherProps` as it's processed into `inputProps` by `useTextField`?

I opted to remove it. We can also include it via something like
```
const mergedProps = { ...otherProps, ...inputProps };
```
, but order matters in this case and it feels like a brittle gotcha moment.

misama-ct (approved):

packages/nimbus/src/components/text-input/text-input.tsx
ByronDWall: I dont believe we need `otherProps` here, but if we run into issues later on we can come back to that assumption

ByronDWall (approved):

PR #142, opened 2025-05-15
fix(IconButton): fix issue where iconButton was causing unexpected behavior in Select component

ByronDWall:
fix(icon button): use merged props from `useContextProps`, but do not use resulting ref

Fixes an issue where the `IconButton` component was causing the popover to display in an unexpected position when the `clear` button is present in the `Select` component
image
image (1)

Issue is fixed by not passing the `ref` generated by `useContextProps` to the `Button` component, and instead allowing the `useButton` hook to generate the appropriate `ref` in the underlying `Button` component.  This can be verified by looking at the [formfield storybook for this branch](https://nimbus-storybook-git-bw-fix-icon-button-issue-commercetools.vercel.app/?path=/story/components-formfield--using-a-select-input).
Screenshot 2025-05-15 at 8 15 11 AM


misama-ct (approved):
Thank you!

jaikamat (approved):

stephsprinkle (approved):

PR #144, opened 2025-05-16
Fct 1335 combobox

ByronDWall:
This pull request introduces significant changes, primarily focusing on the addition of a new `ComboBox` component, updates to existing components, and minor workflow adjustments. The most important changes include the implementation of the `ComboBox` component with its recipes, slots, types, and stories, as well as updates to other components for consistency.

### New `ComboBox` Component:

* **Component Implementation**: Added the `ComboBox` component, including its root, options, option group, and individual option sub-components. These are exported as part of the `ComboBox` object. (`packages/nimbus/src/components/combobox/combobox.tsx`)
* **Styling and Recipes**: Defined the `comboBoxSlotRecipe` for consistent styling, including base styles, size variants, and visual style variants. (`packages/nimbus/src/components/combobox/combobox.recipe.tsx`)
* **Slot Definitions**: Introduced slot components for `ComboBox` elements, such as root, value, button group, options, and option group, using Chakra's slot recipe context. (`packages/nimbus/src/components/combobox/combobox.slots.tsx`)
* **Type Definitions**: Added comprehensive TypeScript types for all `ComboBox` components and their props, ensuring type safety and clarity. (`packages/nimbus/src/components/combobox/combobox.types.ts`)
* **Storybook Stories**: Created Storybook stories for the `ComboBox` component, demonstrating its usage with multiple options and configurations. (`packages/nimbus/src/components/combobox/combobox.stories.tsx`)

### Documentation and Examples:

* **MDX Documentation**: Added an MDX file documenting the `ComboBox` component, including usage examples for sizes, variants, and colors. (`packages/nimbus/src/components/combobox/combobox.mdx`)

### Updates to Existing Components:

* **Button Component**: Updated the `ButtonRecipeProps` interface to use `typeof buttonRecipe` and adjusted the `ButtonRootProps` to be an interface instead of a type. (`packages/nimbus/src/components/button/button.slots.tsx`)
* **Checkbox Component**: Changed the label color for invalid states from `neutral.11` to `black`. (`packages/nimbus/src/components/checkbox/checkbox.recipe.tsx`)

### Workflow Adjustments:

* **GitHub Actions**: Renamed the workflow job from `build-and-deploy` to `build-and-test` for better clarity. (`.github/workflows/build-and-test.yml`)


package.json
```diff
@@ -32,7 +32,11 @@
       "rollup": "catalog:tooling",
       "@babel/runtime": "catalog:tooling",
       "prismjs": "^1.30.0",
-      "typescript": "catalog:tooling"
+      "typescript": "catalog:tooling",
+      "react-aria": "catalog:react",
+      "react-aria-components": "catalog:react",
+      "react-stately": "catalog:react",
+      "@react-aria/interactions": "catalog:react"
```
ByronDWall, line 39: This is here to insure that we dont have any lurking dependency mismatches in our react-aria related dependency graph, since react-aria packages have dependencies on other react-aria packages (eg `react-stately` is a dep of both `react-aria` and `react-aria-components`

packages/nimbus/src/components/combobox/combobox.mdx
```diff
@@ -0,0 +1,93 @@
+---
```
ByronDWall, line 1: docs are coming soon, the combobox is in a state where reviews would be helpful without them though.  I can either add them in this PR or a separate PR, whichever people prefer.

packages/nimbus/src/components/combobox/combobox.recipe.tsx
```diff
@@ -0,0 +1,266 @@
...
+        overflow: "hidden",
+        textOverflow: "ellipsis",
+        fontSize: "350",
+        fontWeight: "400",
+        lineHeight: "400",
+        _placeholder: { opacity: 0.5 },
+      },
+    },
+    options: {
+      ...selectSlotRecipe.base?.options,
```
ByronDWall, line 103: I like the pattern of reusing recipes when we can, but also could see where doing so could end up causing a regression if something changes in the component the recipe is from but should not change in this component.  So I could go either way here.

packages/nimbus/src/components/combobox/combobox.stories.tsx
```diff
@@ -0,0 +1,2057 @@
...
+};
+
+/**
+ * Complex Options with Descriptions Example
+ * Demonstrates:
+ * - handling item objects that do not have an id or name key
+ * - options with rich content including descriptions for both single and multi-select
+ * - complex options with non-standard keys inside sections
+ */
+export const ComplexOptions: Story = {
```
ByronDWall, line 1213: So everyone's aware, I am in the process of adding tests for the rest of the stories.  I am starting the review process without them complete because there is _a lot_ going on in this component, and I want to get everyone's feedback on how it's implemented and what does or does not make sense.

So anything above this point is fair game for 'why didnt you test x', but below here is still a work in progress.

jaikamat:
https://github.com/user-attachments/assets/77ca043e-e2bc-482f-9f1f-c8da70c08bfd

Not sure if this is a bug or expected behavior:
1. The popover opens for me on first focus of the input area
2. Blur the input
3. Subsequent focusing does not display the popover, I'm required to press the icon button

(This also happens with keyboard navigation)

packages/nimbus/src/components/combobox/combobox.types.ts
```diff
@@ -0,0 +1,207 @@
...
+  onSubmitCustomValue?: (value: string) => void;
+  onOpenChange?: RaComboBoxProps<T>["onOpenChange"];
+  autoFocus?: boolean;
+  isLoading?: boolean;
+  isDisabled?: boolean;
+  isReadOnly?: boolean;
+  isRequired?: boolean;
+  isInvalid?: boolean;
+  // name of key in items that is the unique id
+  itemID?: string;
```
misama-ct, line 52: ocd-nit:

```suggestion
  itemId?: string;
```

packages/nimbus/src/components/combobox/components/combobox.multi-select-root.tsx
```diff
@@ -0,0 +1,347 @@
...
+          isDisabled={isDisabled}
+          isReadOnly={isReadOnly}
+          isLoading={isLoading}
+        />
+      </div>
+
+      <ComboBoxPopoverSlot // Set css variable for the trigger width, fall back to 288px if state.triggerWidth is undefined
+        style={
+          {
+            "--trigger-width": triggerWidth ?? themeTokens.size[7200],
```
misama-ct, line 300: You can avoid importing the raw tokens, by using [this approach](https://v2.chakra-ui.com/docs/styled-system/css-variables#creating-scoped-theme-aware-css-variables) to set the css-variable

misama-ct (approved):
- While keyboard-navigating: the focus outline is not the default outline, the current neutral outline clashes with the header-border due to missing safe-space and the background does not seem to change to the blue bg
image
- noticed in the _AsyncSelect_ example: once focused, the placeholder truncation doesn't work anymore, and after typing 2 characters, I would expect it to open and provide me the 3 fake-examples for selection (but it doesn't) 
image
- _LargeDataset_ is kinda unusable, is this an actual case? I feel like with async options there is no need to render 1000 items to the DOM. But even if, we can add some virtualisation later.

packages/nimbus/src/components/checkbox/checkbox.recipe.tsx
```diff
@@ -26,7 +26,7 @@ export const checkboxSlotRecipe = defineSlotRecipe({
     label: {
       flexShrink: 0,
       userSelect: "none",
-      color: "neutral.11",
+      color: "black",
```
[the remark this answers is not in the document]
ByronDWall: good point, and good catch

packages/nimbus/src/components/button/button.types.ts
```diff
@@ -3,12 +3,15 @@ import type { AriaButtonProps } from "react-aria";
...
+  Omit<ButtonRootProps, keyof AriaButtonProps | "slot"> & {
     [key: `data-${string}`]: unknown;
   };
 
 export interface ButtonProps extends FunctionalButtonProps {
   // TODO: evaluate if we should require setting a tone
   // tone: FunctionalButtonProps["tone"];
+  // we need 'null' as a valid slot value for use with components from react-aria-components,
+  // in react-aria slots "An explicit null value indicates that the local props completely override all props received from a parent."
+  slot?: string | null | undefined;
```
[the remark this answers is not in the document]
ByronDWall: Oh interesting, the buttons that toggle the combobox dropdown in the button groups don't have slots declared, and they don't throw. Maybe certain contexts enforce restrictions on the number of buttons or non-slotted children in the dom structure?

I'd be interested to see which contexts make buttons throw without a slot declared

ByronDWall:
> While keyboard-navigating: the focus outline is not the default outline, the current neutral outline clashes with the header-border due to missing safe-space and the background does not seem to change to the blue bg

@misama-ct Good catch on the outline, I'll change that.  According to the figma, the background color is only meant to change on hover in the multiselect options, so that's intentional. And you're right about the padding between the options and input, I think I over-optimized for the option groups when updating styles yesterday, I'll update that as well.

> noticed in the AsyncSelect example: once focused, the placeholder truncation doesn't work anymore, and after typing 2 characters, I would expect it to open and provide me the 3 fake-examples for selection (but it doesn't)

Yeah, this is [a known bug](https://github.com/adobe/react-spectrum/issues/5234) in react-aria, and it is annoying.  There are some sub-optimal workarounds, and it looks like better support [is coming](https://github.com/adobe/react-spectrum/pull/7938), but it's something to think about.  If you type a second character the results should display.

> LargeDataset is kinda unusable, is this an actual case? I feel like with async options there is no need to render 1000 items to the DOM. But even if, we can add some virtualisation later.

yeah. I noticed that too.  It's at least a very rare case - I also think that virtualization  is the answer, and it should be explored in a future PR

packages/nimbus/src/components/button/button.types.ts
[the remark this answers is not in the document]
misama-ct: Try it [out here](https://github.com/commercetools/nimbus/pull/197/files#diff-d35795b0a5ca8bdbbc14e49baebb665b19cda66ad382479541e8a6cdc8151bc0R34). I also added the slot-ability to Text, where the same happens.

packages/nimbus/src/utils/useResizeObserver.ts
```diff
@@ -0,0 +1,60 @@
+/*
```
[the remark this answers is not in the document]
ByronDWall: yes, we need this comment because I copied it direct from the react-aria repo and that is illegal unless I add the license comment to the top

packages/nimbus/src/components/combobox/components/combobox.single-select-input.tsx
```diff
@@ -0,0 +1,70 @@
...
+    onInputChange,
+    ref,
+  } = props;
+
+  // Get input props from ComboBox InputContext and merge with local props
+  const [inputProps, inputRef] = useContextProps({}, ref || null, InputContext);
+  // This component exists because it has to be a child of the root to get the combobox state
+  const state = useContext(ComboBoxStateContext);
+  // this method checks the inputValue against the textValue of every item in the collection, and if it doesnt match any items, sets a custom value
+  const handleInputKeyDown = useCallback(
```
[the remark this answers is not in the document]
ByronDWall: I'm not sure that a utils file would be more helpful in this case for a few reasons.

- We need `useCallback` to make them immutable, so they need to be a react hook. 
- The functions need access to component state as they are written currently, and passing everything in as params and typing the params would involve more boilerplate.
- Each function is unique (this `handleInputKeydown` is different than the one in the multiselect, etc), so there would not really be any reuse of the function in multiple files

That being said, I think that once we have time to optimize the component more by doing things like creating a custom context for it, etc, that might enable us to reuse functions more.  In that case, a utils file would be a good addition.

packages/nimbus/src/components/combobox/components/combobox.multi-select-root.tsx
ByronDWall: I switched to using an imported theme variable, but `sx` doesnt seem to be a valid styling method in chakra `v3`, and I left a comment as to why this needs to be set as an inline style instead of using the `css` prop in the file. in this commit: [9c8b2c9](https://github.com/commercetools/nimbus/pull/144/commits/9c8b2c980f2e9f99fe4570cd63f1724653246e70)

packages/nimbus/src/components/combobox/combobox.recipe.tsx
```diff
@@ -0,0 +1,267 @@
...
+    },
+    optionIndicator: {
+      ...checkboxSlotRecipe.base?.indicator,
+      ...checkboxSlotRecipe.variants?.size.md.indicator,
+    },
+    optionContent: {
+      ...checkboxSlotRecipe.base?.label,
+      ...checkboxSlotRecipe.variants?.size.md.label,
+      whiteSpace: "wrap",
+      width: "calc(100% - var(--nimbus-sizes-600))",
```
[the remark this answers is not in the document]
ByronDWall: fixed in [9c8b2c9](https://github.com/commercetools/nimbus/pull/144/commits/9c8b2c980f2e9f99fe4570cd63f1724653246e70)

packages/nimbus/src/components/combobox/combobox.recipe.tsx
```diff
@@ -0,0 +1,267 @@
...
+      },
+      "& [data-placeholder]": {
+        opacity: 0.5,
+      },
+    },
+    value: {
+      display: "flex",
+      focusRing: "outside",
+      alignItems: "flex-start",
+      paddingRight: "1600!",
```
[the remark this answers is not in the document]
ByronDWall: fixed in [9c8b2c9](https://github.com/commercetools/nimbus/pull/144/commits/9c8b2c980f2e9f99fe4570cd63f1724653246e70)

FilPob:
> @FilPob Something that's been bothering me is the orientation of the checkboxes in the options list, see below: image When the primary label text isn't the default font size (here, they are `textStyle="sm"`), the line-height pushes the text away from the top of the content box, causing the radio buttons to look offset as a result.
> 
> Is it possible that we can center these checkboxes about the content? Would make for a more resilient solution. I feel like consumers will inevitably run into this once they start customizing their options lists.

@jaikamat good catch! I mean I would definitely prefer to have the alignment as is right now... But I would probably be more ok with seeing it aligned to the center, than seeing it mis-aligned all over the place :D So a more resilient solution does makes sense here I guess.

packages/nimbus/src/components/combobox/combobox.recipe.tsx
```diff
@@ -0,0 +1,286 @@
...
+    },
+    optionGroup: {
+      ...selectSlotRecipe.base?.optionGroup,
+    },
+    option: {
+      ...selectSlotRecipe.base?.option,
+      whiteSpace: "wrap",
+    },
+    optionIndicator: {
+      // make sure option is aligned with the first line of text (using lh units)
```
ByronDWall, line 127: @FilPob @jaikamat I think I've found a way to make the checkbox top-aligned while still centering with the first line of text.  The only thing it depends on is that the first line of text is not styled by a child container that changes its' line-height.

packages/nimbus/src/components/combobox/components/combobox.multi-select-root.tsx
```diff
@@ -0,0 +1,347 @@
...
+import { ComboBoxButtonGroup } from "./combobox.multi-select-button-group";
+import {
+  ComboBoxPopoverSlot,
+  ComboBoxMultiSelectInputSlot,
+} from "../combobox.slots";
+import type { ComboBoxMultiSelect } from "../combobox.types";
+
+type MenuTriggerAction = "focus" | "input" | "manual" | undefined;
+
+function getLastValueInSet(set: Set<Key>) {
```
tylermorrisford, line 33: TIL 💡

FilPob (approved):

PR #151, opened 2025-05-22
FCT-1494: Nimbus - Docs - PropsTable

misama-ct:
# Summary

This PR consolidates the 2 existing components that are currently used to display react component properties (PropTable & PropTables),  into a single component (= PropsTable), adds a comprehensive [Export Stats Page](https://nimbus-documentation-git-fct-1494-nimbus-d-ff3b7f-commercetools.vercel.app/home/contribute/stats) and brings the Table and Dialog components (experimental, exclusively used by the docs) into the compound component shape.

## Displaying Component Props
We do not have to differentiate between the component type when we want to display the props in an mdx file:

- **Simple component usage** essentially stays  _as is_ except that the name changed slightly:
  - Before: ```<PropTable id="Icon"/>```
  - After:  ```<PropsTable id="Icon"/>```

- **Compound component usage** 
  - Before: ```<PropTables id="CardRoot, CardHeader, CardContent"/>```
  - After:  ```<PropsTable id="Card"/>```
  
---

If it is a compound component, the PropsTable component will display a button row which allows selecting the respective sub-component.

I already replaced all occurrences of the former components with the new ones, but for the props to show up correctly, we have to export the components individually.

## Export Stats Page

image

I added the [Export Stats Page](https://nimbus-documentation-git-fct-1494-nimbus-d-ff3b7f-commercetools.vercel.app/home/contribute/stats), so that we have a place where we can see what we actually expose to consumers and if it's properly documented. We can extend it with the component status (experimental, alpha, beta etc.) once we have them.

Meanwhile, this page can be used, to check if the individual prop-tables for a component are displayed correctly. Currently this is not the case for most compound components, cause for `react-docgen` to recognize the types the components [need to be exported individually and not as part of an object](https://github.com/commercetools/nimbus/pull/151/files#diff-fdc9cddea72402630596664133ed4f1edd85554437530797f0e1a27fc93e5c8cR19-R27).


packages/nimbus/src/components/alert/alert.tsx
```diff
@@ -16,3 +16,12 @@ export const Alert = {
...
 };
+
+// exports for internal use by react-docgen
+export {
+  AlertRoot as _AlertRoot,
+  AlertTitle as _AlertTitle,
+  AlertDescription as _AlertDescription,
+  AlertActions as _AlertActions,
+  AlertDismissButton as _AlertDismissButton,
+};
```
misama-ct, line 27: Exporting the individual parts of a compound component like this, additionally to the regular export, allows `react-docgen` to parse the types correctly.

Changing the export-name is technically not necessary, but prepending the component-names with an underscore will allow us to find them quickly once this workaround is not necessary anymore.

PR #158, opened 2025-05-29
FCT-1522: `<TimeInput/>` component

misama-ct:
# TimeInput Component Implementation

## Overview
This PR adds a new [TimeInput](https://nimbus-storybook-git-fct-1522-nimbus-compo-d59a65-commercetools.vercel.app/?path=/docs/components-timeinput--docs) component to Nimbus that allows users to enter and manipulate time values. The component is built using React Aria Components and integrates with the existing Nimbus design system.

## Changes
- Added new TimeInput component based on React Aria's TimeField
- Implemented slot-based recipe system with proper styling
- Created size variants (sm, md) with appropriate dimensions and text styles
- Added documentation and Storybook stories with test cases
- Updated package exports to include the new component

## Implementation Details

The TimeInput component:

- Provides an accessible time entry field with keyboard navigation
- Supports both controlled and uncontrolled usage patterns
- Features customizable styling through variants
- Includes proper focus, hover, and disabled states
- Works with FormField for validation and labeling
- Uses internationalization support for localized time formats

## Testing
- Added comprehensive Storybook stories with interaction tests
- Tested various states: default, controlled, invalid, disabled
- Verified keyboard navigation and screen reader accessibility

## Documentation

- No Guidelines yet, plain API examples will have to do for now


packages/nimbus/src/theme/global-css.ts
```diff
@@ -10,7 +10,11 @@ export const globalCss = defineGlobalStyles({
     "--ring-color": "rgba(66, 153, 225, 0.6)",
     "--ring-offset-shadow": "0 0 #0000",
     "--ring-shadow": "0 0 #0000",
+    // Set default chakra-variables for focus ring
     "--focus-ring-color": "colors.primary.7",
+    "--focus-ring-width": "sizes.50",
+    "--focus-ring-offset": "spacing.50",
+    "--focus-ring-style": "solid",
```
misama-ct, line 17: There is no built-in mechanism to apply to focus styles on an element that has the actual input inside (focus-within). I found out that this css-variables are used by the focusRing helper, so I added them here explicitly (for visibility) and made use of them within the recipe.

packages/nimbus/src/components/time-input/time-input.recipe.ts
```diff
@@ -0,0 +1,119 @@
...
+      boxShadow: "inset 0 0 0 var(--border-width) var(--border-color)",
+      width: "100%",
+      alignItems: "center",
+
+      "&[data-focus-within='true']": {
+        // TODO: can't use focusRing prop, find other solution (helper, util, etc.)
+        outlineWidth: "var(--focus-ring-width)",
+        outlineColor: "var(--focus-ring-color)",
+        outlineStyle: "var(--focus-ring-style)",
+        outlineOffset: "var(--focus-ring-offset)",
```
misama-ct, line 30: There is no way to apply the focusRing styles to a "focus-within"-element (an element that is a parent of the currently focused element), so I added the focus styles manually and a TODO comment for cleaning up later.

packages/nimbus/package.json
```diff
@@ -49,6 +49,7 @@
     "@chakra-ui/react": "catalog:react",
     "@commercetools/nimbus-icons": "workspace:^",
     "@commercetools/nimbus-tokens": "workspace:^",
+    "@internationalized/date": "catalog:react",
```
misama-ct, line 52: This package provides utils and types for handling dates and times and is used by react-aria (and even some chakra-components) under the hood.

Currently it is only used in the story file, but we'll have to think about how we expose the included functions to consumers later.

(e.g. Do we require it as an additional dependency in consumer apps or are we exposing them via nimbus-package or an additional 'nimbus-utils'-package...)

packages/nimbus/src/components/time-input/time-input.recipe.ts
```diff
@@ -0,0 +1,119 @@
...
+  // Unique class name prefix for the component
+  className: "nimbus-time-input",
+  slots: ["root", "segmentGroup", "segment"],
+
+  // Base styles applied to all instances of the component
+  base: {
+    root: {
+      display: "inline-block",
+    },
+    segmentGroup: {
```
misama-ct, line 17: This segmentGroup visually mocks the `TextInput` style. It is necessary to do it this way, as technically there is not just one input, but rather 3-6 html inputs ("segments") (... for hour, minute, second, millisecond, dayTime, timezone).

packages/nimbus/src/components/time-input/time-input.stories.tsx
```diff
@@ -0,0 +1,1023 @@
...
+ * - component: references the component being documented
+ */
+const meta: Meta<typeof TimeInput> = {
+  title: "components/TimeInput",
+  component: TimeInput,
+  decorators: [
+    (Story) => (
+      <I18nProvider locale="en-US">
+        <Story />
+      </I18nProvider>
```
misama-ct, line 26: I added the `I18NProvider` as a decorator to the stories. Later, this component needs to wrap consumer apps, same as the NimbusProvider, to govern the locale app-wide.

I am not sure what would be the best integration-approach yet, ideally we integrate it into the `NimbusProvider` (so that nobody has to think about it except us) but I am not sure this is possible or has any downsides.

packages/nimbus/src/components/time-input/time-input.stories.tsx
misama-ct, line 1: This is where Cursor really shined. I would have never been able to cover all properties so quickly. While it did not work magically, with every new story I added, the next story I asked for had less flaws.

Initially I tried it with prompts like "Add stories + interaction tests for each prop", but then it generates massive amounts of code, gets lost in it, some stories work, some don't and I decided I didn't want to dive any further in that swamp and simply went for the no-thrill one-bite-after-another approach:

1. First I wanted to have a story for every property, so I prompted something like this for every property (in a new Chat, that's important!):

    > Create a story, demonstrating the `hideTimeZone` property.

2. Once I had stories for each prop, I made sure the styling was as expected
3. Once styling was done, I asked Cursor again to add interaction tests for the stories (individually again, new chat).
     > "Add an interaction test to the `hideTimeZone` [...]

ddouglasz: These tiny details and explanations are Golden! Thank you :)

apps/docs/src/components/document-renderer/components/live-code-editor/live-code-editor.tsx
```diff
@@ -13,7 +14,7 @@ const baseHooks = {
 };
 
 // funcitons & components available to the live code editor
-const scope = { ...BlehUi, ...baseHooks, Icons: { ...icons } };
+const scope = { ...BlehUi, ...baseHooks, Icons: { ...icons }, Time };
```
ddouglasz, line 17: Are we still referencing `BlehUi`? 😅

packages/nimbus/package.json
ddouglasz: If it is not too heavy, I feel like maybe it would make sense to expose it to the user as a util. That way, we may not need to tell the user that they would have to download a compatible package to use the time input. Also we can manage what we would like to expose.

ddouglasz (approved):
Thank you! 🤩

apps/docs/src/components/document-renderer/components/live-code-editor/live-code-editor.tsx
misama-ct: hahaha, @stephsprinkle  tztztz...

[Fixed](https://github.com/commercetools/nimbus/pull/158/commits/7c17d28662794906fbc1c6f4af8415253bf94405)

packages/nimbus/src/components/time-input/time-input.stories.tsx
jaikamat: I agree, the more batteries-included this is, the easier Nimbus will be to adopt. We should include it in the `NimbusProvider`.

packages/nimbus/package.json
jaikamat: Since the MC is coming from Moment.JS, will there be a lot of transformation consumers will have to do on the migration side of things when they need to create valid string values for this component? Should we take this into account?

ByronDWall: Yeah, I feel like 'date/time handling' is a larger issue than the scope of this PR, but it is something we are going to have to consider very soon.

I could see requiring it as a peer-dep, or making it available as a bundled util.  This should probably be a separate spike/task.

packages/nimbus/src/components/time-input/time-input.stories.tsx
```diff
@@ -0,0 +1,1023 @@
+import type { Meta, StoryObj } from "@storybook/react";
+import { TimeInput } from "./time-input";
+import { Button, FormField, Stack, Text } from "@/components";
+import { parseZonedDateTime, Time } from "@internationalized/date";
```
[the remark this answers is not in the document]
ByronDWall: The only reason I'd disagree with this is if we can potentially replace moment with `@internationalized/date` across all ct apps, in which case it might make sense to make it a peer dep

packages/nimbus/package.json
ByronDWall: fwiw, [this issue](https://github.com/adobe/react-spectrum/discussions/2775) explains the largest gap that I've found in the `@internationalized/date` API, which is that it does not offer a method that returns what the difference between two dates is (it will only tell you that one is older than the other)

packages/nimbus/src/components/time-input/time-input.types.ts
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Examples include: value, defaultValue, onChange, onBlur, onFocus, etc.
+ */
+type ConflictingFieldStateProps = keyof TimeFieldStateOptions<TimeValue>;
+
+/**
+ * Additional properties we want to exclude from the TimeInput component.
+ * These are either deprecated or not intended for use in this component.
+ */
+type ExcludedProps =
```
ByronDWall, line 20: I like this abstraction for excluding types

misama-ct:
> @misama-ct looks great! few points:
> 
> * the placeholder variant: uses a black font-color... Should use the grey placeholder font-style like other inputs
> * The ghost variant is missing in Storybook.

@FilPob I solved the first two issues, please have a look. The disabled state styles, as you said, are shared styles, I wouldn't want to mess with them inside this story.

FilPob (approved):

PR #159, opened 2025-05-29
chore(nimbus): add error handling for `Avatar` src

jaikamat:
## Summary

This PR adds some error handling for the Avatar's `src` attribute. We need this so that Gravatar images will 404 gracefully.

### DIFF
--- a/packages/nimbus/src/components/avatar/avatar.stories.tsx
+++ b/packages/nimbus/src/components/avatar/avatar.stories.tsx
@@ -129,7 +129,7 @@ export const ImageErrorFallback: Story = {
   args: {
     firstName: "Jane",
     lastName: "Smith",
-    src: "https://example.com/non-existent-image.jpg", // This will 404
+    src: "https://www.gravatar.com/avatar/thisWill404?s=200&d=404", // d=404 will return a 404 if the image doesn't exist
     ["aria-label"]: "Jane Smith avatar",
     alt: "Jane Smith",
   },


packages/nimbus/src/components/avatar/avatar.tsx
```diff
@@ -18,13 +21,32 @@ export const Avatar = (props: AvatarProps) => {
...
+    setImageError(false);
+  };
+
+  const onError = () => {
+    setImageLoaded(false);
+    setImageError(true);
+  };
+
+  // Show initials if no src provided, image hasn't loaded yet, or image failed
+  const shouldShowInitials = !src || !imageLoaded || imageError;
```
jaikamat, line 36: While we wait for the image to load, we should show initials. The first and last names are required props anyway.

packages/nimbus/src/components/avatar/avatar.tsx
```diff
@@ -18,13 +21,32 @@ export const Avatar = (props: AvatarProps) => {
...
-        getInitials(firstName, lastName)
+      {shouldShowInitials ? getInitials(firstName, lastName) : null}
+
+      {src && (
+        <Image
+          src={src}
+          alt={alt || fullName}
+          onLoad={onLoad}
+          onError={onError}
+          display={imageLoaded && !imageError ? "block" : "none"}
```
jaikamat, line 48: We hide the image while loading or if there's been an error

misama-ct (approved):

packages/nimbus/src/components/avatar/avatar.tsx
ByronDWall: my one concern about showing initials during load is that it might cause the initials to flash quickly, which could be jarring. it might be better to only display them if theres no src or an error?

PR #172, opened 2025-06-04
refactor(form-field): get the props-table to work

misama-ct:
# Summary

The `FormFieldContext` component is not exported anymore, the PropsTable for `FormField` now shows all sub-components props.

## Changes

- Moved `FormFieldContext` and its types to a new file so that it's not exported by default anymore
- Updated component imports to reference the new context file.
- Added the necessary "types-exports" to the main component file

image

### DIFF
--- a/packages/nimbus/src/components/form-field/form-field.tsx
+++ b/packages/nimbus/src/components/form-field/form-field.tsx
@@ -1,48 +1,9 @@
-import { createContext, type ReactNode } from "react";
 import { FormFieldRoot } from "./components/form-field.root";
 import { FormFieldLabel } from "./components/form-field.label";
 import { FormFieldInput } from "./components/form-field.input";
 import { FormFieldDescription } from "./components/form-field.description";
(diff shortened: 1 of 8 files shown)


packages/nimbus/src/components/form-field/form-field.tsx
```diff
@@ -52,3 +13,12 @@ export const FormField = {
...
   InfoBox: FormFieldInfoBox,
 };
+
+export {
+  FormFieldRoot as _FormFieldRoot,
+  FormFieldLabel as _FormFieldLabel,
+  FormFieldInput as _FormFieldInput,
+  FormFieldDescription as _FormFieldDescription,
+  FormFieldError as _FormFieldError,
+  FormFieldInfoBox as _FormFieldInfoBox,
```
ByronDWall, line 23: Should we also document the types for `FormFieldContext`?

ByronDWall (approved):
Thanks!

PR #199, opened 2025-06-13
chore(nimbus): `RadioInput` documentation

jaikamat:
## Summary

This PR creates the Radio Input documentation, based on [this spec](https://www.figma.com/design/gHbAJGfcrCv7f2bgzUQgHq/NIMBUS-Guidelines?node-id=1852-4333&t=u1DTXK73fFg69asB-4).

Note: I'm still omitting the `Specs` header and going straight into `Properties` since those are TBD and this matches currently-existing docs.


packages/nimbus/src/components/radio-input/radio-input.mdx
```diff
@@ -1,36 +1,185 @@
...
+#### Disabled
+
+```jsx-live
+const App = () => (
+  <RadioInput.Root name="disabled-example" isDisabled>
+    <RadioInput.Option value="option1">Disabled option</RadioInput.Option>
+    <RadioInput.Option value="option2">Also disabled</RadioInput.Option>
+  </RadioInput.Root>
+)
+```
```
misama-ct, line 66: nit: those 2 examples are not in figma (but definitely don't hurt).

packages/nimbus/src/components/radio-input/radio-input.mdx
```diff
@@ -1,36 +1,185 @@
...
+- **Complex or dynamic options:** When the available choices change frequently or depend on other user inputs, a dynamic select or autocomplete field might be better.
+
+### Limited options
+
+Radio buttons work best when there are a limited number of options for a user.
+
+> [!TIP]\
+> **Do**
+>
+> Use when there are a few options available for the user.
```
misama-ct, line 123:
```suggestion
> - Use when there are a few options available for the user.
```
:trollface:

packages/nimbus/src/components/radio-input/radio-input.mdx
```diff
@@ -1,36 +1,185 @@
...
+
+![](/images/radio-input/radio-options-do.png)
+
+> [!CAUTION]\
+> **Don't**
+>
+> Do not use radio buttons when there are an excessive amount of options. Radio button options on default will show every option to a user so it can be overwhelming for a user when there is a lot of information to digest in a small area. Instead, use a dropdown selector to simplify the UI and make it easier to search quickly to find the right option.
```
misama-ct, line 130:
```suggestion
> - Do not use radio buttons when there are an excessive amount of options. Radio button options on default will show every option to a user so it can be overwhelming for a user when there is a lot of information to digest in a small area. Instead, use a dropdown selector to simplify the UI and make it easier to search quickly to find the right option.
```

misama-ct (approved):
Only tiny things.

PR #229, opened 2025-06-30
Craft 1567 | implement jsdom polyfill helper

ByronDWall:
In order for nimbus to be testable in test runners that rely on jsdom (`jest`, `vitest`), it is necessary to set up some browser polyfills. This pull request adds a polyfill helper function to satisfy the Nimbus (Chakra UI) provider.

### DIFF
--- a/packages/nimbus/package.json
+++ b/packages/nimbus/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@commercetools/nimbus",
-  "version": "0.0.8",
+  "version": "0.0.9-rc1",
   "main": "./dist/index.cjs",
   "module": "./dist/index.js",
   "type": "module",
--- a/packages/nimbus/package.json
+++ b/packages/nimbus/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@commercetools/nimbus",
-  "version": "0.0.9-rc1",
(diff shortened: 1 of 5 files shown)


packages/nimbus/package.json
```diff
@@ -9,14 +9,19 @@
...
+        "default": "./dist/index.es.js"
       },
       "require": {
         "types": "./dist/index.d.ts",
         "default": "./dist/index.cjs"
       }
     },
-    "./package.json": "./package.json"
+    "./package.json": "./package.json",
+    "./setupJsdomPolyfills": {
```
misama-ct, line 20: super nit: I'd stick to kebab case

misama-ct (approved):

ddouglasz (approved):

ByronDWall:
> Alright, I'll bite. As you're setting this up, how were you able to determine which polyfills were needed? Are these standard, or a result of errors that have encountered to date?

Errors encountered to date.

packages/nimbus/package.json
ByronDWall: done [d32b48c](https://github.com/commercetools/nimbus/pull/229/commits/d32b48c7a6a45576ba7bbd7c10d0c5150ac8841d)

PR #251, opened 2025-07-07
chore(nimbus): create `Divider` component

jaikamat:
## Summary

This PR creates a Divider component.

### DIFF
--- a/packages/nimbus/src/components/divider/divider.mdx
+++ b/packages/nimbus/src/components/divider/divider.mdx
@@ -1,0 +1,168 @@
+---
+id: Layout-Divider
+title: Divider
+description: A visual divider that separates content sections.
+documentState: ReviewedInternal
+documentAudiences: []
+order: 999
+menu:
+  - Components
+  - Layout
+  - Divider
+tags:
+  - document
+---
+
+# Divider
+
+---
+
+A divider provides a visual separator between sections of content, improving readability and creating clear boundaries in layouts.
+
+## Overview
+
+Dividers are fundamental UI elements used to break up content, improve readability, and guide the user's eye. They provide semantic meaning through proper accessibility attributes while maintaining consistent visual styling.
+
+### Basic Usage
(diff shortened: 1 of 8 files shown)


packages/nimbus/src/components/divider/divider.mdx
jaikamat, line 1: This file is just a placeholder for now

FilPob (approved):

packages/nimbus/src/components/divider/divider.recipe.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Defines the styling variants and base styles using Chakra UI's recipe system.
+ */
+export const dividerRecipe = defineRecipe({
+  className: "nimbus-divider",
+
+  // Base styles applied to all instances of the component
+  base: {
+    border: "none",
+    flexShrink: "0",
+    opacity: "1",
```
misama-ct, line 14: This is the default, no need to set it.

packages/nimbus/src/components/divider/divider.recipe.ts
```diff
@@ -0,0 +1,42 @@
...
+    borderColor: "slateAlpha.3",
```
misama-ct, line 15: @FilPob if we use `colorPalette.3` instead, the Divider would have the `neutral` color (slate) by default but adapt to it's surrounding if a parent set a colorPalette.

e.g. If dropped in the `Alert` component with a certain tone, the divider would adapt that tone.

Apart from that, using an alpha-palette is a bit problematic, cause the result of a11y contrast checks are not predictable anymore (cause those checks mix the divider-color with the current background-color and calculate contrast based on that result).

misama-ct (approved):
I have a suggestion, but no blocker. Thank you.

ddouglasz (approved):

tylermorrisford (approved):

PR #276, opened 2025-07-14
chore(nimbus): deprecate `onClick`

jaikamat:
## Summary

Implements onClick deprecation across Nimbus components with pressable context to encourage migration to `onPress` for consistent cross-platform interactions. This follows React Aria's unified press event system that handles mouse, touch, keyboard, and screen reader interactions.

### ✅ **New Utility Type**
- Added `WithDeprecatedOnClick<T>` utility type that marks `onClick` as deprecated with clear migration guidance

### ✅ **Components Updated**
- **Button**: Applied onClick deprecation to `ButtonProps`
- **Link**: Applied onClick deprecation to `LinkProps` + fixed type conflict with `useLink`
- **ToggleButtonGroup.Button**: Applied onClick deprecation to button component
- **IconButton**: Inherits deprecation from Button (no changes needed)

### ✅ **Developer Experience**
- onClick props now show strikethrough in IDEs
- Clear deprecation message: *"Use onPress instead of onClick for consistent cross-platform interactions"*
- Maintains backward compatibility - onClick still works but is discouraged

## Technical Details

- **Type-safe**: Uses TypeScript utility types to overlay deprecation without breaking existing functionality
- **Selective application**: Only applied to components with React Aria's pressable context (`usePress`, `useButton`, `useLink`)
- **No runtime impact**: Pure TypeScript annotations for developer guidance


packages/nimbus/src/utils/deprecate-on-click.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Utility type to mark onClick as deprecated in favor of onPress
+ */
+export type WithDeprecatedOnClick<T> = Omit<T, "onClick"> & {
```
misama-ct, line 4: 😙🤌

misama-ct (changes requested):
I probably wouldn't have done it differently, unfortunately the documentation-pages for the touched components only show a fraction of the previous props now.

Please have a look if you can find a way to make it work, cause the alternative to the automatic parsing is really inconvenient (manually creating the prop-tables for all components & their subcomponents)

jaikamat:
@misama-ct I've gotten `Button` and `ToggleButton`'s Specs to infer correctly, however `Link` seems to still be broken. I don't have any ideas as to why just this one is giving me heartache.

However, something I've noticed - maybe all this work isn't even required? `React-Aria` already has this documented and we parse with the following:
image

We're banking on this improving DX when using `onClick` inside an editor, but our docs already have this covered.

Essentially, I'm throwing up a white flag. I surrender to this complexity 🏳️

misama-ct:
> @misama-ct I've gotten `Button` and `ToggleButton`'s Specs to infer correctly, however `Link` seems to still be broken. I don't have any ideas as to why just this one is giving me heartache.
> 
> However, something I've noticed - maybe all this work isn't even required? `React-Aria` already has this documented and we parse with the following on `main`: image
> 
> We're banking on this improving DX when using `onClick` inside an editor, but our docs already have this covered.
> 
> Essentially, I'm throwing up a white flag. I surrender to this complexity 🏳️

@jaikamat Does it make sense to commit at least the Button changes that are working? After all you spend some time on it and it might already prevent most of the cases.  At first glance the parser adjustments don't seem to have a negative effect either on other components (I compared a hand full of prop-tables with the main-branch docs-live-site without any noticeable difference).

jaikamat:
@misama-ct I should be clear that `Link` has regression. Only two props are inferred, it used to be the whole set. Does that change your opinion on mergeability?

jaikamat:
I'm going to close this for now, we at least have the work available to us should we need to pick back up where we left off.

I really don't think continuing down this path is a good idea for several reasons:
1. Complicating the parsing logic, making future maintainability difficult
2. React-Aria already documents this and we parse it correctly
3. This improves some components but creates regression in others (`Link`)

As it stands, I don't feel this work is worth the tradeoff of just getting strikethroughs in IDE's for product devs, for a small subset of supported components.

PR #279, opened 2025-07-16
Craft 1612 nimbus component datatable

ddouglasz:
This PR implements the Nimbus data table component. 
As part of the main scope the data table work. this PR is now modified to mainly review the `<Header />` of the data table even though the other marked features are working alright.  Figma Design [here](https://www.figma.com/design/AvtPX6g7OGGCRvNlatGOIY/NIMBUS-design-system?node-id=6046-29101&m=dev)

Features/TODO: 
- [x] Select row 
- [x] Row click
- [x] Sorting
- [x] Nested rows
- [x] Sticky Header
- [x] Hover on rows
- [x] Adjustable columns
- [x] Search and highlight
- [x] Custom columns
- [x] Visible columns
- [x] Condensed
- [x] Truncate table
- [x] Table footer should accept reactNode to display any component.
- [x] use design tokens for figma design implementation
- [x] Checkbox clickable area = 24 x 24
- [x] Disable rows
- [x] Copy to clipboard
- [x] Click for more details buttons
- [x] Select all rows
- [x] Clean up inline styling 
- [x] horizontal scroll 
- [ ] Single Row truncation
- [ ] Pin row to top
- [ ] Drag and reorder [Nice to have]
- [ ] Async loading

Reference Priorities from Figma
image

### DIFF
--- a/packages/nimbus/src/components/data-table/components/data-table.column.tsx
+++ b/packages/nimbus/src/components/data-table/components/data-table.column.tsx
@@ -18,13 +18,16 @@ export const DataTableColumn: DataTableColumnComponent = ({
   tabIndex,
   ...otherProps
 }) => {
-  const { sortDescriptor } = useDataTableContext();
+  const { sortDescriptor, isResizable } = useDataTableContext();
   const isActive = sortDescriptor?.column === column?.id;
+  const isColumnResizable =
+    column?.isResizable !== undefined ? column?.isResizable : isResizable;
   /**
    * TODO: Pass styles to DataTableColumnSlot from props,
    * except for height/minHeight/maxHeight which enable react-aria's
    * column resizing
    */
+
   return (
     <DataTableColumnSlot unstyled={unstyled} tabIndex={tabIndex} asChild>
       <RaColumn ref={ref} {...otherProps}>
@@ -53,7 +56,7 @@ export const DataTableColumn: DataTableColumnComponent = ({
                   <ArrowDownward />
                 </DataTableHeaderSortIcon>
               )}
-              {column?.isResizable && (
+              {isColumnResizable && (
                 <ColumnResizer aria-label="Resize column">
                   {({ isResizing, isFocused }) => (
                     <DataTableColumnResizer
--- a/packages/nimbus/src/components/data-table/components/data-table.header.tsx
+++ b/packages/nimbus/src/components/data-table/components/data-table.header.tsx
@@ -42,7 +42,8 @@ export const DataTableHeader = forwardRef<
         <DataTableColumn
           id="selection"
           className="selection-column-header"
-          width={70}
+          maxWidth={72}
+          minWidth={72}
           allowsSorting={false}
           isInternalColumn={true}
         >
@@ -51,10 +52,9 @@ export const DataTableHeader = forwardRef<
       )}
       {showExpandColumn && (
         <DataTableColumn
-          id="expand"
-          width={20}
-          minWidth={20}
-          maxWidth={20}
+          className="expand-column-header"
+          maxWidth={24}
+          minWidth={24}
           allowsSorting={false}
           aria-label="Expand rows"
           isInternalColumn={true}
@@ -79,7 +79,7 @@ export const DataTableHeader = forwardRef<
               isInternalColumn={isDetailsColumn}
               width={column.width}
               defaultWidth={column.defaultWidth}
-              minWidth={column.minWidth}
+              minWidth={column.minWidth ?? 150}
               maxWidth={column.maxWidth}
               column={column}
             >
--- a/packages/nimbus/src/components/data-table/components/data-table.row.tsx
+++ b/packages/nimbus/src/components/data-table/components/data-table.row.tsx
@@ -192,14 +192,7 @@ export const DataTableRow = forwardRef(function DataTableRow<
          * need to be in the same order in the header and row components*/}
         {/* Selection checkbox cell if selection is enabled */}
         {selectionBehavior === "toggle" && (
-          <DataTableCell
-            data-slot="selection"
-            isDisabled={isDisabled}
-            style={{
-              alignItems: "center",
-              justifyContent: "center",
-            }}
-          >
+          <DataTableCell data-slot="selection" isDisabled={isDisabled}>
             <Box
               display="flex"
               alignItems="center"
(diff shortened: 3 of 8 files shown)


packages/nimbus/src/components/data-table/components/data-table.header.tsx
```diff
@@ -0,0 +1,189 @@
+import { forwardRef, useState, useCallback } from "react";
```
ddouglasz, line 1: This is the file to concentrate on for this PR

FilPob:
oh and one more thing: I have the feeling the sorting icons are still coming from UI Kit aren't they? They should be coming from the MUI library and be the `arrow_upward` and `arrow_downward`

FilPob:
Perfect! thanks :)

packages/nimbus/src/components/data-table/components/data-table.row.tsx
```diff
@@ -0,0 +1,289 @@
+import { forwardRef, useState } from "react";
+import { Row as RaRow, Cell as RaCell } from "react-aria-components";
```
ByronDWall, line 2: We should use the built-in collection optimizations from react-aria. [See example from the table content docs](https://react-spectrum.adobe.com/react-aria/Table.html#content)
```suggestion
import { Row as RaRow, Cell as RaCell, Collection as RaCollection } from "react-aria-components";
```

packages/nimbus/src/components/data-table/components/data-table.header.tsx
```diff
@@ -0,0 +1,212 @@
...
+                  padding: 0, 
+                  margin: "-1px", 
+                  overflow: "hidden", 
+                  clip: "rect(0, 0, 0, 0)", 
+                  whiteSpace: "nowrap", 
+                  border: 0 
+                }}>
+                  Details
+                </span>
+              </RaColumn>
```
ByronDWall, line 203: I'm not sure I like the 'show details' button in a column that always renders after the first column of data. I know of at least one table (audit log) where this is not the case. I personally think this should be added by consumers in whichever cell makes sense to them. In any case, I definitely think this needs more discussion.

packages/nimbus/src/components/data-table/components/data-table.header.tsx
```diff
@@ -0,0 +1,212 @@
...
+                  position: "absolute", 
+                  width: "1px", 
+                  height: "1px", 
+                  padding: 0, 
+                  margin: "-1px", 
+                  overflow: "hidden", 
+                  clip: "rect(0, 0, 0, 0)", 
+                  whiteSpace: "nowrap", 
+                  border: 0 
+                }}>
```
ByronDWall, line 200: we use this style object enough that it should be a slot or at least a class defined in the header slot.

packages/nimbus/src/components/data-table/components/data-table.row.tsx
```diff
@@ -0,0 +1,289 @@
...
+    const highlightCell = (value: any) =>
+      search && typeof value === "string" ? (
+        <Highlight query={search}>{value}</Highlight>
+      ) : (
+        value
+      );
+
+    return (
+      <>
+        <RaRow
```
ByronDWall, line 81: A lot of the checking happening for whether a row is disabled that's happening here is handled by react-aria if the `isDisabled` prop is set.
```suggestion
        <RaRow
          isDisabled={isDisabled}
```

packages/nimbus/src/components/data-table/components/data-table.row.tsx
```diff
@@ -0,0 +1,289 @@
...
+                      <KeyboardArrowRight />
+                    )}
+                  </DataTableExpandButton>
+                </Box>
+              ) : null}
+            </RaCell>
+          )}
+
+          {/* Data cells */}
+          {visibleCols.map((col, index) => {
```
ByronDWall, line 166:
```suggestion
         <Collection items={columns}>
           {(columns) => {
```

packages/nimbus/src/components/data-table/components/data-table.header.tsx
```diff
@@ -0,0 +1,106 @@
...
+      }}
+      columns={activeColumns}
+      {...props}
+    >
+      {/** Internal/non-data columns like selection and expand
+       * need to be in the same order in the header and row components*/}
+      {selectionBehavior === "toggle" && (
+        <DataTableColumn
+          id="selection"
+          className="selection-column-header"
```
misama-ct, line 44: This component uses a slot component under the hood, but adds a `className` and applies styling via the applied `className` through the recipe. Why not via the slot property?

packages/nimbus/src/components/data-table/data-table.recipe.tsx
misama-ct, line 1: The recipe file or styling of the table components in general is worrying me. It is very different from all other recipes and the differences can only partially be explained with the complexity of the data-table. I think think by creating enough slot components, it should be possible to get rid of className references and many of the nested selectors.

misama-ct (commented):
I don't know how helpful my comments are, I feel like DataTable is completely different in regards of complexity than the components we did before, so it is hard for me to reason if something was necessary to be done in a certain way.

packages/nimbus/src/components/data-table/components/data-table.cell.tsx
```diff
@@ -0,0 +1,23 @@
...
+  Cell as RaCell,
+  type CellProps as RaCellProps,
+  ButtonContext as RaButtonContext,
+} from "react-aria-components";
+
+export const DataTableCell = forwardRef<
+  HTMLTableCellElement,
+  RaCellProps & { isDisabled?: boolean }
+>(function DataTableCell({ children, isDisabled, ...rest }, ref) {
+  // Basically here for disabling any child buttons via context (magic) when the row is disabled
```
jaikamat, line 12: Nice strategy. As opposed to prop-drilling it to all buttons?

packages/nimbus/src/components/data-table/components/data-table.column.tsx
```diff
@@ -0,0 +1,72 @@
...
+}) => {
+  const { sortDescriptor } = useDataTableContext();
+  const isActive = sortDescriptor?.column === column?.id;
+  /**
+   * TODO: Pass styles to DataTableColumnSlot from props,
+   * except for height/minHeight/maxHeight which enable react-aria's
+   * column resizing
+   */
+  return (
+    <DataTableColumnSlot unstyled={unstyled} tabIndex={tabIndex} asChild>
```
jaikamat, line 29: Probably a dumb question, but why do we allow for `unstyled` to be supported explicitly?

packages/nimbus/src/components/data-table/data-table.recipe.tsx
ByronDWall: @misama-ct here is the [jira ticket](https://commercetools.atlassian.net/browse/CRAFT-1711) to address this concern, we should prioritize it to be worked on as soon as this PR is approved/merged

ByronDWall (approved):
at this point, I think this is a good base from which we can start doing smaller changes and updates from. Iterating on this huge PR will make it impossible to close. Thanks for all the work @ddouglasz

PR #284, opened 2025-07-18
tech-time(docs): Improve dev-experience of nimbus consumers

misama-ct:
Nothing too exciting, essentially just adjustments of comments and docs, but hey, I think for consumers it'll be great if they can quickly access the docs.

I first though I'd need to create placeholders within the JSDoc comments and have some fancy process that replaces the placeholder with the corresponding live-domain (produciton or preview-branch), but honestly, all that matters is the live-domain and this kind of repetitive search and replace operations are perfect use-case for cursor.

image

> [!NOTE]
> Oh, and I also fixed all markdown files that had an additional ruler below the h1 heading (the h1 heading already has a bottom-border).

### DIFF
--- a/packages/nimbus/src/components/accordion/accordion.mdx
+++ b/packages/nimbus/src/components/accordion/accordion.mdx
@@ -17,8 +17,6 @@ figmaLink: >-
 
 # Accordion
 
----
-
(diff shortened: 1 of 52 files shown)


jaikamat (approved):
Nice thought behind this change!

PR #303, opened 2025-07-28
CRAFT-1596: Adjust disabled states for inputs

valoriecarli:
References:
[Figma](https://www.figma.com/design/AvtPX6g7OGGCRvNlatGOIY/NIMBUS-design-system?m=auto&node-id=266-377&t=fK3LADaa9AGswSqN-1)
[Ticket](https://commercetools.atlassian.net/browse/CRAFT-1596)

The following input components have been updated to apply `bg: "neutral.2"` when the `[data-disabled='true']` attribute is present.

These updates affect both the base components and any composed components that rely on them:

- `DateInput` (base)
  - `DatePicker`
  - `DateRangePicker`

- `TextInput` (base)
  - `Password`
  - `FomField` (prop avail, use controls to check)

- `TimeInput` (base)
  - `DatePickerTimeInput`
  - `DateRangePickerTimeInput`

- `NumberInput` (prop avail, use controls to check).
- `MultiLineTextInput`
- `Select` (`SelectInput`) 
- `Combobox`

### DIFF
--- a/packages/nimbus/src/components/combobox/combobox.recipe.tsx
+++ b/packages/nimbus/src/components/combobox/combobox.recipe.tsx
@@ -167,7 +167,6 @@ export const comboBoxSlotRecipe = defineSlotRecipe({
       solid: {
         root: {
           width: "7200",
-          // bg: "purple",
         },
         value: {
           ...selectSlotRecipe.variants?.variant.outline.trigger,


FilPob:
@valoriecarli  thanks! you done everything correct but the combination of the bg-color and lowering opacity in the disabled state somehow looks differently than in Figma. Can we please try to set the bg-color to `neutral.3` instead of `neutral.2` ?

Also some Inputs are not listed:
- Combobox
- SelectInput

FilPob:
Awesome! :) looks much better now. I found one more missing and that is the multi-select Combobox shown here: https://nimbus-storybook-git-craft-1596-disabled-s-476f9c-commercetools.vercel.app/?path=/story/components-combobox--disabled-and-read-only-comboboxes

Approving anyways. thanks!

misama-ct (approved):

PR #323, opened 2025-08-11
CRAFT-1654 - Correct focus in DatePicker/ DateRangePicker 'footer' 

valoriecarli:
[1654](https://commercetools.atlassian.net/browse/CRAFT-1654)

**Issue**:
The footer time input segments in both `DatePicker` & `DateRangePicker` currently shift focus immediate back to the first segment after a keystroke, leading to a frustrating user experience.

**Why?**
The `Date(Range)PickerTimeInput `component has a `useEffect` that listenes for changes to `dateValue`. Designed to auto-focus the first time segment when users selected a date from the calendar, it was also triggering when time values changed through keyboard interaction, causing unwanted focus jumps.

TL; DR - "Whenever time changes, always focus the first segment"..which interfered with user keyboard navigation.

**Fix:**
Added a focus detection check before the auto-focus behavior. Checks if a user is currently focused on any time input segment, and only performs auto-focus if no time segment has focus.

This preserves the intended auto-focus behavior (when selecting dates from calendar) while preventing focus thievery during active keyboard interaction with time segments.

Added a corresponding test for each component's appropriate Story. 

eg:

### DIFF
--- a/packages/nimbus/src/components/date-picker/components/date-picker.time-input.tsx
+++ b/packages/nimbus/src/components/date-picker/components/date-picker.time-input.tsx
@@ -22,6 +22,8 @@ export const DatePickerTimeInput = ({
   // Focus the time input when date changes (user selects a date from calendar)
   useEffect(() => {
     // Check if date changed
+    let timeoutId: NodeJS.Timeout | undefined;
+
     if (dateValue && previousDateRef.current?.compare(dateValue) !== 0) {
       // Only auto-focus if no time input segment currently has focus
       // This prevents stealing focus during user interaction with time segments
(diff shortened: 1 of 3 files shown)


packages/nimbus/src/components/date-picker/components/date-picker.time-input.tsx
```diff
@@ -23,20 +23,29 @@ export const DatePickerTimeInput = ({
...
+      const activeElement = document.activeElement;
+      const hasTimeSegmentFocus =
+        container?.contains(activeElement) &&
+        activeElement?.getAttribute("role") === "spinbutton";
 
-          if (firstSegment) {
-            firstSegment.focus();
+      if (!hasTimeSegmentFocus) {
+        // Small delay to ensure the DOM is ready
+        setTimeout(() => {
```
ddouglasz, line 36: I think we can clearTimeOut to prevent memory leak. 
```
useEffect(() => {
  let timeoutId;

  if (dateValue && previousDateRef.current?.compare(dateValue) !== 0) {
    const container = timeInputRef.current;
    const activeElement = document.activeElement;
    const hasTimeSegmentFocus =
      container?.contains(activeElement) &&
      activeElement?.getAttribute("role") === "spinbutton";

    if (!hasTimeSegmentFocus) {
      timeoutId = setTimeout(() => {
        if (container) {
          const firstSegment = container.querySelector(
            '[role="spinbutton"]'
          ) as HTMLElement;
          firstSegment?.focus();
        }
      }, 50);
    }
  }

  previousDateRef.current = dateValue;

  return () => {
    if (timeoutId) {
      clearTimeout(timeoutId);
    }
  };
}, [dateValue]);
```

ddouglasz (approved):
Only one minor catch, otherwise 🙌🏽

PR #324, opened 2025-08-18
CRAFT-1670: update react-aria related dependencies to latest version

misama-ct:
Rather unspectacular: upgrading the react-aria related packages was not possible cause some behavior changed and broke our tests, so I manually upgraded the packages and adjusted the tests.

### DIFF
--- a/apps/docs/package.json
+++ b/apps/docs/package.json
@@ -40,7 +40,7 @@
     "react": "catalog:react",
     "react-aria": "catalog:react",
     "react-aria-components": "catalog:react",
-    "react-docgen-typescript": "^2.2.2",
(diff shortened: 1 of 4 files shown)


apps/docs/vite.config.ts
```diff
@@ -20,7 +20,7 @@ export default defineConfig({
             "react-syntax-highlighter",
           ],
           react: ["react", "react-dom"],
-          mdx: ["@mdx-js/mdx", "@mdx-js/react"],
```
misama-ct, line 23: = residue from a former dependency

tylermorrisford (approved):
delicious

PR #325, opened 2025-08-18
Craft 1692 | only select row with checkbox, not row click

ByronDWall:
This pull request significantly refines the row click and selection handling logic in the `DataTableRow` component to address React Aria limitations and improve UX consistency. The update introduces utility functions for detecting interactive elements, ensures robust event filtering, and enhances memory management by carefully attaching and cleaning up event listeners. Additionally, accessibility is improved for selection controls.

**Event Handling & Row Click Logic Improvements:**

* Added `getIsTableRowChildElementInteractive` and `stopPropagationToNonInteractiveElements` utility functions to reliably distinguish between interactive and non-interactive row children, preventing accidental row clicks when users interact with buttons, checkboxes, or selection controls.
* Overhauled the row click handling strategy to use native DOM event listeners (pointerdown and click) with capture phase, ensuring that custom row click handlers work even when selection is enabled, and that events are filtered to avoid conflicts with interactive elements.

**Ref & Event Listener Management:**

* Improved the callback ref logic to guarantee that event listeners are only attached once per row instance, preventing duplicate listeners and potential memory leaks.
* Enhanced the cleanup effect to reliably remove both pointerdown and click event listeners when rows are unmounted, further safeguarding against memory leaks in dynamic table scenarios.

**Accessibility Enhancements:**

* Added an explicit `aria-label="select row"` to the row selection `Checkbox` for improved screen reader support and accessibility compliance.


tylermorrisford (approved):
lovely

PR #331, opened 2025-08-19
chore(nimbus): Create `RichTextInput`

jaikamat:
## Summary

This PR creates the RichTextInput. The goal here was to port the UI-Kit version over with sufficient feature parity by shamelessly stealing the bulk of the domain logic and slapping a new UI on top.

I would recommend playing with this and comparing it to what we currently have in UI-Kit, which is [here](https://uikit.commercetools.com/?path=/story/form-inputs-richtextinput--playground-example)

image

## Considerations
This PR also makes a couple additional changes:

1. It removes checkboxes from the Menu Item component, as per direction from @FilPob 
2. It adds some minor changes to the Toolbar divider


tylermorrisford (approved):
This looks really nice. I honestly tried to break it in the storybook and couldn't. Since it's a port from ui-kit (and your porting powers are wizard-level and well documented [*cough* navbar]) I focused on the slots and component file, and `is gorgeous`.

misama-ct (commented):
I checked out the branch and went through the stories. The component looks great and works as expected in my opinion, however, the code is a bit difficult to navigate since this is the first component that has required so much custom logic.

It would be helpful if you could organize the code and file structure to be compatible with our other components but extending it where needed with additional folders (config, constants, utils, hooks, lib) to house custom code that does not necessarily need to live in a component directly.

In my opinion, this would be very beneficial as this won't be the last component that requires custom functionality. `RichTextInput` could act as a blueprint for the structure, and the more our components look alike in terms of structure, the easier it is for developers to get familiar with code they haven't worked on before. This approach would also make it easier to extract functionality for reuse later (...the HTML serialization/deserialization logic we'll likely need again somewhere else).

packages/nimbus/src/components/rich-text-input/components/rich-text-editor.tsx
```diff
@@ -0,0 +1,181 @@
...
+  // Create editor with plugins
+  const editor = useMemo(() => {
+    const baseEditor = createEditor();
+    return withLinks(withHistory(withReact(baseEditor)));
+  }, []);
+
+  // Handle keyboard shortcuts
+  const handleKeyDown = useCallback(
+    (event: React.KeyboardEvent) => {
+      // Handle undo/redo shortcuts
```
ddouglasz, line 79: 👌🏽

ddouglasz (approved):
Huge body of work here! 🙌🏽
Works great. I think once the already pointed out concerns are tackled, from my end, its good to go.

FilPob (approved):

jaikamat:
@misama-ct See e6c117c6005e4a8bd1df4b437ecd8034020cce5c for code reorganization. Let me know if this works for you

misama-ct (approved):

PR #335, opened 2025-08-20
chore(n): fix `SplitButton` icon API

jaikamat:
## Summary

This PR alters the SplitButton consumer API to prefer a prop for `Icon` rather than a slot.


ByronDWall (approved):
This is good work right here.

PR #347, opened 2025-08-22
refactor(text-input): transition to slot recipe system and enhance st…

misama-ct:
Hints & Remarks. I hope it helps.

### DIFF
--- a/packages/nimbus/src/components/text-input/text-input.recipe.tsx
+++ b/packages/nimbus/src/components/text-input/text-input.recipe.tsx
@@ -1,101 +1,76 @@
-import { defineRecipe } from "@chakra-ui/react";
+import { defineSlotRecipe } from "@chakra-ui/react";
 
 /**
  * Recipe configuration for the TextInput component.
- * Defines the styling variants and base styles using Chakra UI's recipe system.
+ * Defines the styling variants and base styles using Chakra UI's slot recipe system.
  */
-export const textInputRecipe = defineRecipe({
+export const textInputSlotRecipe = defineSlotRecipe({
+  slots: ["root", "leadingElement", "input", "trailingElement"],
   // Unique class name prefix for the component
   className: "nimbus-text-input",
 
   // Base styles applied to all instances of the component
   base: {
-    "& .nimbus-text-input-container": {
-      position: "relative",
-      display: "flex",
-      alignItems: "center",
+    root: {
+      display: "inline-flex",
+      cursor: "text",
       borderRadius: "200",
-      width: "100%",
-      height: "100%",
-      overflow: "hidden", // Ensure elements stay within borders
-      focusVisibleRing: "outside",
-    },
-
-    _focusVisible: {
-      // Ensure focus ring surrounds the entire input, including elements
-      zIndex: 1,
-    },
-
-    _disabled: {
-      "& .nimbus-text-input-container": {
-        bg: "neutral.3",
-      },
-      layerStyle: "disabled",
-    },
-
-    "&[data-invalid='true']": {
-      "& .nimbus-text-input-container": {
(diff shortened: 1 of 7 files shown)


packages/nimbus/src/components/text-input/text-input.recipe.tsx
```diff
@@ -1,196 +1,118 @@
-import { defineRecipe } from "@chakra-ui/react";
+import { defineSlotRecipe } from "@chakra-ui/react";
 
 /**
  * Recipe configuration for the TextInput component.
- * Defines the styling variants and base styles using Chakra UI's recipe system.
+ * Defines the styling variants and base styles using Chakra UI's slot recipe system.
  */
-export const textInputRecipe = defineRecipe({
+export const textInputSlotRecipe = defineSlotRecipe({
```
misama-ct, line 7: 1. You probably were not aware, but for single element recipes there is the `defineSlotRecipe` function, for components that combine multiple elements - like here, now that we want to add icons - you'd need to change to `defineSlotRecipe`

packages/nimbus/src/components/text-input/text-input.recipe.tsx
```diff
@@ -1,196 +1,118 @@
...
+  slots: ["root", "leadingElement", "input", "trailingElement"],
```
misama-ct, line 8: 2. Here you'd define the individual slots.

packages/nimbus/src/components/text-input/text-input.tsx
```diff
@@ -21,50 +21,47 @@ import { textInputRecipe } from "./text-input.recipe";
...
     const [styleProps, otherProps] = extractStyleProps(remainingProps);
     const { inputProps } = useTextField(otherProps, ref);
 
     // Focus the input when clicking on the wrapper
     const handleWrapperClick = (e: React.MouseEvent) => {
-      // Don't trigger if clicked on a child with onClick handler
-      if ((e.target as HTMLElement).onclick) {
-        return;
-      }
+      // TODO: reliably ignore interactive elements
```
misama-ct, line 36: This will be a little tricky as you can't just check for an `onClick` handler. What if people use good old links or `onPress` instead of `onClick`?

packages/nimbus/src/components/text-input/text-input.slots.tsx
```diff
@@ -1,48 +1,40 @@
...
- * Root component that provides the styling context for the TextInput component.
- * Uses Chakra UI's recipe context system for consistent styling across instances.
- */
-export const TextInputRootSlot = withContext<
+export interface TextInputRootProps
+  extends HTMLChakraProps<
+    "div",
+    RecipeVariantProps<typeof textInputSlotRecipe>
+  > {}
+export const TextInputRootSlot = withProvider<
```
misama-ct, line 19: You already had the icon-wrapper slot-components here, I added one more for the actual input, only difference introduced by the slotRecipe is that the root needs to be used with `withProvider`, all sub-components with `withContext`.

packages/nimbus/src/components/text-input/text-input.recipe.tsx
```diff
@@ -1,196 +1,118 @@
...
   // Unique class name prefix for the component
   className: "nimbus-text-input",
 
   // Base styles applied to all instances of the component
   base: {
-    "& .nimbus-text-input-container": {
-      position: "relative",
-      display: "flex",
-      alignItems: "center",
+    root: {
```
misama-ct, line 14: Instead of the complicated science with the absolute positioning of the icons and the adjustments of multiple paddings, a flex-box solution is sufficient, if you can manage to get the border on the container/root element and leave the actual input unstyled (apart from the font).

PR #377, opened 2025-09-02
Nimbus Tooling - Update handling of `@chakra-ui/react` external dependency to use a regular expression so that chakra is not bundled with nimbus

ByronDWall:
This pull request improves the external configuration for Chakra UI in the build setup. It also registers the DataTable slot recipe in the theme system, and refactors the DataTable component to consistently use the `dataTableSlotRecipe` naming convention, and updates related type interfaces and imports throughout the codebase.

**Build Configuration**

* Updated the Vite configuration to use a regular expression for Chakra UI externals, improving compatibility with different import patterns.

**Theme System Integration**

* Registered `dataTableSlotRecipe` in the slot recipes theme system by importing and adding it to the `slotRecipes` object in `slot-recipes/index.ts`. [[1]](diffhunk://#diff-a0c0b51c3a05b3e71eb9faa9d95dd314e0fbd04d50aff7e44ab46771ffc14365R17) [[2]](diffhunk://#diff-a0c0b51c3a05b3e71eb9faa9d95dd314e0fbd04d50aff7e44ab46771ffc14365R49)

**DataTable Slot Recipe Refactoring**

* Renamed `dataTableRecipe` to `dataTableSlotRecipe` in `data-table.recipe.tsx` and updated all imports and usages to match the new name for consistency. [[1]](diffhunk://#diff-779c868c5450e23137905b0e9e8a3753315efc4b0e0dcaa65af0589f7a10cb90L7-R7) [[2]](diffhunk://#diff-a3472ec7ee81d74909f1357a6493d2ab298370ac4ac9a1e63e6773b7eb74b45fL7-R13) [[3]](diffhunk://#diff-8082eb10164ed5b969327ed4482093d2b13cced05c978fe18855e2bbb1ac6274L12-R12)
* Updated type interfaces from `DataTableRecipeProps` to `dataTableSlotRecipeProps` and adjusted references in `data-table.slots.tsx` to use the new interface. [[1]](diffhunk://#diff-a3472ec7ee81d74909f1357a6493d2ab298370ac4ac9a1e63e6773b7eb74b45fL7-R13) [[2]](diffhunk://#diff-a3472ec7ee81d74909f1357a6493d2ab298370ac4ac9a1e63e6773b7eb74b45fL25-R28)
* Changed the recipe type used in `DataTableVariantProps` to reference `dataTableSlotRecipe` instead of the old name in `data-table.types.ts`.

**Version Update**

* Bumped the package version from `0.0.12-rc1` to `0.0.12-rc4` in `package.json`.

### DIFF
--- a/packages/nimbus/vite.config.ts
+++ b/packages/nimbus/vite.config.ts
@@ -48,7 +48,7 @@ const external = [
   "react/jsx-runtime",
 
   // UI frameworks & styling.
-  "@chakra-ui/react",
+  new RegExp("@chakra-ui/react?[^.].*$"),
   // TODO: evaluate whether it makes more sense for `react-aria` and related packages to be bundled w/the library as they are currently,
   //       or declared as peer deps to reduce unintentional code duplication if a consuming app already has react-aria related libraries installed (eg @internationalized/date or react-stately).
 
--- a/.changeset/gold-planes-shout.md
+++ b/.changeset/gold-planes-shout.md
@@ -1,0 +1,6 @@
+---
+"@commercetools/nimbus": patch
+---
+
+Update build step to handle multiple chakra-ui entrypoints as peer dependencies
+instead of erroneously including them in the build
--- a/packages/nimbus/package.json
+++ b/packages/nimbus/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@commercetools/nimbus",
-  "version": "0.0.12-rc3",
+  "version": "0.0.12-rc4",
   "main": "./dist/index.cjs",
   "module": "./dist/index.js",
   "type": "module",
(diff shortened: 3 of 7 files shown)


packages/nimbus/vite.config.ts
```diff
@@ -48,7 +48,7 @@ const external = [
   "react/jsx-runtime",
 
   // UI frameworks & styling.
-  "@chakra-ui/react",
+  new RegExp("@chakra-ui/react?[^.].*$"),
```
ByronDWall, line 51: This is the main change that made the `DataTable` work in the MC, otherwise we were bundling all the chakra subdirectories (e.g. `@chakra-ui/react/styled-system`) with `nimbus`, but needed to add `@chakra-ui/react` as a dep in the consuming app. This led to the bundled chakra methods not working with the installed chakra library.

This enables `@chakra-ui/react` to be a peer dependency as intended.

The other option would be to totally bundle chakra with nimbus like we do with `react-aria`.  I'm starting to think we should either make both libraries peer-deps or bundle both, but that's a discussion for another time.

misama-ct (approved):

PR #411, opened 2025-09-05
Nimbus - Data table - fix interactive columns against horizontal scroll

ddouglasz:
- The selection checkbox buttons of the data table, along with the collapsible button and the pin row buttons currently scroll along with the data table. These three columns should be fixed while scrolling horizontally. they should always remain in their position while the rest of the table columns scroll horizontally. 
- Make the pin rows feature of the Data table sticky and always visible without it scrolling out of view.
- Fix pin hover effect that somehow stopped worked.
- Use the opportunity to clean up unused leftover code. 

How to test:
- Go to the all features story 
- Enable `Show all columns"
- Enable Text truncation for a simplified ui. 
- Interact with the select, expand and pin buttons, and also scroll the table horizontally. 

**Result**: these 3 buttons should always stay fixed when the table is scrolled horizontally.

### DIFF
--- a/packages/nimbus/src/components/data-table/components/data-table.row.tsx
+++ b/packages/nimbus/src/components/data-table/components/data-table.row.tsx
@@ -336,7 +336,11 @@ export const DataTableRow = <T extends DataTableRowItem = DataTableRowItem>({
            * need to be in the same order in the header and row components*/}
           {/* Selection checkbox cell if selection is enabled */}
           {selectionBehavior === "toggle" && (
-            <DataTableCell data-slot="selection" isDisabled={isDisabled}>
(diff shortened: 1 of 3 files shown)


tylermorrisford (approved):

ddouglasz:
> Only thing I noticed is that the sorting does not affect the pinned items

@misama-ct Thank you for taking a thorough look! Yes this is how it works at the MC current implementation and i think that is deliberate since the user wants to isolate pinned items from the rest of the table behaviour.

ByronDWall:
@ddouglasz This is definitely an improvement and looks good to me.

One question I have is whether we ever intend to also let users pin columns?

PR #438, opened 2025-09-15
CRAFT-1759 - Package Nimbus i18n for App-Kit Integration

valoriecarli:
**1.)** This PR introduces a new data-only package for Nimbus i18n translations, specifically designed for `mc-appkit` integration.

Taking a simpler approach than uikit... may/ may not work.

TL;DR This is to enable Nimbus components to display localized text in all MC applications.

Why package like this?
- Contains zero code - only translation JSON files in multiple formats (raw + compiled AST)
- No need for entrypoints, dependencies, preconstruct, etc
- Direct consumption by appkit, merge with other sources in `packages/i18n/src/load-i18n.ts`
- Tree-shakeable by nature

**2.)** Made a few changes to the i18n scripts

**3.)** Note that the README isn't really a part of this, but only a draft. Not sure if this is better kept here, or w/in the Nimbus package itself. tbd.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -9,6 +9,7 @@
   "commit": false,
   "fixed": [
     [
+      "@commercetools-nimbus/i18n",
       "@commercetools/nimbus",
       "@commercetools/nimbus-tokens",
       "@commercetools/nimbus-icons"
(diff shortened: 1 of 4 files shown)


misama-ct (approved):
Thank you! I appreciate the README.md efforts, even if not finished yet.

PR #442, opened 2025-09-16
docs(card): replaced placeholders with illustrations

stephsprinkle:
#### Summary
- Card doc examples now have illustrations rather than placeholders
- Updated a heading in multiline text input to match spec (found during reconcilation)


packages/nimbus/src/components/multiline-text-input/multiline-text-input.mdx
```diff
@@ -55,9 +60,10 @@ const App = () => (
  );
 ```
 
-#### Readability
+#### Resizability
```
stephsprinkle, line 63: This is the heading update. The rest is formatting.

PR #451, opened 2025-09-19
fix(data-table): add zIndex to data table column styling for improved…

ddouglasz:
FIx z-index bug with scrolled table body overlapping the header.


packages/nimbus/src/components/data-table/data-table.recipe.tsx
```diff
@@ -176,6 +176,7 @@ export const dataTableSlotRecipe = defineSlotRecipe({
         textStyle: "sm",
         fontSize: "300",
         height: "1000",
+        zIndex: 14,
```
tylermorrisford, line 179: at least it isn't 9999

tylermorrisford (approved):

PR #460, opened 2025-09-23
chore(nimbus): create `CollapsibleMotion` component

jaikamat:
## Summary
Introduces a new CollapsibleMotion compound component built on React Aria's disclosure primitives with proper accessibility support.

Features
- Compound component API: CollapsibleMotion.Root, CollapsibleMotion.Trigger, CollapsibleMotion.Content
- React Aria integration: Full accessibility with useDisclosure and useDisclosureState
- Controlled & uncontrolled modes: Supports both defaultExpanded and isExpanded patterns
- Comprehensive tests: 6 Storybook interaction tests covering all use cases
- Integrates chakra-ui's `Presence` component for animation

### DIFF
--- a/packages/nimbus/src/components/collapsible-motion/collapsible-motion.mdx
+++ b/packages/nimbus/src/components/collapsible-motion/collapsible-motion.mdx
@@ -1,41 +1,47 @@
 ---
-id: collapsible-motion
+id: Components-CollapsibleMotion
 title: CollapsibleMotion
 description: A compound component for creating smooth collapsible content with accessibility support
+order: 999
 menu:
   - Components
   - Layout
-category: layout
-tags: ["collapsible", "accordion", "disclosure", "animation", "accessibility"]
+  - CollapsibleMotion
+tags: ["component", "collapsible", "accordion", "disclosure", "animation", "accessibility"]
 ---
 
 # CollapsibleMotion
 
 A modern, accessible implementation of collapsible content with smooth height animations. Built with React Aria for accessibility and Chakra UI for consistent theming.
 
-## Features
+## Overview
 
-- 🎨 Smooth height animations with customizable duration
-- ♿ Full accessibility support with proper ARIA attributes
-- 🎛️ Both controlled and uncontrolled modes
-- 🎯 Focus management for screen readers
-- 📱 Responsive and mobile-friendly
-- 🎨 Chakra UI theming integration
+CollapsibleMotion is a compound component that provides smooth, accessible collapsible content with height animations. It combines React Aria's accessibility features with Chakra UI's theming system to deliver a robust foundation for disclosure patterns, accordions, and expandable content sections.
 
-## Import
+### Key Features
 
-```jsx-live
-import { CollapsibleMotion } from "@nimbus/components";
-```
+- **Smooth height animations**: Customizable duration with automatic content measurement
+- **Full accessibility support**: Proper ARIA attributes and screen reader compatibility
+- **Controlled and uncontrolled modes**: Flexible state management options
+- **Focus management**: Content visibility handled properly for screen readers
+- **Responsive design**: Works seamlessly across all screen sizes
+- **Chakra UI integration**: Consistent theming with the design system
 
-## Usage
+### Resources
+
+Deep dive on details and access design library.
+
+- [React ARIA Disclosure Docs](https://react-spectrum.adobe.com/react-aria/useDisclosure.html)
+- [ARIA Disclosure Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/disclosure/)
+
+## Variables
 
 ### Basic Usage
 
 The simplest way to use CollapsibleMotion is in uncontrolled mode with default settings:
 
 ```jsx-live
-function BasicExample() {
+const App = () => {
   return (
     <CollapsibleMotion.Root defaultExpanded={false}>
       <Button mb={4}>Toggle Content</Button>
@@ -49,7 +55,7 @@ function BasicExample() {
       </CollapsibleMotion.Content>
     </CollapsibleMotion.Root>
   );
-}
+};
 ```
 
 ### Controlled Mode
@@ -57,7 +63,7 @@ function BasicExample() {
 For more complex scenarios, you can control the expanded state from a parent component:
 
 ```jsx-live
-function ControlledExample() {
+const App = () => {
   const [isExpanded, setIsExpanded] = useState(false);
 
   return (
@@ -84,7 +90,7 @@ function ControlledExample() {
       </CollapsibleMotion.Root>
     </div>
   );
-}
+};
 ```
 
 ### Custom Animation Settings
@@ -92,7 +98,7 @@ function ControlledExample() {
 You can customize the minimum height and animation duration:
 
 ```jsx-live
-function CustomAnimationExample() {
+const App = () => {
   return (
     <CollapsibleMotion.Root defaultExpanded={false}>
       <Button mb={4}>Toggle Content</Button>
@@ -113,7 +119,7 @@ function CustomAnimationExample() {
       </CollapsibleMotion.Content>
     </CollapsibleMotion.Root>
   );
-}
+};
 ```
 
 ### Dynamic Content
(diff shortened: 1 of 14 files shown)


packages/nimbus/src/components/collapsible-motion/collapsible-motion.stories.tsx
```diff
@@ -0,0 +1,304 @@
...
+type Story = StoryObj<typeof meta>;
+
+/**
+ * Basic uncontrolled example with compound component pattern
+ */
+export const Basic: Story = {
+  render: () => (
+    <CollapsibleMotion.Root defaultExpanded={false}>
+      <CollapsibleMotion.Trigger>
+        <Button mb={4}>Toggle Content</Button>
```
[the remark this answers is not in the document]
jaikamat: Yep, claude generated these stories. I didn't audit them for styling, hence the issues

packages/nimbus/src/components/collapsible-motion/collapsible-motion.stories.tsx
```diff
@@ -0,0 +1,304 @@
...
+    // Wait for animation to complete
+    await new Promise((resolve) => setTimeout(resolve, 250));
+    expect(content).toBeVisible();
+  },
+};
+
+/**
+ * Controlled example where parent manages the state
+ */
+export const Controlled: Story = {
```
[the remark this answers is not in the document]
jaikamat: Good thinking! See d526afe3fa91d442f6d1da8764210ad7941a0b8e

packages/nimbus/src/components/collapsible-motion/collapsible-motion.mdx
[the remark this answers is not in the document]
stephsprinkle: Is `CollapsibleMotion` going to need design guidelines or just engineering guidelines?

misama-ct (approved):

packages/nimbus/src/components/collapsible-motion/collapsible-motion.mdx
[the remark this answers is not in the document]
jaikamat: Paging @FilPob, do we need design guidelines for this component? Right now what we have is autogenerated by AI, so wondering how much work to put into this aspect of it

packages/nimbus/src/components/collapsible-motion/components/collapsible-motion-trigger.tsx
```diff
@@ -0,0 +1,48 @@
...
+export const CollapsibleMotionTrigger = forwardRef<
+  HTMLButtonElement,
+  CollapsibleMotionTriggerProps
+>(({ children, onClick, asChild, ...props }, forwardedRef) => {
+  const { buttonProps, isDisabled } = useCollapsibleMotionContext();
+
+  // Use mergeProps to properly compose event handlers like React Aria does
+  const componentProps = mergeProps(props, buttonProps, {
+    disabled: isDisabled,
+    onClick, // Consumer's custom onClick handler
```
ByronDWall, line 25: why are we supporting `onClick` here?  Isn't the idea that the child will be a flavor of the nimbus button (eg an `IconButton` instead of a `Button`) that uses `onPress`?

packages/nimbus/src/components/collapsible-motion/components/collapsible-motion-root.tsx
```diff
@@ -0,0 +1,67 @@
...
+      isDisabled,
+      buttonProps,
+      panelProps,
+      panelRef,
+    };
+
+    return (
+      <CollapsibleMotionContext.Provider value={contextValue}>
+        <CollapsibleMotionRootSlot ref={forwardedRef} {...styleProps} asChild>
+          <div {...functionalProps}>{children}</div>
```
ByronDWall, line 60: do we need this wrapper div?  couldn't we put all the props on the slot and have it wrap `children` directly?

packages/nimbus/src/components/collapsible-motion/components/collapsible-motion-content.tsx
```diff
@@ -0,0 +1,31 @@
...
+>(({ children, ...props }, forwardedRef) => {
+  const { panelProps, panelRef } = useCollapsibleMotionContext();
+
+  // Create a local ref and merge with both panelRef (for React Aria) and forwardedRef (for user)
+  const localRef = useRef<HTMLDivElement>(null);
+  const ref = useObjectRef(mergeRefs(localRef, panelRef, forwardedRef));
+
+  return (
+    <CollapsibleMotionContentSlot ref={ref} {...panelProps} {...props} asChild>
+      <div>{children}</div>
```
ByronDWall, line 26: Is the `div` necessary here or could we use the slot w/o `asChild`?

packages/nimbus/src/components/collapsible-motion/collapsible-motion.types.ts
```diff
@@ -0,0 +1,131 @@
...
+  [key: `data-${string}`]: unknown;
+}
+
+/**
+ * Props for CollapsibleMotion.Trigger component
+ * The clickable element that toggles the disclosure state
+ * Extends button HTML attributes for direct button rendering
+ */
+export interface CollapsibleMotionTriggerProps
+  extends React.ButtonHTMLAttributes<HTMLButtonElement> {
```
ByronDWall, line 109: Do we not want the trigger to be stylable? should we extend the trigger slot props instead?

packages/nimbus/src/components/collapsible-motion/collapsible-motion.tsx
```diff
@@ -0,0 +1,16 @@
+import { CollapsibleMotionRoot } from "./components/collapsible-motion-root";
+import { CollapsibleMotionTrigger } from "./components/collapsible-motion-trigger";
+import { CollapsibleMotionContent } from "./components/collapsible-motion-content";
+
+// Exports for internal use by react-docgen
+export {
+  CollapsibleMotionRoot as _CollapsibleMotionRoot,
+  CollapsibleMotionTrigger as _CollapsibleMotionTrigger,
```
ByronDWall, line 8: The UI Kit [collapsible motion component](https://github.com/commercetools/ui-kit/blob/main/packages/components/collapsible-motion/src/collapsible-motion.tsx#L0-L1) only handles setting enter/exit animations on the "panel"/"content", it doesn't concern itself with the `trigger` at all.

As an animation util, I think this makes sense because there are use cases where components need to be animated on enter, but the event that controls the display/hide is not directly controlled by the user.

Are these components necessary, or could we simplify things to match the UI Kit version?

packages/nimbus/src/components/collapsible-motion/components/collapsible-motion-content.tsx
```diff
@@ -0,0 +1,31 @@
...
+  CollapsibleMotionContentProps
+>(({ children, ...props }, forwardedRef) => {
+  const { panelProps, panelRef } = useCollapsibleMotionContext();
+
+  // Create a local ref and merge with both panelRef (for React Aria) and forwardedRef (for user)
+  const localRef = useRef<HTMLDivElement>(null);
+  const ref = useObjectRef(mergeRefs(localRef, panelRef, forwardedRef));
+
+  return (
+    <CollapsibleMotionContentSlot ref={ref} {...panelProps} {...props} asChild>
```
ByronDWall, line 25: if the consumer passes an `id` to the content slot, it will overwrite the `id` from context, and result in the aria attributes being misconfigured in the DOM, should the spreads here be reversed? Or should the context be handled differently to account for the props passed to the compound components?

ByronDWall (commented):
Overall the approach makes sense in isolation.

I have some questions about what this component is actually meant to do. Hopefully we could simplify it a fair amount by putting a wrapper around chakra's [presence](https://www.chakra-ui.com/docs/components/presence) component and calling it a day?

packages/nimbus/src/components/collapsible-motion/collapsible-motion.tsx
jaikamat: UI-Kit's components render props exposing `toggle` kind of implicitly makes it concerned with colocating a trigger element though, right?

We don't actually _need_ the trigger in the Nimbus component (there's a controlled trigger-less story that covers that), it's really just a bonus to make the consumer's life easier.

FilPob (approved):

PR #477, opened 2025-09-29
chore(nimbus): create `SearchInput`

jaikamat:
## Summary

This PR creates a SearchInput component, which uses RAC's [SearchField](https://react-spectrum.adobe.com/react-aria/SearchField.html#searchfield) internally.

### DIFF
--- a/packages/nimbus/src/components/search-input/search-input.tsx
+++ b/packages/nimbus/src/components/search-input/search-input.tsx
@@ -56,6 +56,7 @@ export const SearchInput = (props: SearchInputProps) => {
             slot="null"
             size="2xs"
             variant="ghost"
+            tone="primary"
             aria-label={intl.formatMessage(messages.clearInput)}
             onPress={() => state.setValue("")}
             opacity={state.value ? 1 : 0}


packages/nimbus/src/components/search-input/search-input.recipe.tsx
```diff
@@ -0,0 +1,160 @@
...
+      "&::-webkit-search-cancel-button": {
+        display: "none",
+      },
+
+      "&::-webkit-search-decoration": {
+        display: "none",
+      },
+    },
+
+    clearButton: {
```
ByronDWall, line 70: @jaikamat the `Select` and `Combobox` components both use an `IconButton` (e.g. [here](https://github.com/commercetools/nimbus/blob/2cd30955d9a23d40887fd568dbde36afdeb31125/packages/nimbus/src/components/combobox/components/combobox.single-select-button-group.tsx#L21)) to render the clear button.

Perhaps it would make more sense to do so here as well, both for consistency and for the sake of not duplicating styles?

packages/nimbus/src/components/search-input/search-input.stories.tsx
```diff
@@ -0,0 +1,785 @@
...
+            {...args}
+            variant={variant}
+            placeholder={variant as string}
+            aria-label={`${variant as string}-disabled`}
+          />
+        ))}
+      </Stack>
+    );
+  },
+  play: async ({ canvasElement, step }) => {
```
ByronDWall, line 117: It might be worth making sure that the clear button is also disabled in this test

ddouglasz (approved):
🙌🏽

packages/nimbus/src/components/search-input/search-input.recipe.tsx
jaikamat: Thanks @ByronDWall, see 5ff2ceb0c1c6064e1e551b7b6f1f6eecd2c23ef0, should fix it

packages/nimbus/src/components/search-input/search-input.stories.tsx
jaikamat: Thanks, see 546328c097b4d9bcd8ead3292d9e0028aefb6e2f

PR #494, opened 2025-10-02
chore(nimbus): create `DataTable` base story tests

jaikamat:
## Summary

This PR creates some basic `DataTable` "Base" story tests to avoid regression when it comes time to create the `DataTableManager`. It also corrects a few spelling/case errors.

### DIFF
--- a/packages/nimbus/src/components/data-table/data-table.stories.tsx
+++ b/packages/nimbus/src/components/data-table/data-table.stories.tsx
@@ -412,12 +412,198 @@ export const Base: Story = {
     },
   },
   parameters: {
+    a11y: {
+      config: {
+        rules: [
+          {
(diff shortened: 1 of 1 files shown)


packages/nimbus/src/components/data-table/data-table.stories.tsx
```diff
@@ -412,13 +412,199 @@ export const Base: Story = {
...
+            id: "color-contrast-apca-custom",
+            enabled: false,
+          },
+          {
+            id: "color-contrast",
+            enabled: false,
+          },
+        ],
+      },
+    },
```
jaikamat, line 428: We do this elsewhere, so there is precedent. The test was emitting color contrast errors. I didn't think I had to solve that here. Might be another revision with design, if needed.

PR #498, opened 2025-10-07
CRAFT-1757: fix(select): fix Select components' ClearButton

misama-ct:
TBD

- Added a `ref` prop to `SelectOptionProps` for better element reference handling.
- Updated `SelectClearButton` to use `slot={null}` for improved event handling.
- Refactored `SelectOptionGroup` to utilize `extractStyleProps` for style management.
- Modified `SelectOption` to separate style and functional props, enhancing flexibility.
- Cleaned up `SelectRoot` component layout for better structure and readability.

### DIFF
--- a/packages/nimbus/src/components/select/components/select.option-group.tsx
+++ b/packages/nimbus/src/components/select/components/select.option-group.tsx
@@ -1,5 +1,6 @@
 import { type ReactNode } from "react";
-
+import { extractStyleProps } from "@/utils/extractStyleProps";
+import { chakra } from "@chakra-ui/react/styled-system";
 import {
   ListBoxSection as RaListBoxSection,
   Header as RaHeader,
@@ -12,6 +13,7 @@ export const SelectOptionGroup = <T extends object>(
   props: SelectOptionGroupProps<T> & { ref?: React.Ref<HTMLDivElement> }
 ) => {
   const { ref, label, items, children, ...restProps } = props;
+  const [styleProps, functionalProps] = extractStyleProps(restProps);
 
   // Validate that children is a function when items is provided
   if (items && typeof children !== "function") {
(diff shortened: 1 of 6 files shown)


packages/nimbus/src/components/select/components/select.option-group.tsx
```diff
@@ -21,24 +23,26 @@ export const SelectOptionGroup = <T extends object>(
   }
 
   return (
-    <RaListBoxSection ref={ref} {...restProps}>
-      <SelectOptionGroupSlot asChild>
-        <RaHeader>{label}</RaHeader>
-      </SelectOptionGroupSlot>
+    <chakra.section {...styleProps} asChild>
```
misama-ct, line 26: To make sure that style-props are not ending up in nirvana, I wrapped the ListBoxSection with a chakra-enabled component which does accept potentially applied style-props.

packages/nimbus/src/components/select/components/select.root.tsx
```diff
@@ -83,7 +83,7 @@ export const SelectRoot = function SelectRoot({
               </Flex>
             )}
 
-            <Flex my="auto" w="600" h="600" pointerEvents="none">
+            <Flex my="auto" w="600" h="600">
```
misama-ct, line 86: Since the parent element already forbids `pointerEvents`, it's unnecessary here.

packages/nimbus/src/components/select/select.stories.tsx
```diff
@@ -360,6 +360,44 @@ export const Clearable: Story = {
...
+          () => {
+            const clearButtonAfter = clearableSelect.querySelector(
+              '[aria-label="Clear selection"]'
+            );
+            expect(clearButtonAfter).not.toBeInTheDocument();
+          },
+          { timeout: 1000 }
+        );
+      }
+    );
```
misama-ct, line 400: Previous tests did not catch that the clear-button is not working. So I added this test, which fails when `slot` on the SelectClearButton's `IconButton` is not set to `null`.

PR #510, opened 2025-10-08
CRAFT-1783 DateRangePickerField component

tylermorrisford:
draaaaaft - This feels mostly on-the-rails but if you see anything amiss please yell.

PR #520, opened 2025-10-10
CRAFT-1784 - Text Input Field

valoriecarli:
This PR introduces the `TextInputField` component.

Relates to:
[DateRangePickerField](https://github.com/commercetools/nimbus/pull/510)
[Ticket](https://commercetools.atlassian.net/browse/CRAFT-1784)
[DL proposal](https://docs.google.com/document/d/15IkiuIeXIGbgOCqw8_kCqtHHTwUBRDGxF967husFbOk/edit?tab=t.0)

### DIFF
--- a/packages/nimbus/src/patterns/fields/text-input-field/text-input-field.tsx
+++ b/packages/nimbus/src/patterns/fields/text-input-field/text-input-field.tsx
@@ -41,6 +41,7 @@ export const TextInputField = ({
   isInvalid = false,
   value,
   placeholder,
+  type = "text",
   ...rest
 }: TextInputFieldProps) => {
   // Determine if we should show errors
@@ -58,7 +59,7 @@ export const TextInputField = ({
       <FormField.Input>
         <TextInput
           {...rest}
-          type="text"
+          type={type}
           placeholder={placeholder}
           onChange={onChange}
           value={value}
--- a/packages/nimbus/src/patterns/fields/text-input-field/text-input-field.types.ts
+++ b/packages/nimbus/src/patterns/fields/text-input-field/text-input-field.types.ts
@@ -18,6 +18,10 @@ export type TextInputFieldProps = Omit<TextInputProps, "type"> &
     "isRequired" | "isInvalid" | "isDisabled" | "isReadOnly" | "id"
   > & {
     /**
+     * HTML input type attribute
+     */
+    type?: "email" | "password" | "search" | "text";
+    /**
      * Label text for the input field (required for accessibility)
      */
     label: ReactNode;


packages/nimbus/src/patterns/fields/text-input-field/text-input-field.types.ts
```diff
@@ -0,0 +1,42 @@
...
+   * Info button will only be visible when this prop is passed.
+   */
+  infoBox?: ReactNode;
+
+  /**
+   * Error object - only truthy values will be rendered
+   * Compatible with FieldErrors format
+   * @see {@link https://github.com/commercetools/nimbus/blob/main/packages/nimbus/src/components/field-errors/field-errors.types.ts#L8}
+   */
+  errors?: Record<string, boolean>;
```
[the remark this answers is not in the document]
tylermorrisford: Between this and the change of the errors type name, I nominate this PR to be merged first and then I'll clean up any issues in [my branch](https://github.com/commercetools/nimbus/pull/510). @valoriecarli @ByronDWall we're good with that plan?

packages/nimbus/src/patterns/fields/text-input-field/text-input-field.tsx
```diff
@@ -0,0 +1,67 @@
...
+  infoBox,
+  onChange,
+  value,
+  placeholder,
+  ...rest
+}: TextInputFieldProps) => {
+  const [styleProps, functionalProps] = extractStyleProps(rest);
+
+  return (
+    <FormField.Root {...functionalProps}>
```
ByronDWall: I think that we should actually consider using the new `extractAriaAttributes` [utility](https://github.com/commercetools/nimbus/pull/540) to pull the necessary aria props for the field, and pass the rest of the props to `TextInput`.

That way the user can pass all the styles, slots if they want to, etc etc

@misama-ct I'd also like to know what your thoughts are here, and also if you have any concerns about these components being able to use their react-aria contexts

misama-ct: Very similar to [this](https://github.com/commercetools/nimbus/pull/510#discussion_r2428478484).

Essentially I think it's not necessary to use the utility here, cause a11y-wise we're safe.

As long as you use `<FormField.Label>` and the Input used inside `<FormField.Input>` supports ref-forwarding, the `aria-*` attributes are all created automatically and are referencing each other correctly.

Below is the base example from storybook. An `id` is automagically generated and this id "links" `<label>` with TextInput's html-`<input/>` via `aria-labelledby`.
If the `FormField.Description` component was used, it would create a similar "link" via `aria-describedby` between the two elements. 

```html
<div data-testid="field-root" class="nimbus-ui-form-field__root css-1cr7e4q">
  <div class="nimbus-ui-form-field__label css-1fylhuh">
    <label id="react-aria4402672854-«r3q»" for="react-aria4402672854-«r3p»">
      Input Label (column)
    </label>
  </div>
  <div class="nimbus-ui-form-field__input css-12r042w">
    <div class="nimbus-text-input__root css-10j7cyj" data-invalid="false" >
      <input class="nimbus-text-input__input css-k18112" id="react-aria4402672854-«r3p»" aria-labelledby="react-aria4402672854-«r3q»" placeholder="Enter some text here" tabindex="0" data-rac="" type="text" value="" >
    </div>
  </div>
</div>
```

(I would like to have the utility though cause I think it's useful for other cases)

ByronDWall: From what I've seen, if you pass a 'custom' `id` directly to the `input` via props, the `id` referenced in the `FormField` components' `aria` attributes is incorrect.

Which means that no matter what `id` should be split off from `restProps` in the prop declaration and passed to `FormField.Root` (sorry @valoriecarli for the back and forth).

Although, this may mean that we need to make `TextInput` consume `TextFieldContext` via `useContextProps`, similar to how `Button` [handles context](https://github.com/commercetools/nimbus/blob/d20c68922d99d2f97a0c2fcaebd1e1a8c7101523/packages/nimbus/src/components/button/button.tsx#L24).

It might end up that we need to update a few components to listen for context?

packages/nimbus/src/patterns/fields/text-input-field/text-input-field.tsx
```diff
@@ -0,0 +1,86 @@
...
+          value={value}
+        />
+      </FormField.Input>
+      {description && (
+        <FormField.Description>{description}</FormField.Description>
+      )}
+      {hasErrors && (
+        <FormField.Error>
+          <FieldErrors
+            id={`${id}-errors`}
```
ByronDWall: `FormField.Error` handles setting an id, and all the aria attributes, etc. So this id isn't necessary in this case
```suggestion
```

ByronDWall (approved):
One concern about the text input's `type`, but other than that looking good!

PR #530, opened 2025-10-14
chore(nimbus): restructure types (part two)

jaikamat:
## Summary

This is part two of the recent TS refactors. The goal here was to streamline all the types files so that they have similar patterns and structures, document them in an ordered fashion, and add jsdoc comments.

### DIFF
--- a/packages/nimbus/src/components/tooltip/tooltip.types.ts
+++ b/packages/nimbus/src/components/tooltip/tooltip.types.ts
@@ -5,6 +5,7 @@ import type {
   RecipeProps,
   UnstyledProp,
 } from "@chakra-ui/react";
+import type { ExcludePolymorphicFromProps } from "@/components/utils/type-helpers";
 
 // ============================================================
 // RECIPE PROPS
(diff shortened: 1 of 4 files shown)


jaikamat:
@misama-ct Interesting, I'll have a look. Not sure why consumers wouldn't have access to props as we `SlotRecipeProps<"componentName">`, I would imagine this would carry over to the stories themselves if consumer editors would be affected. Can you elaborate so I understand better?

jaikamat:
After 467d0e4ab5f1bf50acc55dc3a00473d9fa4cfa91 and f13cf4871979503f435d0ddc17ccb0cf3d5efe35, the build no longer shows any warnings or errors. Using a custom utility type throughout did the trick to make these types portable.

packages/nimbus/src/components/tooltip/tooltip.types.ts
```diff
@@ -1,52 +1,35 @@
...
- * internally, but do not make them available to the consumer.
- *
- * Long rambling background:
- * React-Aria's components cannot be configured to use `as` and `asChild` internally,
- * and cannot be directly styled by chakra's styledSystem. Therefore components
- * from `react-aria-components` should be wrapped in a chakra `withContext`
- * root component to set the styles onto the `r-a-c` component using `asChild`.
- * This means that we need to allow polymorphism internally, but should not
- * allow it in the external props api since it would not work.
- */
```
ByronDWall: We should make util types for this I think, but that's a separate PR

ByronDWall (approved):
yes, yeah, sure, ok

PS: I think we should have more utility types at some point in the future

PR #531, opened 2025-10-14
Craft 1749 Nimbus component data table manager

ddouglasz:
Nimbus component data table manager. 
 Features

- [x] design layout
- [x] functional column manager
- [x] functional layout settings
- [x] search columns
- [x] reorder columns
- [x] Reset columns
- [ ] functional dynamic tab for custom settings(follow up PR)

### DIFF
--- a/packages/nimbus/src/components/data-table/constants.tsx
+++ b/packages/nimbus/src/components/data-table/constants.tsx
@@ -1,4 +1,4 @@
 export const UPDATE_ACTIONS = {
   TOGGLE_TEXT_VISIBILITY: "toggleTextVisibility",
   TOGGLE_ROW_DENSITY: "toggleRowDensity",
-};
+} as const;
--- a/packages/nimbus/src/components/data-table/components/data-table.visible-columns-panel.tsx
+++ b/packages/nimbus/src/components/data-table/components/data-table.visible-columns-panel.tsx
@@ -16,17 +16,17 @@ import {
   VisibilityOff,
   Visibility,
 } from "@commercetools/nimbus-icons";
-import type { TColumnListItem } from "../data-table.types";
+import type { ColumnManagerListItem } from "../data-table.types";
 import { messages } from "../data-table.i18n";
 
 // Type for items with onRemoveItem provided by DraggableList.Root
-type TColumnListItemWithRemove = TColumnListItem & {
+type TColumnListItemWithRemove = ColumnManagerListItem & {
   onRemoveItem?: (key: Key) => void;
 };
 
 // Type guard to check if item has onRemoveItem
 function hasRemoveItem(
-  item: TColumnListItem
+  item: ColumnManagerListItem
 ): item is TColumnListItemWithRemove {
   return "onRemoveItem" in item && typeof item.onRemoveItem === "function";
 }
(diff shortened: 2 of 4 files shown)


packages/nimbus/src/components/data-table/components/data-table.manager.tsx
```diff
@@ -0,0 +1,286 @@
...
+                              })
+                            )}
+                            onUpdateItems={(items) =>
+                              handleVisibleColumnsReorder(
+                                items as ColumnListItem[]
+                              )
+                            }
+                            aria-label={formatMessage(
+                              messages.hiddenColumnsAriaLabel
+                            )}
```
ByronDWall: an `aria-label` isn't necessary if you use the `FormField` to associate a label with the list

ByronDWall (commented):
It's definitely in a good place, for the most part I have a few questions about the handling of the reorder

packages/nimbus/src/components/data-table/components/data-table.manager.tsx
```diff
@@ -0,0 +1,286 @@
...
+ * ```tsx
+ * <DataTable.Root columns={columns} rows={rows}>
+ *   <DataTable.Manager />
+ *   <DataTable.Table>
+ *     <DataTable.Header />
+ *     <DataTable.Body />
+ *   </DataTable.Table>
+ * </DataTable.Root>
+ * ```
+ */
```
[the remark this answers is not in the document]
ddouglasz: The problem with using the field approach is that I have to make some modifications to the way the label is rendered as there is a search input between the label and the draggable list. Also the whole component need some style modifications to match the design.

packages/nimbus/src/components/data-table/constants.tsx
```diff
@@ -0,0 +1,4 @@
+export const UPDATE_ACTIONS = {
```
ddouglasz: This file is likely going to grow into a bunch of settings constants for managing table configurations to decide which action would be triggered.

ddouglasz:
Okay, I am actually impressed with these pointers from claude @misama-ct I see you 🙌🏽
Will take care of the valid ones in the morning.

packages/nimbus/src/components/data-table/components/data-table.manager.tsx
```diff
@@ -0,0 +1,240 @@
...
+        <Settings />
+      </IconButton>
+    </Tooltip.Root>
+  );
+
+  return (
+    <>
+      {defaultTrigger}
+      <Drawer.Root isOpen={isOpen} onOpenChange={setIsOpen} placement="right">
+        <Drawer.Content width="640px">
```
ByronDWall: nit: can we use a design token for the width here? either `xl` (574px) or `2xl` (672px)?

packages/nimbus/src/components/data-table/components/data-table.manager.tsx
```diff
@@ -0,0 +1,240 @@
...
+        variant="ghost"
+        tone="primary"
+        size="xs"
+        aria-label={formatMessage(messages.settings)}
+        onClick={() => setIsOpen(true)}
+      >
+        <Settings />
+      </IconButton>
+    </Tooltip.Root>
+  );
```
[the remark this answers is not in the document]
ddouglasz: Oh, I overthought the approach to conditionally render a trigger component from the user but after cleaning that up, since we wont be needing it, I totally forgot to just add this into the main component. Done, thanks!
https://github.com/commercetools/nimbus/pull/531/commits/064f579580a637bdff624f7a42142c2e69dd499f

packages/nimbus/src/components/data-table/data-table.recipe.tsx
```diff
@@ -199,8 +195,13 @@ export const dataTableSlotRecipe = defineSlotRecipe({
       // Multiline header truncation using webkit line clamp
     },
     table: {
-      width: "fit-content",
+      // width: "fit-content",
       tableLayout: "fixed",
```
misama-ct: TIL: I either never heard about the `table-layout` css-property or last time I used it was so long ago I forgot it exist.

ByronDWall (approved):
Other than the `i18n` id's and a couple small nit's, this looks really good. Thanks!

ddouglasz:
All feedback updated here: https://github.com/commercetools/nimbus/pull/531/commits/afd2a6777b2d88e2495f8654e69fd459a9e45ded

PR #540, opened 2025-10-15
CRAFT-1804: create extractAriaAttributes utility

tylermorrisford:
This PR creates an `extractAriaAttributes` utility function.

Needs to come after [this](https://github.com/commercetools/nimbus/pull/509).

PR #542, opened 2025-10-16
CRAFT-1804: create extractAriaProps

tylermorrisford:
Clean branch - same work as [here](https://github.com/commercetools/nimbus/pull/540)

This work also renames the existing extractStyleProps utility filename to `extract-style-props.ts` and updates all imports for that file.

### DIFF
--- a/docs/file-type-guidelines/compound-components.md
+++ b/docs/file-type-guidelines/compound-components.md
@@ -53,10 +53,10 @@ export const Menu = {
   Trigger: MenuTrigger,
   Content: MenuContent,
 };
-
 ```
 
-Note: Every compound component must have a Root component as the first property to provide configuration and state management.
+Note: Every compound component must have a Root component as the first property
+to provide configuration and state management.
 
 ### 3. Root Accepts Configuration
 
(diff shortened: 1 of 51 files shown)


docs/file-type-guidelines/compound-components.md
```diff
@@ -113,7 +113,7 @@ import { useSlotRecipe } from '@chakra-ui/react/styled-system';
 import { MenuTrigger as RaMenuTrigger } from 'react-aria-components';
 import { MenuRootSlot } from '../menu.slots';
 import type { MenuRootProps } from '../menu.types';
-import { extractStyleProps } from '@/utils/extractStyleProps';
+import { extractStyleProps } from "@/utils/extract-style-props";
```
ByronDWall: instead of importing from the `extract-style-props` file in `@utils`, you should import directly from `@utils`. That way if we refactor filenames/file structure within `/utils`, as long as the type is exported in `/utils/index.ts` the type will still be imported into wherever it's used without any need to make changes.
```suggestion
import { extractStyleProps } from "@/utils";
```

PR #554, opened 2025-10-17
CRAFT-1658: replace tone with colorpalette

tylermorrisford:
This PR replaces the `tone` property with `colorPalette` in components, tests, and docs.

It appears that TextInput does not use a 'tone' property, but we mention tone in the docs [here](https://github.com/commercetools/nimbus/blob/main/packages/nimbus/src/components/text-input/text-input.mdx#tone) - this work removes those references in favor of the `variant` props.

Jira: https://commercetools.atlassian.net/browse/CRAFT-1658

### DIFF
--- a/packages/nimbus/src/components/button/button.types.ts
+++ b/packages/nimbus/src/components/button/button.types.ts
@@ -4,7 +4,7 @@ import type {
   RecipeProps,
   UnstyledProp,
 } from "@chakra-ui/react";
-import type { SemanticPalettesOnly } from "../../type-utils/shared-types";
+import type { SemanticPalettesOnly } from "@/type-utils";
 
 // ============================================================
 // RECIPE PROPS
--- a/packages/nimbus/src/components/toggle-button/toggle-button.types.ts
+++ b/packages/nimbus/src/components/toggle-button/toggle-button.types.ts
@@ -1,6 +1,6 @@
 import type { HTMLChakraProps, RecipeProps } from "@chakra-ui/react";
 import type { ToggleButtonProps as RaToggleButtonProps } from "react-aria-components";
-import type { SemanticPalettesOnly } from "../../type-utils/shared-types";
+import type { SemanticPalettesOnly } from "@/type-utils";
 
 // ============================================================
 // RECIPE PROPS


packages/nimbus/src/components/toggle-button/toggle-button.recipe.ts
```diff
@@ -57,13 +57,10 @@ export const toggleButtonRecipe = defineRecipe({
...
     },
-    tone: {
-      ...buttonRecipe.variants?.tone,
-    },
   },
   defaultVariants: {
     size: "md",
     variant: "outline",
-    tone: "primary",
+    colorPalette: "primary",
```
ByronDWall: You don't need this to be a `variant`, since it is a styled-system prop and is not a value in this recipe's `variants` array.

Instead, you should be able to declare this on line 7:
```
...buttonRecipe.base,
"colorPalette": "primary"
```

packages/nimbus/src/components/toggle-button/toggle-button.mdx
```diff
@@ -172,19 +172,19 @@ const App = () => (
 );
 ```
 
-### Tone
+### Color Palette
 
 #### Primary (default)
 
-The default tone for most toggle buttons.
+The default colorPalette for most toggle buttons.
```
ByronDWall:
```suggestion
The default `colorPalette` for most toggle buttons.
```

packages/nimbus/src/components/toggle-button/toggle-button.stories.tsx
```diff
@@ -180,19 +180,23 @@ export const Tones: Story = {
             <Box width="100px">{tone as string}:</Box>
             {variants.map((variant) => (
               <Stack key={variant} gap="400" align="center" direction="row">
-                <ToggleButton tone={tone} width="128px" variant={variant}>
+                <ToggleButton
+                  colorPalette={tone}
```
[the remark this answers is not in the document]
tylermorrisford: good catch, should have already happened

tylermorrisford:
Screenshot 2025-10-21 at 7 46 51 PM
Looks like `colorPalette` is not a part of Chakra's RecipeProps for button. Should we instead recreate this type with the same values? 

`colorPalette?: "primary" | "critical" | "neutral" | "info" | undefined;`

packages/nimbus/src/components/text-input/text-input.mdx
```diff
@@ -47,30 +55,22 @@ Available for the times a more condensed text input is needed.
...
-### Tone
-
-**Placeholder**
-
-This is the styling for placeholder text. It is a shade lighter than active text.
-
-```jsx-live
-const App = () => <TextInput placeholder="input value"/>
-```
+### Variant
```
tylermorrisford: Swapping `Variant` in for `Tone` here; not saying this is legendary prose or anything but the input doesn't have a `tone` prop.

docs/file-type-guidelines/recipes.md
```diff
@@ -267,7 +267,7 @@ base: {
 
 // ✅ Color palette usage examples
 variants: {
-  tone: {
+  colorPalette: {
```
misama-ct: I think @ByronDWall mentioned it before, this is a native chakra system-property and does not need to be a recipe prop, it works on all components which support style-props.

Component:
```js
<>
  <Button colorPalette="primary"/>
  <Button colorPalette="tomato"/> 
</>
```

Somewhere in the recipe you just use `colorPalette` as placeholder:
```js
// Solid
{
   color: 'colorPalette.contrast',
   bg: 'colorPalette.9'
}
// Subtle
{
  bg: "colorPalette.3",
  color: "colorPalette.11",
}
```

Since colorPalettes work all the same in regards to contrast, step 5 on the primary colorPalette has the same brightness as step 5 of the tomato colorPalette -> this is what makes the palettes interchangeable and allows a clean separation between color and look/variant (solid, subtle, ghost, etc)

packages/nimbus/src/components/toggle-button-group/toggle-button-group.types.tsx
```diff
@@ -22,9 +22,9 @@ type ToggleButtonGroupRecipeVariantProps = {
    */
   size?: "xs" | "md";
   /**
-   * Tone variant
+   * Color palette variant
```
ByronDWall:
```suggestion
   * Available color palettes
```

ByronDWall (approved):
I have a few nits, but my care amount on them is low.

Conditionally approved pending updates based on Michael and Jai's reviews.

Also please add a changeset

jaikamat:
> The different options are an internal design decision, not coming to us from Chakra, correct?

You weren't asking me - but yes. Chakra's `colorPalette` supports all configured system colors out of the box, but these affected components, we are making a conscious decision to narrow the type to a set that design chooses to support so we can limit consumers from going too wild

misama-ct (approved):
Let's get this merged!

PR #596, opened 2025-10-23
fix(docs): preserve search query when clicking outside input

### LINKED ISSUE #595: Issue with the search in the docs-app
The search in the documentation app has an issue. If I open the search dialog an enter a search term and keyboard navigate to an item, everything works. If I however, click outside the input (but still inside the modal) the queryterm disappears from the input (the results remain, but are not clickable anymore.

jaikamat:
Add allowsCustomValue prop to ComboBox to prevent React Aria from clearing the input value when focus is lost. This fixes the issue where clicking outside the search input (but inside the modal) would clear the query and make results non-clickable.

Fixes #595

Generated with [Claude Code](https://claude.ai/code)


ByronDWall (approved):
Approved, I would say that this should use the nimbus combobox, but no reason for that update until the combobox is updated again.

Issue #619, opened 2025-10-27
Bug: Card component


misama-ct:
Card has a default tab-index applied despite not being interactive, please get rid of it.

PR #620, opened 2025-10-27
fix: remove default tabIndex from Card component

misama-ct:
### Investigation and Fix for Card tabIndex Bug ✅

Fixed the [issue](https://github.com/commercetools/nimbus/issues/619) by removing unnecessary focus-related code from the Card component.

**Todo List:**
- [x] Locate Card component files
- [x] Review Card implementation for tabIndex usage
- [x] Remove inappropriate tabIndex and focus ring logic
- [x] Update the test that expects Card to be focusable
- [x] Build and verify changes
- [x] Commit and push changes
- [x] Provide PR creation link

**Changes made:**
1. ✅ Removed `tabIndex={0}` from CardRoot (line 44)
2. ✅ Removed `useFocusRing()` hook and related imports (`mergeProps` from react-aria)
3. ✅ Removed focus-related data attributes (`data-focus`, `data-focus-visible`)
4. ✅ Updated story test to remove keyboard focus test step

**Why this fix is correct:**
The Card component is a layout/display container, not an interactive element. It should only be focusable when it contains interactive content or is explicitly made interactive by the consumer (e.g., with an `onClick` handler). The hardcoded `tabIndex={0}` was making all Cards focusable, which creates unnecessary tab stops and poor keyboard navigation UX.

### DIFF
--- a/.changeset/hungry-points-fry.md
+++ b/.changeset/hungry-points-fry.md
@@ -1,0 +1,5 @@
+---
+"@commercetools/nimbus": patch
+---
+
+removed unnecessary tabIndex property from Card


ByronDWall:
@misama-ct the link-overlay is really interesting - I think we should definitely consider exposing it for composition.

also, plz add a changeset before merging

PR #628, opened 2025-10-28
CRAFT-1856: KBD docs

tylermorrisford:
This PR creates the `<Kbd>` component design docs. Jira ticket [here](https://commercetools.atlassian.net/browse/CRAFT-1856)

### DIFF
--- a/packages/nimbus/src/components/kbd/kbd.mdx
+++ b/packages/nimbus/src/components/kbd/kbd.mdx
@@ -22,10 +22,6 @@ The KBD component visually renders a keystroke or key combination to highlight i
 
 The keyboard key component provides semantic and visual representation for keyboard input, such as a shortcut, key press, or command line input, ensuring clarity and consistency when providing user instructions.
 
-### Resources
-
-Deep dive into implementation details.
-
 ## Variables
 
 Get familiar with the features. 


packages/nimbus/src/components/kbd/kbd.mdx
```diff
@@ -1,42 +1,90 @@
...
   - key
-
 ---
 
-# Kbd
+# KBD
+
+The KBD component visually renders a keystroke or key combination to highlight it within textual instructions.
+
+### Overview
```
stephsprinkle: Claude is right, the hierarchy needs to be updated

Image

packages/nimbus/src/components/kbd/kbd.mdx
```diff
@@ -1,42 +1,90 @@
...
 
-The Kbd component renders keyboard shortcuts or key combinations with proper
-styling to visually represent keyboard keys. It uses a monospace font and
-keyboard-like appearance to make it clear that the text represents a key or key
-combination.
+The keyboard key component provides semantic and visual representation for keyboard input, such as a shortcut, key press, or command line input, ensuring clarity and consistency when providing user instructions.
 
-## Usage
+#### Resources
```
stephsprinkle: Since we don't have a link to the design system in Figma, do we want this section?

tylermorrisford: This is a good question - I only kept it so we'd have some kind of visual continuity with other docs, but I'm not sure how important that is. @ellelevy what do you think?

ddouglasz (approved):

PR #630, opened 2025-10-28
CRAFT-1822 | combobox refactor

ByronDWall:
Refactored ComboBox to use a unified context-based architecture for better state management and added several missing features that were causing usability issues.

  Why this change was necessary:

  The previous ComboBox implementation had fragmented state management that made it difficult to coordinate behavior between the input, dropdown, and selection components. This caused issues with:

  - Keyboard navigation not working reliably
  - Multi-select tags not syncing properly with selections
  - Async data loading being difficult to implement
  - Missing critical features like custom option creation and advanced filtering
  - Inconsistent behavior between single and multi-select modes

  The new architecture consolidates state management into a single context pattern, enabling proper coordination between all component parts and making it straightforward to add features like async loading, custom filters, and creatable options. This brings ComboBox up to par with other production-grade autocomplete components and unblocks several user-facing feature requests.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -408,6 +408,9 @@ importers:
       react-router-dom:
         specifier: ^7.11.0
         version: 7.11.0(react-dom@19.2.0(react@19.2.0))(react@19.2.0)
+      react-stately:
+        specifier: 3.42.0
+        version: 3.42.0(react@19.2.0)
       react-syntax-highlighter:
         specifier: ^15.6.6
         version: 15.6.6(react@19.2.0)
@@ -429,6 +432,9 @@ importers:
       to-vfile:
         specifier: ^8.0.0
         version: 8.0.0
+      use-debounce:
+        specifier: ^10.0.0
+        version: 10.0.6(react@19.2.0)
       vite-plugin-markdown:
         specifier: ^2.2.0
         version: 2.2.0(vite@7.3.0(@types/node@24.10.1)(terser@5.44.0)(tsx@4.21.0)(yaml@2.8.1))


packages/nimbus/src/components/combobox/combobox.mdx
```diff
@@ -46,26 +46,36 @@ const App = () => {
...
-      <ComboBox.Root size="md" defaultItems={options} placeholder="Search...">
-        {(item) => (
-          <ComboBox.Option id={item.id} textValue={item.name}>
-            <Stack gap="100">
-              <Text textStyle="sm">{item.name}</Text>
-              <Text textStyle="xs" color="neutral.11">{item.description}</Text>
-            </Stack>
-          </ComboBox.Option>
-        )}
+      <ComboBox.Root size="md" items={options} placeholder="Search...">
```
valoriecarli: Note that 0b1a6268305352b3d42c20d5c25b82842d3dfff6 added slight changes so that the code examples would render appropriately. 

Prior to the update, all Visual options displayed as `Error:  cannot be rendered outside a collection.`
And the Async example displayed as `ReferenceError: useAsyncList is not defined`

packages/nimbus/src/components/combobox/combobox.stories.tsx
```diff
@@ -203,2648 +96,5869 @@ const mockFn = fn();
...
+/**
+ * A11y Keyboard: No Keyboard Traps
+ * Tests that user can Tab out of component
+ */
+export const AccessibilityKeyboardNoTraps: Story = {
+  render: () => {
+    return (
+      <Stack direction="column" gap="400">
+        <ComposedComboBox aria-label="First combobox" items={simpleOptions} />
+        <ComposedComboBox
```
tylermorrisford: This variant:solid story fails for me in the Storybook preview but *not* on the branch... approving anywayz

tylermorrisford (approved):

packages/nimbus/src/components/combobox/combobox.stories.tsx
jaikamat: It's a light mode/dark mode thing. Fails for me on dark mode too, good catch

PR #674, opened 2025-11-13
CRAFT - 1869 - Docs - `Field errors`

valoriecarli:
[Figma](https://www.figma.com/design/gHbAJGfcrCv7f2bgzUQgHq/NIMBUS-Guidelines?node-id=4706-5182)
[Ticket](https://commercetools.atlassian.net/browse/CRAFT-1869)


packages/nimbus/src/components/field-errors/field-errors.mdx
```diff
@@ -16,485 +16,296 @@ tags:
...
-  );
-};
-```
 
-## Custom messages for built-in types
+> [!TIP]\
+> When to use
+- **Validation Fails on Blur/Touch:** Use an error message immediately after the user attempts to interact with the field (clicks in and clicks out) and the input fails validation (e.g., they leave a required field empty).
+- **Preventing Form Submission:** Use an error message to clearly indicate which fields must be corrected before the form can be successfully submitted.
+- **Providing Actionable Correction:** Use the message to tell the user exactly how to fix the problem (e.g., "Password must contain one number" or "Please enter 8 or more characters").
```
stephsprinkle:
```suggestion
- **Providing actionable correction:** Use the message to tell the user exactly how to fix the problem (e.g., "Password must contain one number" or "Please enter 8 or more characters").
```

PR #682, opened 2025-11-17
Craft 1828 table manager custom columns

ddouglasz:
Make provision for users to be able to handle a 3rd tab for custom table settings in the data table manager.

image


ddouglasz:
> We may have to start creating multiple story files, 4000 lines is just too long 😄

I agree 😅

PR #690, opened 2025-11-18
CRAFT 1991 | update `colors.bg` and `colors.fg` global css values to use color palettes instead of named css colors

ByronDWall:
This pull request updates the color token definitions and documentation in the Nimbus design system to improve clarity and consistency. The main changes involve refining the description of contrast colors, updating the semantic token values for background and foreground colors, and fixing a typo in the info color token.

**Documentation improvements:**
* Refined the description for the `contrast` color in `colors.mdx` to clarify its use as high-contrast text (black/white) on solid backgrounds.

**Semantic token updates:**
* Updated the default values for background (`bg.DEFAULT`) and foreground (`fg.DEFAULT`) colors in `colors.ts` to use neutral palette tokens instead of hardcoded color values.

**Bug fixes:**
* Corrected a typo in the `info` color token value from `{colors.ingo.8}` to `{colors.info.8}` in `colors.ts`.


packages/nimbus/src/theme/semantic-tokens/colors.ts
```diff
@@ -44,7 +38,7 @@ export const colors = defineSemanticTokens.colors({
       value: "{colors.positive.8}",
     },
     info: {
-      value: "{colors.ingo.8}",
+      value: "{colors.info.8}",
     },
   },
```
misama-ct: Do we need the border category? I never used it and also don't remember why I left it here. I assume there was a reason, but well ...

(Nothing against semantic tokens, but we might want to research and figure out how to most effectively combine them with the `colorPalette` feature. Would extending the colorPalette placeholder work equally? -> `colorPalette.border`, `colorPalette.solid`, `colorPalette.solid-hover`)

PR #694, opened 2025-11-19
chore(docs): `LoadingSpinner` engineering docs

jaikamat:
## Summary

Does what the title says.


ddouglasz (approved):
Its funny how I am reading all these docs and I can immediately see the value in going through them before using even the smallest components 🎯

PR #711, opened 2025-11-20
chore(docs): `MoneyInput` engineering docs

jaikamat:
## Summary

Does what the title says.

### DIFF
--- a/packages/nimbus/src/components/money-input/money-input.dev.mdx
+++ b/packages/nimbus/src/components/money-input/money-input.dev.mdx
@@ -337,6 +337,221 @@ const App = () => {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


packages/nimbus/src/components/money-input/money-input.dev.mdx
```diff
@@ -0,0 +1,427 @@
...
+        value={value}
+        currencies={["USD"]}
+        isInvalid
+      />
+    </Stack>
+  );
+}
+```
+
+### Modern event handlers
```
ddouglasz: I'm not sure we need to use the term "Modern" here. I think we can just use "Event handlers". What do you think.

PR #751, opened 2025-12-01
fix(docs): adds a fields route for the patterns page

tylermorrisford:
This pull request introduces a `Field` route to group the existing fields under the `Patterns` route. See below!

Before:
Screenshot 2025-12-01 at 5 00 27 PM

After:
Screenshot 2025-12-01 at 5 01 04 PM

This had been bugging me, I created a very after-the-fact ticket [here](https://commercetools.atlassian.net/browse/CRAFT-2011).

# Build Issues Encountered

🚨 UPDATE: 🚨  We had imported `BrowserRouter` from `react-router` instead of from `react-router-dom`; it appears that react-router-dom is deprecated and only used for updating from react-router v6 to react-router v7, which the docs app is on... which makes this seem like a band-aid fix for a problem I didn't know I was creating. 

I am happy to close this (or close this for me if you like) if there is an easier way to enable the field components to be shown not just in the left sidebar but... where they should be seen in the body of the page.

### DIFF
--- a/apps/docs/src/main.tsx
+++ b/apps/docs/src/main.tsx
@@ -1,7 +1,7 @@
 import { StrictMode } from "react";
 import { createRoot } from "react-dom/client";
 import App from "./app.tsx";
-import { BrowserRouter } from "react-router";
+import { BrowserRouter } from "react-router-dom";
 
 // Enable granular HMR for MDX files instead of full page reloads
 if (import.meta.hot) {


valoriecarli:
@tylermorrisford leaving it here so i dont forget... i think it's interesting too that all checks were successful. 🤔

ddouglasz:
> @tylermorrisford leaving it here so i dont forget... i think it's interesting too that all checks were successful. 🤔

I think that was before the last commit he pushed to fix it 🤘🏼

tylermorrisford:
@misama-ct I know you're _dying_ to look at this... 🙏 😍

PR #785, opened 2025-12-04
CRAFT-1954: Switch engineering docs

tylermorrisford:
This PR creates the `Switch` component engineering docs, wouldn't ya know.

### DIFF
--- a/packages/nimbus/src/components/switch/switch.dev.mdx
+++ b/packages/nimbus/src/components/switch/switch.dev.mdx
@@ -168,7 +168,7 @@ const App = () => {
       <Text fontSize="350">
         Current state: <strong>{isEnabled ? 'On' : 'Off'}</strong>
       </Text>
-      <Button size="sm" onClick={() => setIsEnabled(!isEnabled)}>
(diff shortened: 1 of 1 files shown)


packages/nimbus/src/components/switch/switch.dev.mdx
```diff
@@ -0,0 +1,365 @@
...
+
+  return (
+    <Stack gap="400">
+      <Switch isSelected={isEnabled} onChange={setIsEnabled}>
+        Notifications enabled
+      </Switch>
+      <Text fontSize="350">
+        Current state: <strong>{isEnabled ? 'On' : 'Off'}</strong>
+      </Text>
+      <Button size="sm" onClick={() => setIsEnabled(!isEnabled)}>
```
jaikamat: only nit, we don't have a small size for Button. Maybe we remove the `size` prop in this example. 

Also I have NO IDEA why Claude INSISTS on using `sm` for `Button`, it's driving me NUTS

tylermorrisford: well, now that you said it out loud 👀 😆

PR #786, opened 2025-12-04
CRAFT-1955: Time-Input Engineering Docs

tylermorrisford:
Creates the engineering docs for `time-input`.


packages/nimbus/src/components/time-input/time-input.dev.mdx
```diff
@@ -0,0 +1,724 @@
...
+      aria-label="Time with trailing icon"
+      defaultValue={new Time(17, 0)}
+    />
+    <TimeInput
+      leadingElement={<Icon as={Clock} />}
+      trailingElement={<Icon as={Schedule} />}
+      aria-label="Time with both icons"
+      defaultValue={new Time(12, 30)}
+    />
+  </Stack>
```
misama-ct: Image

Within the docs-app, the Icons are not available in the global namespace (since some icon names are very generic and would overwrite global js objects & functions).

Instead, they are accessible under the _Icons_ namespace. Assuming `Clock` exists as an icon in the nimbus-icons package, it would be `<Icon as={Icons.Clock} />`

packages/nimbus/src/components/time-input/time-input.dev.mdx
```diff
@@ -0,0 +1,724 @@
...
+        aria-label="Controlled time input"
+      />
+      <Stack direction="row" gap="300">
+        <Button size="sm" onPress={() => setValue(new Time(9, 0))}>
+          Set to 9:00 AM
+        </Button>
+        <Button size="sm" onPress={() => setValue(new Time(17, 0))}>
+          Set to 5:00 PM
+        </Button>
+        <Button size="sm" variant="secondary" onPress={() => setValue(null)}>
```
jaikamat: Once again, AI out here slapping us with that `sm` Button prop. Let's just remove it so we have the default Button.

jaikamat (approved):

PR #818, opened 2025-12-11
chore(nimbus): `ComboBox` unit tests and bugfixes

jaikamat:
## Summary

This PR creates a large amount of unit tests for the combobox to ensure appropriate functionality. It also contains some small bug fixes I discovered along the way.


FilPob:
@jaikamat thanks! UI wise I found one little thing: There should be no focus ring when hovering the menu items. This is only the cace in the multi-select variations... In single selects the hover states are correct. 
image

PR #820, opened 2025-12-11
CRAFT-1963: Stack engineering docs

tylermorrisford:
Creates engineering docs for the `Stack` component. 

🙇‍♂️ thank you to @misama-ct for the assist on the missing props table 📶

### DIFF
--- a/packages/nimbus/src/components/stack/stack.dev.mdx
+++ b/packages/nimbus/src/components/stack/stack.dev.mdx
@@ -367,10 +367,18 @@ const App = () => (
     <Box>
       <Text fontWeight="600" mb="300">Icon Buttons:</Text>
       <Stack direction="row" gap="100">
-        <IconButton icon={Edit} aria-label="Edit" size="xs" variant="ghost" />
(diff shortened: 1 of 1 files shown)


packages/nimbus/src/components/stack/stack.dev.mdx
```diff
@@ -0,0 +1,459 @@
...
+
+If your use case requires tracking and analytics for this component, it is good practice to add a **persistent**, **unique** id to the component:
+
+```tsx
+<Stack id="main-navigation-stack">
+  {/* children */}
+</Stack>
+```
+
+## API reference
```
tylermorrisford: Our PropsTable component returns a 'nothing found' message, so I had Claude roll his own. What say ye?

misama-ct: There is no proper types file, thus the issue. I usually wouldn't do it but you won't be online today, I wont be online till next year, so I took the liberty and just pushed a tiny commit fixing this.

misama-ct: ... and then my fix broke some stories, so I pushed another fix that is fixing the fixes effects.

PR #826, opened 2025-12-12
CRAFT - 1956 - [Eng Docs] - `Box`

valoriecarli:
[Ticket reference](https://commercetools.atlassian.net/browse/CRAFT-1956)

### DIFF
--- a/packages/nimbus/src/components/box/box.dev.mdx
+++ b/packages/nimbus/src/components/box/box.dev.mdx
@@ -88,7 +88,7 @@ const App = () => (
 
 ### As a flex container
 
-Use flexbox props to create flexible layouts in both horizontal and vertical orientations:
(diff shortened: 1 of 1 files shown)


packages/nimbus/src/components/box/box.dev.mdx
```diff
@@ -0,0 +1,370 @@
...
+      Box with shadow
+    </Box>
+    <Box p="100" border="solid-50" borderColor="primary.6" borderRadius="200" shadow="3">
+      Border, radius, and shadow
+    </Box>
+  </Stack>
+)
+```
+
+### As a flex container
```
stephsprinkle: Should we add a "see `Flex`" tip here? Same goes for the grid example.

PR #833, opened 2025-12-15
chore(poc): explore intents

jaikamat:
image

This is rad

See this [article](https://shopify.engineering/mcp-ui-breaking-the-text-wall) which is the only place I could find an example of the intents system.

```
When a user clicks "Add to Cart" inside an embedded component, what happens? MCP UI solves this with an intent-based message system. Components don't directly modify state—they bubble up intents that the agent interprets
```

This is a powerful paradigm which allows us to cut scope on creating event handlers while also leveraging agents for stateful interactions.

### DIFF
--- a/apps/mcp-ui-poc/client/src/components/chat-interface.tsx
+++ b/apps/mcp-ui-poc/client/src/components/chat-interface.tsx
@@ -10,7 +10,6 @@ import {
 } from "@commercetools/nimbus";
 import { isUIResource } from "@mcp-ui/client";
 import { ClaudeClient } from "../lib/claude-client";
-import { FormSubmissionDialog } from "./form-submission-dialog";
 import { ChatInput } from "./chat-input";
 import { ChatLoadingIndicator } from "./chat-loading-indicator";
 import type { Message } from "../types/virtual-dom";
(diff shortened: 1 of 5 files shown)


apps/mcp-ui-poc/server/src/elements/button.ts
```diff
@@ -8,6 +8,8 @@ export interface ButtonElementArgs {
   isDisabled?: boolean;
   type?: "button" | "submit" | "reset";
   ariaLabel?: string;
+  /** Optional intent to emit when button is pressed */
+  intent?: Intent;
```
jaikamat: @ByronDWall It is powerful, if not risky, to give the agent complete control over the intent.

One of the issues I've been grappling with lately is _how much control_ we exert over these UI templates when it comes to statefulness, that is, what priority data do _we_ attach to handlers (product name, id, sku, etc), and I chose to lean into the inverse of this to give agents more control (continued below)

apps/mcp-ui-poc/client/src/components/chat-interface.tsx
```diff
@@ -159,6 +160,67 @@ export function ChatInterface() {
     }
   }
 
+  // Handle intents emitted from UI components
+  async function handleIntentEmit(intent: Intent) {
+    console.log("🎯 Intent received:", intent);
+
+    // Combine description + structured payload for Claude
+    const message = `${intent.description}
```
jaikamat: We don't have to show a message to the user. It's just here for DX

jaikamat:
Closing as the architecture has moved beyond this

PR #837, opened 2025-12-17
CRAFT-2027-Export messages to `NimbusProvider`

valoriecarli:
This PR adds internationalization (i18n) support to Nimbus components. Consumers will pass a `locale` prop, and the `NimbusProvider` automatically load translation messages.

### DIFF
--- a/.changeset/export-i18n-messages.md
+++ b/.changeset/export-i18n-messages.md
@@ -2,5 +2,4 @@
 "@commercetools/nimbus": minor
 ---
 
-Export translation message utilities and hook to enable consumers to load and
-provide Nimbus i18n messages to the NimbusProvider.
+Add i18n support to Nimbus components via the `NimbusProvider`.


packages/nimbus/src/components/nimbus-provider/nimbus-provider.mdx
```diff
@@ -36,24 +36,95 @@ function App() {
 }
 ```
 
-## Set up locale
+## Set up translation messages
 
-The `NimbusProvider` supports internationalization through the `locale` prop. You can specify a [BCP47 language tag](https://en.wikipedia.org/wiki/IETF_language_tag) to set the locale for all Nimbus components.
+To enable translations for Nimbus components, you need to load and provide translation messages to the `NimbusProvider`. This is a **one-time setup at your application's root level**. Once configured, all Nimbus components throughout your application will automatically use the correct translations. Nimbus provides utilities to load these messages from the `@commercetools/nimbus-i18n` package.
```
ByronDWall:
```suggestion
To enable internationalized messages in Nimbus components, you need to load and provide translation messages to the `NimbusProvider`. This is a **one-time setup at your application's root level**. Once configured, all Nimbus components throughout your application will automatically render the correct internationalized messages. Nimbus provides utilities to load these messages from the `@commercetools/nimbus-i18n` package.
```

packages/nimbus/src/utils/i18n.ts
```diff
@@ -0,0 +1,118 @@
...
+        const messagesModule =
+          await import("@commercetools/nimbus-i18n/compiled-data/en.json");
+        return (messagesModule.default || messagesModule) as IntlMessages;
+      } catch {
+        return {};
+      }
+    }
+    return {};
+  }
+}
```
ByronDWall: in order to use this hook `@commercetools/nimbus-i18n` _must_ be installed - at the very least the catch block needs to handle errors if the hook is used without the messages installed explaining that the messages need to be installed.

PR #877, opened 2026-01-12
chore(nimbus): `DataTable.Manager` documentation

jaikamat:
This one was a weird one. Since `DataTable.Manager` isn't its own component, but rather a `DataTable` compound component, I had to break convention here and extend the `DataTable` docs with its information. It's just a weird component in that it warrants its own documentation.

Which, honestly, is fine. It's colocated with with the only component that it's compatible with anyway.


tylermorrisford (approved):
looks good!

PR #887, opened 2026-01-13
CRAFT-1931- Eng docs - `List`

valoriecarli:
Ill give you one hint.

### DIFF
--- a/packages/nimbus/src/components/list/list.dev.mdx
+++ b/packages/nimbus/src/components/list/list.dev.mdx
@@ -18,16 +18,14 @@ import { List } from '@commercetools/nimbus';
 
 ### Basic usage
 
-The simplest implementation uses semantic HTML with the `asChild` prop:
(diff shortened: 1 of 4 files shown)


packages/nimbus/src/components/list/list.dev.mdx
```diff
@@ -0,0 +1,439 @@
...
+
+```jsx-live-dev
+const App = () => (
+  <List.Root asChild>
+    <ul>
+      <List.Item>First item</List.Item>
+      <List.Item>Second item</List.Item>
+      <List.Item>Third item</List.Item>
+    </ul>
+  </List.Root>
```
misama-ct: The list uses an `ul` tag as default. Any clue why `asChild` is used all over the place in many other examples here? One example demonstrating it with `ol` would be enough.

tylermorrisford (approved):
Is so beautiful!

PR #927, opened 2026-01-20
CRAFT-1903 - Split `Implementation` tab (#1)

valoriecarli:
[Associated ticket](https://commercetools.atlassian.net/browse/CRAFT-1903)

This PR covers the following:
- Relocated the API Reference (aka Props, Properties, Specs, etc) to only exist under the `Implementation` tab.
  - If components lack eng docs, but they are currently in review and have the API section, they were also removed.
- Standardized leftover heading capitalization.
- Additional misc. cleanup

Unrelated but corrected:
- Fixed component name mapping: combo-box → combobox in i18n compilation script
- i18n cleanup that can't wait until next week.

Disclaimer - There is a chance something was missed. If so, I'll grab it on the next round.

### DIFF
--- a/packages/nimbus/src/components/image/image.dev.mdx
+++ b/packages/nimbus/src/components/image/image.dev.mdx
@@ -328,6 +328,22 @@ export const Example = () => (
 
 <PropsTable id="Image" />
 
+
+The Image component accepts all standard HTML `<img>` attributes including:
+- `src`: Image source URL
+- `alt`: Alternative text for accessibility
+- `srcSet`: Responsive image sources
+- `sizes`: Size hints for responsive images
+- `loading`: "lazy" or "eager" loading behavior
+- `decoding`: "async", "sync", or "auto" for image decoding
+- `crossOrigin`: CORS settings
+- `referrerPolicy`: Referrer policy for image requests
+
+Plus Chakra UI style props and the following enhanced props:
+- `fit`: Object fit behavior (maps to `object-fit` CSS property)
+- `align`: Object position (maps to `object-position` CSS property)
+
+
 ## Common patterns
 
 ### Product image display


tylermorrisford (approved):
After all the rebasing I get why you'd want to slam that intl cleanup in here. Data-table is the only component I didn't see here - approving anywayz 🏆

valoriecarli:
> After all the rebasing I get why you'd want to slam that intl cleanup in here. Data-table is the only component I didn't see here - approving anywayz 🏆

@tylermorrisford the intl generation of the `combo-box` folder was driving me nuts (should just be `combobox`). i didn't realize initially that that naming convention pattern differs from the others.
Good catch on `DataTable`! This one isn't represented here as it's already split. Thank yyyooooouuu!

packages/nimbus/src/components/image/image.dev.mdx
```diff
@@ -328,20 +328,6 @@ export const Example = () => (
 
 <PropsTable id="Image" />
 
-The Image component accepts all standard HTML `<img>` attributes including:
```
stephsprinkle: Not needed?

packages/nimbus/src/components/number-input/number-input.mdx
```diff
@@ -20,6 +20,13 @@ A number input allows users to enter numerical values and adjust them incrementa
 
 A number input allows users to enter a numeric value and adjust it using stepper buttons. It provides a quick and precise way to modify values within defined parameters, especially for small incremental changes.
 
+
+### Key features
```
stephsprinkle: Same question here

packages/nimbus/src/components/text/text.dev.mdx
```diff
@@ -152,14 +152,6 @@ As a non-interactive display component, Text does not require keyboard interacti
 
 <PropsTable id="Text" />
 
-The Text component extends Chakra UI's Text component and accepts all style props including:
```
stephsprinkle: Should this be removed?

packages/nimbus/src/components/combobox/combobox.mdx
```diff
@@ -20,6 +20,15 @@ A combobox is an input that combines a text field with a pop-up list of options.
 
 The combobox component provides both selection and text input capabilities. It displays a list of options that users can choose from, while also allowing them to type directly into the text field. This functionality improves user experience by assisting with value selection and, in some cases, prevents invalid entries.
 
+### Key features
```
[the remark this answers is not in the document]
valoriecarli: @stephsprinkle Good question, and I actually have that in my scribbled notes to return to once we split everything. 
I found that this section was only used a handful of times. 
For `combobox` I only moved it to the same position as the others (as it made more sense above than buried at the bottom).
You will see I have a spot in the pr description for followups, ill add this to it.

packages/nimbus/src/components/text/text.dev.mdx
valoriecarli: As discussed, we will omit this for now.

PR #946, opened 2026-01-22
chore(nimbus): fix jest resolution using wrapper implementation

jaikamat:
## Summary

Jest's CommonJS resolver cannot handle Chakra UI subpath exports, causing test failures for  consumers. This fix aligns these 2 components with the pattern already used by 95% of Nimbus components (Box, Text, Heading, Stack, etc.).

After this is merged, I will once again cut a new release

### DIFF
--- a/packages/nimbus/src/components/flex/flex.tsx
+++ b/packages/nimbus/src/components/flex/flex.tsx
@@ -1,4 +1,4 @@
-import { Flex as ChakraFlex } from "@chakra-ui/react";
+import { Flex as ChakraFlex } from "@chakra-ui/react/flex";
 import type { FlexProps } from "./flex.types";
 
 /**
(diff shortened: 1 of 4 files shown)


tylermorrisford (approved):
Sensible.

packages/nimbus/src/components/flex/flex.tsx
```diff
@@ -1,8 +1,28 @@
+import { Flex as ChakraFlex } from "@chakra-ui/react";
```
ByronDWall: is there another way around this? We use the modular paths intentionally so that we do not bloat the consuming apps bundle with the full @chakra-ui/react package

see: https://chakra-ui.com/guides/component-bundle-optimization#modular-component-imports

ByronDWall (commented):
Is there another way?

PR #965, opened 2026-01-26
CRAFT-2052: Remove disabled state from Avatar component

misama-ct:
## Summary

- Remove `isDisabled` prop from Avatar component
- Remove `_disabled` styles from Avatar recipe
- Remove disabled state examples from documentation
- Update OpenSpec to reflect the removal
- Exclude `openspec/` folder from prettier formatting to preserve manual formatting in specification files

Avatar is a non-interactive leaf component (renders as `<figure>`) that should not have a disabled state. Disabled states are only meaningful for interactive elements. When Avatar is used within an interactive context (e.g., wrapped in a Button), the wrapper component provides appropriate disabled styling.

## Test plan

- [x] TypeScript compiles without errors
- [x] Package builds successfully
- [x] All Avatar story tests pass (6 tests)
- [x] OpenSpec validation passes


valoriecarli (approved):
🙌🏻

PR #978, opened 2026-01-28
chore(docs): update docs headers and nav titles to sentence case

ddouglasz:
This PR updates all component headers and nav titles in the docs to to sentence case. Eg `RangeCalendar` → `Range calendar`

### DIFF
--- a/packages/nimbus/src/patterns/fields/date-range-picker-field/date-range-picker-field.mdx
+++ b/packages/nimbus/src/patterns/fields/date-range-picker-field/date-range-picker-field.mdx
@@ -1,13 +1,14 @@
 ---
 id: Patterns-DateRangePickerField
-title: DateRangePickerField
+title: Date range picker field
+exportName: DateRangePickerField
 description: This pattern component integrates the DateRangePicker within a FormField using a simple, consistent, and opinionated API.
 documentState: InitialDraft
 order: 999
 menu:
   - Patterns
   - Fields
-  - DateRangePickerField
+  - Date range picker field
 tags:
   - component
   - pattern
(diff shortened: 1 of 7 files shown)


packages/nimbus/src/components/collapsible-motion/collapsible-motion.dev.mdx
```diff
@@ -1,5 +1,5 @@
 ---
-title: CollapsibleMotion Component
+title: Collapsible motion component
```
misama-ct: I copied the component-name from code into the search and it should bring up a search-result, but does not anymore:

Image

`title` was the only field where the full component name (as exported from nimbus) was mentioned, and now that it's changed, you can't find the component name you see in your code-editor anymore via copy & pasting it into the search.

---

I briefly mentioned this issue in the design system meeting and suggested an alternative new field (e.g. `displayName`) in the frontmatter-section for entering a custom display name but unfortunately this did not make it into the ticket. 

However, for a key that is called `title` It probably is better if it really _just_ is a title for the document and not a multi-purpose field which may or may not contain an implicit relation to an exported component, hook or function. So I am actually fine with this change, but ...

I still would like the search to return results when entering the react component-name only, so what about a new `exportName` field in the frontmatter schema that can be used for an explicit mapping instead (= "This document documents the following nimbus-export") and is discoverable by the search.

Regular docs (e.g. design tokens or adr's) just wouldn't have this key in their frontmatter-section. Having this field would also allow us to create a CI job that can reliably check if all exports have a corresponding mdx-file.

WDYT? @commercetools/craft-team-fe-tag-all 

```suggestion
title: Collapsible motion component
exportName: CollapsibleMotion
```

PR #1032, opened 2026-02-05
fix(docs): limit sidebar width

stephsprinkle:
#### Summary

- Constrained side nav to 200-400px (is this the correct width?)
- Limited character width of main content to 80 chars
- Main content grows with browser width, min width 80 chars

**Limitation:** this is not "responsive" in any way, is this something we want to consider in the future?


valoriecarli (approved):
> Constrained side nav to 200-400px (is this the correct width?)

my search tells me this is w/in the normal range, but i know someone who might know for certain...

PR #1039, opened 2026-02-06
CRAFT-2097: Resolve Table and Data table confusion by removing unused component

ddouglasz:
This PR removes the initially created Data table placeholder component from Nimbus in Favour of the actual Data table. Users tend to be confused if the Table component should be used as it tends to affect the visibility of the actual data table.

This PR also replaces parts of the docs where the initial Table component was used, with the Datatable Component.


stephsprinkle:
I thought we were keeping this one around. Perhaps it would be better to rename it as `SimpleTable`?

PR #1059, opened 2026-02-11
Interlink/improve documentation between the Table and DataTable Component

ddouglasz:
**Description**
Create a documentation page for the Table (HTML table) component. The goal is to provide clear implementation guidelines and establish a "directional bridge" between the simple and complex table components.

**Key Requirements**
Usage Guidelines: Define when to use a Plain Table (e.g., static data, small datasets, read-only content) vs. the Complex Data Table.

**Cross-Linking Strategy:**
In Plain Table Docs: Add a prominent "When to upgrade" section linking to the Complex Data Table for features like pagination, sorting, or server-side filtering.

In Complex Data Table Docs: Add a "Keep it simple" note linking back to the Plain Table for basic layouts to avoid unnecessary overhead.

**Technical Snippet**: 
Include a basic HTML/CSS example of the plain table structure.

### DIFF
--- a/packages/nimbus/src/components/data-table/data-table.types.ts
+++ b/packages/nimbus/src/components/data-table/data-table.types.ts
@@ -205,7 +205,7 @@ export type DataTableProps<T extends object = Record<string, unknown>> = Omit<
 /**Combined props for the TableHeader element (Chakra styles + Aria behavior). */
 export type DataTableHeaderProps<T extends DataTableColumnItem> =
   RaTableHeaderProps<T> &
-  DataTableHeaderSlotProps & { ref?: Ref<HTMLTableSectionElement> };
+    DataTableHeaderSlotProps & { ref?: Ref<HTMLTableSectionElement> };
 
 /** Combined props for the Column element (Chakra styles + Aria behavior). */
 export type DataTableColumnProps = RaColumnProps &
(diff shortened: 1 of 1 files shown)


packages/nimbus/src/components/data-table/data-table.dev.mdx
```diff
@@ -1147,6 +1189,12 @@ application-specific logic.
 
 ## Resources
 
+### Related components
+
+- **[Table Component](/components/data-display/table)** - Use the simpler Table component for static, read-only data without interactive features
+
+### External documentation
+
```
stephsprinkle: Can we leave these as a direct child of `Resources`?

stephsprinkle (commented):
Looking good! It looks like the `API reference` section for `Table` is wildly different from our standard, might be worth looking into.

packages/nimbus/src/components/data-table/data-table.dev.mdx
```diff
@@ -42,6 +42,48 @@ const App = () => {
...
+- Sorting columns in ascending or descending order
+- Row selection for bulk actions
+- Column management (show/hide, reorder, resize)
+- Search and filtering across data
+- Pagination for large datasets (> 50 rows)
+- Server-side data operations
+- Nested/hierarchical data with expansion
+- Dense data visualization with configurable row spacing
+
+**Example decision:**
```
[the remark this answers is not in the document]
ddouglasz: Good point, although the way I see it, I think other example on the page that captures bulk selection, row pinning, nested rows and so on is already an example of a complex use case, thats why I think just listing the feature already gives the user enough information.

packages/nimbus/src/components/data-table/data-table.dev.mdx
ddouglasz: I agree!

PR #1061, opened 2026-02-12
fix(button): CRAFT-2106 | fix double-firing event handlers

misama-ct:
## Summary

Fixes a bug where Button event handlers (`onClick`, `onPress`, `onFocus`, `onBlur`) fired twice per interaction. What started as a one-line root cause fix cascaded into a broader cleanup when I discovered that several downstream components had been unknowingly relying on the previous behavior.

Supersedes #1060.

## The bug

`contextProps` was merged into _both_ `useButton()` (which wraps handlers internally via `usePress`/`useFocusable`) _and_ the final `mergeProps()` call. Because `mergeProps` **chains** function props, every DOM event handler was invoked twice — once by the raw prop, once by the hook's wrapper.

## Initial plan

The fix was conceptually trivial: stop double-merging `contextProps`. Replace `mergeProps` with JSX spread ordering so that `useButton`'s processed handlers **overwrite** the raw ones (`{...contextProps} {...buttonProps}` — last spread wins) instead of chaining alongside them.

## Why I had to deviate

Applying the simple fix exposed three issues in downstream components that had been masked by the double-firing:

### 1. DataTable's capture-phase `pointerdown` handler was too aggressive

DataTable stops `pointerdown` propagation at the row level to prevent accidental selection. But the old logic stopped it for _all_ non-selection elements — including buttons. This prevented `usePress` from starting its press sequence on any `IconButton` inside a table row.

This only worked before because the double-merge kept a raw `onClick` alongside `usePress`'s handler. When `usePress` couldn't fire (pointerdown was stopped), the raw `onClick` still caught the native click event.

**Fix:** Simplified the capture handler to only stop propagation for non-interactive areas. Replaced the raw `DataTableExpandButton` slot with `IconButton`. Migrated internal buttons from `onClick` to `onPress`.

### 2. CollapsibleMotion passed `disabled` as an HTML attribute, but `useButton` overwrites it

`CollapsibleMotionTrigger` was passing `{ disabled: isDisabled }` via Chakra's `asChild` pattern. After fixing the spread order, `useButton`'s output (which doesn't know about the HTML `disabled` from context) would overwrite it.

I initially tried mapping `disabled` → `isDisabled` inside Button, then reverted that — it would have special-cased one HTML attribute without justification. Instead, I fixed it at the source: `CollapsibleMotionTrigger` now wraps children in a `ButtonContext.Provider` to pass `isDisabled` through React Aria's context, which `useButton` already knows how to consume.

### 3. Manual `aria-disabled` override conflicted with `useButton`'s element-aware logic

Button was manually setting `aria-disabled` regardless of element type. `useButton` already handles this correctly — native `<button>` gets `disabled`, non-native elements (`as="a"`) get `aria-disabled`. Removed the manual override, which required updating assertions in `localized-field` stories that expected `aria-disabled` on native buttons.

## What I ended up with

A Button component that properly delegates all behavior to `useButton`, with no manual overrides or workarounds:

```tsx
const { buttonProps, isPressed } = useButton({ ...contextProps, elementType }, contextRef);

<ButtonRoot
  {...contextProps}      {/* recipe props (size, variant) survive — useButton doesn't emit them */}
  {...buttonProps}       {/* hook's processed handlers overwrite raw ones via spread order */}
  as={as}
  asChild={asChild}
  data-pressed={isPressed || undefined}
>
```

Along the way, I also:
- Added `data-pressed` to the DOM — `useButton` already exposes `isPressed` but it wasn't being surfaced as a data attribute, so now it's available for styling via CSS selectors
- Added `@deprecated` JSDoc annotations to native HTML props (`onClick`, `disabled`, `tabIndex`, etc.) that have React Aria equivalents — this makes consumers' editors show a strike-through when using them, nudging towards `onPress`, `isDisabled`, `excludeFromTabOrder`, etc.
- Added the `sm` button size to the documentation examples
- Replaced hardcoded docs examples with `button.docs.spec.tsx` (auto-injected, tested)
- Added regression stories: `EventHandlersFireOnce`, `DOMPropFiltering`, `DisabledAsLink`

## Changes

**Core fix:**
- `button.tsx` — Replace `mergeProps` with JSX spread ordering, add `data-pressed`
- `button.types.ts` — Deprecate native HTML props with JSDoc, add React Aria equivalents
- `collapsible-motion-trigger.tsx` — Wrap `asChild` children in `ButtonContext.Provider`

**DataTable cleanup:**
- `data-table.row.tsx` — Simplify capture handler, replace expand button slot with `IconButton`, `onClick` → `onPress`
- `data-table.manager.tsx` — Settings button `onClick` → `onPress`
- `data-table.visible-columns-panel.tsx` — Reset button `onClick` → `onPress`
- `data-table.slots.tsx` / `data-table.recipe.ts` — Remove unused `expandButton` slot

**Tests & docs:**
- `button.stories.tsx` — Add `EventHandlersFireOnce`, `DOMPropFiltering`, `DisabledAsLink` stories
- `button.docs.spec.tsx` — New consumer implementation test suite
- `button.dev.mdx` — Auto-inject docs specs, add `sm` size example
- `localized-field.stories.tsx` — Fix disabled assertions (`aria-disabled` → `toBeDisabled()`)

## Test plan

- [x] `EventHandlersFireOnce` story: onClick, onPress, onFocus, onBlur, onKeyDown each fire exactly once
- [x] `DOMPropFiltering` story: React Aria props via context don't leak to DOM
- [x] `DisabledAsLink` story: non-native elements get `aria-disabled` + `role="button"`
- [x] Button stories: all pass
- [x] DataTable stories: all pass (row pinning, nested table, selection, clickable rows)
- [x] CollapsibleMotion stories: all pass
- [x] Localized-field stories: all pass
- [x] TypeScript: clean
- [x] Visual check: expand/pin buttons render at correct size in DataTable
- [x] Visual check: button variants (size, colorPalette) render correctly

CRAFT-2106


packages/nimbus/src/components/button/button.dev.mdx
misama-ct: The file was not aware of the recently added size, so I just added it. Apart from that the examples were hardcoded (as this was the first component that added the 4 tabs) so it needed to be updated.

packages/nimbus/src/components/button/button.types.ts
```diff
@@ -38,12 +38,62 @@ export type ButtonRootSlotProps = Omit<
...
+   * `aria-disabled` automatically based on `isDisabled`.
+   */
+  "aria-disabled"?: boolean | "true" | "false";
+
+  /**
+   * @deprecated Use `excludeFromTabOrder` instead — it provides a semantic
+   * boolean to remove the button from the tab order.
+   */
+  tabIndex?: number;
+};
```
misama-ct: Those types are native `html`-button attributes that overlap with functionality provided by `react-aria`. 

We don't use or demo native props anywhere, but consumers may _habitually_ use them and if they do, the deprecation messages will lead the way to a correct implementation.

ByronDWall: should we make this a type util and add it to all components that use react-aria under the hood for interactions?

ByronDWall (approved):
Overall this looks like a great change to me. I tested the data table manually and it seems to behave as expected, but the row clicking there is probably something to keep an eye on.

One question about using the deprecated type outside the button, but nothing that stops me from approving.

PR #1071, opened 2026-02-16
feat(nimbus): add automatic font loading to NimbusProvider

ddouglasz:
Nimbus uses Inter as font of choice, however, it does not provide a way to load it yet but rather relies on it being available. This is the case for the MC as long as ui-kit is running in parallel, however, if nimbus is used standalone, it will fall back to the next best available fallback font.

We need to find/provide a way to load the fonts. Do we have restrictions?

Acceptance criteria:

Fonts are provided by nimbus, in case they aren’t already in the application context

Resources

- See this [MC ADR](https://github.com/commercetools/merchant-center-frontend/blob/main/docs/architecture-decisions/0054-loading-fonts-from-google-fonts-css-api-v2.md#L5)

- The fonts themselves are loaded via a preconnect link that is declared in app-kit’s application.html [template](https://github.com/commercetools/merchant-center-application-kit/blob/e3e45dc8a35b2254a0d54a50b94682538681a488/packages/mc-html-template/html-docs/application.html#L20)


packages/nimbus/src/components/nimbus-provider/hooks/use-font-loader.ts
```diff
@@ -0,0 +1,91 @@
+import { useEffect } from "react";
+
+/**
+ * URL for Google Fonts CSS API v2 to load Inter font family.
+ * Loads weights 400, 500, 600, 700 with display=swap for optimal loading.
+ */
+const GOOGLE_FONTS_URL =
+  "https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&display=swap";
```
ByronDWall: I think we have 100-900 as design tokens

Image

packages/nimbus/src/components/nimbus-provider/hooks/use-font-loader.ts
misama-ct: I think instead of a hook, thanks to react 19, you can just create a very simple utility component using the code snippet provided by google directly without complicated runtime behavior. React will take care of hoisting and moves the link-tags automatically into the head-section of the app/page so that you don't have to worry about anything.

```jsx
const InterFontLoader = () => (
  <>
    <link rel="preconnect" href="https://fonts.googleapis.com">
    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
    <link href="https://fonts.googleapis.com/css2?family=Inter:wght@100..900&display=swap" rel="stylesheet">
  </>
)
```

In the `NimbusProvider` you just have to make sure that the component is only rendered if `loadFonts` is set to `true`.

```jsx
const NimbusProvider = ({loadFonts = false, ...props}) => (
    <>
      {loadFonts && <InterFontLoader/>}
      {/* ... stuff ... */}
    </>
)
```

PR #1084, opened 2026-02-17
fix(bundle): sets global github actions user for temp container

tylermorrisford:
This pull request makes a minor update to the GitHub Actions workflow configuration. The change ensures that the Git user name and email are set globally, which can help prevent issues with Git operations in CI environments.


valoriecarli (approved):

ByronDWall (approved):
this better work

PR #509, opened 2025-10-08
feat(tests): enhance testing framework and guidelines

misama-ct:
## Summary

This PR enhances Nimbus's testing infrastructure by introducing a dual testing strategy that clearly separates unit tests for pure functions from Storybook interaction tests for UI components.

### Two Distinct Testing Approaches

1. **Unit Tests (Vitest)** - Fast, isolated tests for pure functions
2. **Storybook Tests** - Browser-based tests for UI components and interactions

### When to Use Each

**Unit Tests (`*.test.ts`) are for:**
- Pure functions with deterministic outputs
- Data transformations (parsers, formatters, converters)
- Validation logic (type guards, emptiness checks)
- Calculations (financial math, sorting algorithms)
- Security utilities (sanitization, XSS prevention)
- String manipulation (locale sorting, text processing)

**Storybook Tests (`*.stories.tsx` with play functions) are for:**
- UI interactions (clicks, typing, keyboard navigation)
- Component behavior (state changes, rendering)
- Visual regressions (screenshots, layout)
- Accessibility (ARIA attributes, screen reader support)
- Integration (component composition, context providers)

### New Configuration Files

- **`vitest.unit.config.ts`** - Dedicated config for fast unit tests
  - No browser environment overhead
  - Excludes `*.stories.tsx` files
  - Optimized for pure function testing
  - Coverage reporting for utilities

- **`vitest.storybook.config.ts`** - Renamed from `vitest.config.ts`
  - Browser-based testing with Playwright
  - Tests component stories with play functions
  - Chromium headless mode for CI

### Updated Commands

```bash
# Run fast unit tests (new)
pnpm test:unit

# Run unit tests in watch mode (new, for TDD)
pnpm test:unit:watch

# Generate coverage report (new)
pnpm test:coverage

# Run Storybook tests (existing)
pnpm test:storybook

# Run all tests sequentially (new)
pnpm test:all

# Run all tests (parallel, existing)
pnpm test
```

### New Guidelines Document

Added comprehensive **Unit Testing Guidelines** (`docs/file-type-guidelines/unit-tests.md`) covering:

- Clear decision matrix for when to use unit tests vs Storybook tests
- Test structure and organization patterns
- AAA pattern (Arrange, Act, Assert)
- Edge case and error condition testing
- Test data management strategies
- Vitest-specific features and matchers
- Performance testing guidelines
- Security testing patterns
- Best practices and common pitfalls

### Updated Documentation

[...]

PR #1094, opened 2026-02-18
chore(deps): major dependency upgrades (eslint v10, glob v13, vite-tsconfig-paths v6, jsdom v28)

misama-ct:
## Summary

Major dependency version upgrades:

- **eslint** `^9.39.2` → `^10.0.0`
- **@eslint/js** `^9.39.2` → `^10.0.1`
- **glob** `^11.0.3` → `^13.0.5`
- **vite-tsconfig-paths** `^5.1.4` → `^6.1.1`
- **jsdom** `^27.4.0` → `^28.1.0`

### Additional cleanup

- Remove `eslint-plugin-react-hooks` from catalog and app dependencies — was never wired into `eslint.config.mjs`. When I attempted to configure it, the v7 `recommended` preset (which now includes 17 React Compiler rules) produced 90 violations, many of which are false positives for React Aria patterns. Proper reintroduction tracked in [CRAFT-2117](https://commercetools.atlassian.net/browse/CRAFT-2117).
- Remove unused `@types/jsdom` (only declares module types for direct imports — jsdom is only used as a Vitest environment string)
- Fix `no-useless-assignment` violations now caught by `@eslint/js` v10 recommended rules

## Notes

- One remaining peer dep warning from `@typescript-eslint/utils` (transitive dep of `eslint-plugin-storybook`) — cosmetic, will resolve when they publish an update
- ESLint 10 requires Node `^20.19.0 || ^22.13.0 || >=24`
- Glob v12/v13 breaking changes are CLI-only (shell option removal, CLI moved to separate package) — no API changes affect the codebase
- vite-tsconfig-paths v6 major bump was internal refactoring only — no breaking API changes
- jsdom v28 breaking change is a resource loading API overhaul — doesn't affect us since jsdom is only used as a Vitest test environment

## Test plan

- [x] `pnpm lint` passes cleanly
- [x] `pnpm typecheck` passes
- [x] `pnpm test` passes
- [x] CI passes

[CRAFT-2117]: https://commercetools.atlassian.net/browse/CRAFT-2117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/nimbus/package.json
+++ b/packages/nimbus/package.json
@@ -74,7 +74,6 @@
     "@testing-library/user-event": "catalog:tooling",
     "@types/escape-html": "^1.0.4",
     "@types/is-hotkey": "^0.1.10",
-    "@types/jsdom": "catalog:tooling",
     "@types/react": "catalog:react",
     "@types/react-dom": "catalog:react",
     "@vitejs/plugin-react": "catalog:tooling",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -86,9 +86,6 @@ catalogs:
     '@testing-library/user-event':
       specifier: ^14.6.1
       version: 14.6.1
-    '@types/jsdom':
-      specifier: ^27.0.0
-      version: 27.0.0
     '@vitejs/plugin-react':
       specifier: ^5.1.4
       version: 5.1.4
@@ -621,9 +618,6 @@ importers:
       '@types/is-hotkey':
         specifier: ^0.1.10
         version: 0.1.10
-      '@types/jsdom':
-        specifier: catalog:tooling
-        version: 27.0.0
       '@types/react':
         specifier: catalog:react
         version: 19.2.14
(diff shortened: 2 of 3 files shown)


packages/nimbus/src/components/combobox/utils/filters.spec.ts
```diff
@@ -478,7 +478,7 @@ describe("filter utilities", () => {
           const text = node.textValue.toLowerCase();
           const lowerInput = input.toLowerCase();
 
-          let score = 0;
+          let score;
```
misama-ct: `@eslint/js` v10 expanded the detection of the `no-useless-assignment` rule in `eslint:recommended`. The initial `= 0` is useless here because every branch of the if/else either assigns a new value to `score` or returns early. This was not caught by v9 but is now flagged by v10. No behavioral change — just removing the dead initializer.

packages/nimbus/src/components/data-table/utils/rows.utils.tsx
```diff
@@ -87,16 +87,14 @@ export function sortRows<T extends object>(
...
-          bSortValue = String(bValue).toLowerCase();
-        }
+        const aSortValue =
+          typeof aValue === "number" && typeof bValue === "number"
+            ? aValue
+            : String(aValue).toLowerCase();
+        const bSortValue =
+          typeof aValue === "number" && typeof bValue === "number"
+            ? bValue
+            : String(bValue).toLowerCase();
```
misama-ct: Same `no-useless-assignment` rule from `@eslint/js` v10. The original code assigned `aValue`/`bValue` to `let` variables, then immediately overwrote them in every branch of the if/else (the number branch even re-assigned the same value). Refactored to `const` with ternary — cleaner and satisfies the lint rule. No behavioral change.

packages/nimbus/package.json
```diff
@@ -74,7 +74,6 @@
     "@testing-library/user-event": "catalog:tooling",
     "@types/escape-html": "^1.0.4",
     "@types/is-hotkey": "^0.1.10",
-    "@types/jsdom": "catalog:tooling",
     "@types/react": "catalog:react",
```
misama-ct: `@types/jsdom` was added when the unit testing infrastructure was set up (PR #509), as a precautionary companion to the `jsdom` dependency. However, it only declares types for `import from "jsdom"` (the JSDOM constructor etc.) — it provides no global type augmentations.

Since jsdom is exclusively used as a Vitest test environment via the string `environment: "jsdom"` and is never imported directly, these types were never actually needed.

PR #1228, opened 2026-03-10
fix(rich-text): adjust flaky test

tylermorrisford:
### Summary
- Fix flaky OnChangeCallback story test for RichTextInput that intermittently fails in CI with "Change count: 0"     
- Root cause: userEvent.type was firing before the editor finished initializing after focus, causing keystrokes to be
  lost
- Wait for the editor to signal readiness (first onChange) before typing, and assert a minimum change count instead of
  an exact value to tolerate editor lifecycle timing differences

Test plan

- Verify OnChangeCallback story test passes locally
- Confirm no more flaky failures in CI across multiple runs

### DIFF
--- a/packages/nimbus/src/components/rich-text-input/rich-text-input.stories.tsx
+++ b/packages/nimbus/src/components/rich-text-input/rich-text-input.stories.tsx
@@ -839,7 +839,9 @@ export const OnChangeCallback: Story = {
     const canvas = within(canvasElement);
     const editor = canvas.getByRole("textbox");
 
-    // Focus the editor and wait for it to be ready before typing
+    // Focus the editor and wait for it to be ready before typing.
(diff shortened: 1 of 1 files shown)


packages/nimbus/src/components/rich-text-input/rich-text-input.stories.tsx
```diff
@@ -854,6 +856,8 @@ export const OnChangeCallback: Story = {
       const changeCountElement = canvas.getByText(/Change count: \d+/);
       const count = Number(changeCountElement.textContent?.match(/\d+/)?.[0]);
       expect(count).toBeGreaterThanOrEqual(5);
+      // Guard against runaway onChange loops (a known ProseMirror failure mode)
+      expect(count).toBeLessThan(50);
```
valoriecarli: 🙌🏻

PR #1359, opened 2026-04-13
refactor(card): rework Card component architecture and API

misama-ct:
> [!NOTE]
> Card was one of the early components and we made a couple of questionable decisions. This PR addresses them. There are breaking changes, but fortunately no consumers yet.

### Architecture: direct rendering replaces context registration

Card.Header and Card.Content previously returned `null` and used `useEffect` to register JSX into parent state via context callbacks. This pattern — unique among all Nimbus compound components — introduced re-render risk, stale closures, and a double render cycle.

Now Header, Body, and Footer are simple passthrough components that render their slot directly, matching Dialog.Body, Dialog.Footer, and every other Nimbus compound component.

### Naming: Content → Body, Footer added

- `Card.Content` renamed to `Card.Body` (aligns with Dialog.Body, Drawer.Body convention)
- `Card.Footer` added (matches Dialog, DefaultPage, and every major UI library)

### Props: four visual props → standard `variant` + `size`

`cardPadding`, `borderStyle`, `elevation`, and `backgroundStyle` are replaced by the standard Nimbus `variant` and `size` props.

**`size`** (sm | md | lg, default: md) — sets `--card-spacing` CSS variable:

| Size | `--card-spacing` |
|------|-----------------|
| sm | `spacing.300` |
| **md** (default) | `spacing.400` |
| lg | `spacing.600` |

**`variant`** (default: outlined) — eight kebab-case names enumerating every combination of three independent visual axes that designers expose in Figma: `outlined` (border), `elevated` (shadow), `muted` (background). Names list each enabled axis in the fixed order `outlined-elevated-muted`, joined with `-`; the all-off case is `plain`.

| Variant | Border | Shadow | Background |
|---------|:------:|:------:|:----------:|
| `plain` |  |  | default |
| **`outlined`** _(default)_ | ✓ |  | default |
| `elevated` |  | ✓ | default |
| `outlined-elevated` | ✓ | ✓ | default |
| `muted` |  |  | muted |
| `outlined-muted` | ✓ |  | muted |
| `elevated-muted` |  | ✓ | muted |
| `outlined-elevated-muted` | ✓ | ✓ | muted |

The previous `filled` variant has been renamed to `muted` to match the design-token vocabulary. The kebab convention is intentional: this is the first Nimbus component to expose a variant matrix wide enough that camelCase names (e.g. `outlinedElevatedMuted`) become hard to read; a fixed axis order means consumers never have to guess between `elevated-outlined` and `outlined-elevated`.

### Accessibility: slot-based ARIA wiring

Card.Root uses React Aria's `useSlotId` + `Provider` pattern to automatically wire `aria-labelledby` and `aria-describedby` when `<Heading slot="title">` or `<Text slot="description">` are present. The card itself stays a plain `<div>` — no implicit `role` is applied; consumers set `role` explicitly (e.g. `role="article"`) when a landmark role is needed. This avoids semantic conflicts when a Card is embedded inside an interactive wrapper (`<button>`, `<a>`).

### Layout: `inline-flex` → `flex`

Cards are now block-level containers that fill available width by default.

### Free-form layout: padding-safe `Card.Root` without slots

When `Card.Root` is used without any `Card.Header` / `Card.Body` / `Card.Footer` slot children, padding lives on the Root itself (via a `:has()` selector) so arbitrary children never render flush against the border. The slot-based path is unchanged: when slots are present, they own their padding and Root stays unpadded.

---

## Consumer API

```tsx
<Card.Root variant="outlined-elevated" size="md">
  <Card.Header>Product details</Card.Header>
  <Card.Body>Description and metadata</Card.Body>
  <Card.Footer>Action buttons</Card.Footer>
</Card.Root>
```

## Files

- `openspec/changes/rework-card-component/` — proposal, spec delta, and task list
- `packages/nimbus/src/components/card/` — full implementation, stories, docs spec, MDX docs
- `.claude/skills/nimbus-code-connect/` — skill override updated so future regeneration of `card.figma.tsx` reproduces the new variant composition

## Test plan

- [ ] `pnpm --filter @commercetools/nimbus typecheck` passes
- [ ] `pnpm lint` passes
- [ ] `pnpm test:dev packages/nimbus/src/components/card/card.stories.tsx` — all stories pass (8 variant cells assert in the matrix story)
- [ ] `pnpm test:dev packages/nimbus/src/components/card/card.docs.spec.tsx` — all doc specs pass
- [ ] Visual review in Storybook: 4×2 variant matrix, all sizes, part combinations, separator spacing, slot-based accessibility, free-form (no-slots) padding


misama-ct:
@claude do a thorough review of this branches changes

misama-ct:
@claude please review

.changeset/rework-card-component.md
```diff
@@ -0,0 +1,34 @@
+---
+"@commercetools/nimbus": minor
```
ByronDWall: This is a breaking component API change. Should this be a major release?

packages/nimbus/src/components/card/card.recipe.ts
```diff
@@ -5,59 +5,85 @@ import { defineSlotRecipe } from "@chakra-ui/react/styled-system";
...
   className: "nimbus-card",
 
   base: {
     root: {
       colorPalette: "slate",
-      display: "inline-flex",
+      display: "flex",
+      flexDirection: "column",
       alignItems: "flex-start",
       borderRadius: "300",
```
ByronDWall, line 18: So, I think that having direct text children in `Card.Root` without `Card.Body` results in different styling behavior than the previous `Card.Root`. I'm not sure if this is a regression or not.

What if `Card.Root` gets padding if the child components aren't rendered?

```suggestion
      borderRadius: "300",
      // Apply padding directly when no slot children provide their own.
      "&:not(:has(.nimbus-card__header, .nimbus-card__body, .nimbus-card__footer))":
        {
          p: "var(--card-spacing)",
        },
```

Screenshots of original and updated storybook for comparison:

Image

Image

ByronDWall (approved):
I think this is a definite improvement of the `Card`, and overall looks good. One question on padding of `Card.Root` if the header/body/footer aren't used, but it's genuinely a question.

The only thing that I think really needs to change is upping the release version to a major release, since this is a breaking api change for the card.

.changeset/rework-card-component.md
misama-ct: Technically, yes. However, I did go through the MC repos we take care of and couldn't find anyone using `Card`, so I'd suggest to just slip it in as a minor update. Do you want to do a quick search on your own? I mean I'm also fine with a v3 release, I was just seeing major releases for more meaningful changes 😄 

```
I scanned every @commercetools/nimbus import across all 11 maintained repos.

No one is using Card from Nimbus.

The only nimbus consumer today is merchant-center-frontend — almost entirely the agentic-channels feature 
under packages-application/application-project-settings, plus application-orders/src/entry-point.tsx (just NimbusProvider).
Those files pull in 30+ Nimbus components (Box, Stack, Flex, Grid, Heading, Text, Button, Dialog, FormField,
the various Input fields, Accordion, Alert, Badge, Tabs, Table, DataTable, Pagination, Tooltip, Switch, ToggleButtonGroup, etc.)
— but Card is not in the imported set of any of them. 
  ```

misama-ct: We could make it a formally correct minor change by keeping the previous props alive, mapping them to the new ones and Annotate all of them as `@deprecated`.

1. Provide `Card.Body` component under `Card.Content` namespace
2. `cardPadding` → `size` (direct rename, value mapping is identity: `sm|md|lg`)                                                   
4. More complicated:  `borderStyle` + `elevation` + `backgroundStyle` combinations → `variant`  

(ugly + painful, I'd rather do a v3 release)

ByronDWall: Id much rather just do a v3 release

tylermorrisford (approved):

packages/nimbus/src/components/card/card.recipe.ts
misama-ct: (oh, and this one has been taken care of as well.)

PR #1425, opened 2026-04-28
fix(Separator): missing `orientation` prop

ragafus:
The `orientation` prop is documented in https://nimbus-documentation.vercel.app/components/navigation/toolbar/guidelines#toolbar-in-use but when using it on another repo, I get a TS error.


ragafus:
Postinstall script was not running in my local env. Fixed https://nimbus-documentation.vercel.app/home/getting-started/installation#typescript-support

PR #1411, opened 2026-04-23
fix(Avatar): avoid runtime errors with edge `userFirstName` and `userLastName` values

ragafus:
…

PR #1434, opened 2026-04-29
feat(avatar): handle missing/edge-case firstName and lastName values

misama-ct:
## Summary

- Relax `Avatar`'s `firstName` and `lastName` to optional, add a `Person` icon fallback when neither yields a usable initial, and make initials extraction trim-aware and Unicode codepoint-safe.
- Adds a new localized `avatarLabelGeneric` aria-label for the no-name case (`"User avatar"` in English; English fallback for other locales until Transifex syncs).
- Side fix: patch `packages/i18n/scripts/generate-dictionaries.ts` so message keys that follow an interpolated message (one using `\${args.foo}`) are no longer silently dropped from the generated `*MessageKey` union type. This incidentally restores 2 keys (`page`, `pagination`) for `Pagination`.

Resolves the same underlying concern as #1411 with a fuller behavior contract and a pre-locked spec.

## Background

Per investigation in this conversation, the existing `getInitials` crashes on empty strings (`"".split("")[0]` is `undefined`, then `.toUpperCase()` throws). PR #1411 narrowed that to `charAt(0)`, fixing only the empty-string crash. The Merchant Center has user records where one or both name fields are `undefined`, empty, or whitespace-only — and the type contract claimed they were required strings, masking the mismatch.

This change locks the full behavior matrix in spec form first (see OpenSpec change `harden-avatar-edge-case-names`), then drives the implementation via TDD.

## Behavior matrix

| Input | Visible | aria-label |
|---|---|---|
| `firstName="John"`, `lastName="Doe"` | `JD` | "Avatar image for John Doe" |
| ` " John "`, `"Doe"` | `JD` (trimmed) | "Avatar image for John Doe" |
| `"John"`, missing | `J` | "Avatar image for John" |
| missing, `"Doe"` | `D` | "Avatar image for Doe" |
| both missing/empty/whitespace | `<Person />` icon | "User avatar" (generic) |
| `"👨"`, `"Doe"` | `👨D` (codepoint-safe) | full name |

## OpenSpec

Proposal artifacts live at `openspec/changes/harden-avatar-edge-case-names/`:

- [`proposal.md`](./openspec/changes/harden-avatar-edge-case-names/proposal.md)
- [`design.md`](./openspec/changes/harden-avatar-edge-case-names/design.md)
- [`specs/nimbus-avatar/spec.md`](./openspec/changes/harden-avatar-edge-case-names/specs/nimbus-avatar/spec.md) — delta against the existing `nimbus-avatar` capability
- [`tasks.md`](./openspec/changes/harden-avatar-edge-case-names/tasks.md)

## Test plan

- [x] `pnpm test:dev packages/nimbus/src/components/avatar/` — 37/37 against source
- [x] `pnpm --filter @commercetools/nimbus build && pnpm test packages/nimbus/src/components/avatar/` — 39/39 against built bundle
- [x] `pnpm --filter @commercetools/nimbus typecheck` — clean for avatar (only pre-existing baseline errors in `toast.manager.ts` and `aspect-ratios.ts` remain, unchanged from `main`)
- [x] `pnpm lint` — clean
- [ ] Visual check across `2xs`/`xs`/`md` plus light/dark modes via `pnpm start:storybook`
- [ ] After merge: comment on #1411 closing it with a pointer to this change

## Notes for reviewers

- Type relaxation (`string` → `string | undefined`) is a contract-tightening change for consumers who relied on TypeScript loopholes — actually a correctness win, but may surface latent bugs at call sites that previously coerced `undefined` to `"undefined"`. Documented in `avatar.dev.mdx`.
- `Person` icon sized via `"& > svg": { width: "70%", height: "70%" }` in the avatar recipe; only matches the direct-child icon, not nested SVGs from other slots.
- New `avatarLabelGeneric` i18n key seeds with the English default in all locale data files; Transifex will pick up real translations on the next sync.

### DIFF
--- a/packages/i18n/scripts/generate-dictionaries.ts
+++ b/packages/i18n/scripts/generate-dictionaries.ts
@@ -73,7 +73,11 @@ function dirToVariableName(dirName: string): string {
  */
 async function getMessageKeys(localePath: string): Promise<string[] | null> {
   try {
-    const content = await fs.readFile(localePath, "utf-8");
+    const rawContent = await fs.readFile(localePath, "utf-8");
+    // Strip template-literal interpolations like `${args.name}` first, so
+    // their inner `}` doesn't prematurely terminate the [^}]+ match below
+    // and silently drop later message keys.
+    const content = rawContent.replace(/\$\{[^}]*\}/g, "");
     // Extract keys from export default { "key": ... } or { key: ... }
     const match = content.match(/export default\s*\{([^}]+)\}/s);
     if (!match) return null;
--- a/packages/nimbus/src/components/avatar/utils/get-initials.spec.ts
+++ b/packages/nimbus/src/components/avatar/utils/get-initials.spec.ts
@@ -1,14 +1,5 @@
-/**
- * Unit tests for Avatar's pure helpers.
- *
- * Component behavior (rendered DOM, image fallback, icon fallback) is
- * covered by avatar.stories.tsx play functions. These JSDOM-fast tests
- * exercise the initials extraction edge-case matrix that would otherwise
- * require many browser-test permutations.
- */
-
 import { describe, it, expect } from "vitest";
-import { getInitials, getFullName } from "./avatar";
+import { getInitials } from "./get-initials";
 
 describe("getInitials", () => {
   it("returns uppercase first character of each name", () => {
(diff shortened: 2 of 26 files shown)


packages/i18n/scripts/generate-dictionaries.ts
```diff
@@ -73,7 +73,11 @@ function dirToVariableName(dirName: string): string {
  */
 async function getMessageKeys(localePath: string): Promise<string[] | null> {
   try {
-    const content = await fs.readFile(localePath, "utf-8");
+    const rawContent = await fs.readFile(localePath, "utf-8");
+    // Strip template-literal interpolations like `${args.name}` first, so
+    // their inner `}` doesn't prematurely terminate the [^}]+ match below
+    // and silently drop later message keys.
+    const content = rawContent.replace(/\$\{[^}]*\}/g, "");
```
tylermorrisford, line 80: lovely!

packages/nimbus/src/components/avatar/utils/get-initials.spec.ts
tylermorrisford, line 1: nit: aren't these helpers [heavy quotes] "_theoretically_" tested in the stories? Only mentioning because other components may have helpers that aren't individually tested. Between the docs.spec.tsx file and the stories do we have enough coverage?

tylermorrisford (approved):
approving with one tests question, because it is magnificent

packages/nimbus/src/components/avatar/utils/get-initials.spec.ts
ByronDWall: nit: explicitly testing the helpers is good, actually

ByronDWall (approved):
This change makes sense and lgtm!

packages/nimbus/src/components/avatar/utils/get-initials.spec.ts
misama-ct: > nit: aren't these helpers [heavy quotes] "theoretically" tested in the stories? 

> nit: explicitly testing the helpers is good, actually

Exactly, if another component would want to reuse the  `getInitials` function in the future we'd have to lift the function up into the global `utils` folder but have no explicit tests for them.

What went wrong though (and none of us noticed): those util functions should not have ended up within the component file but [should have rather been pulled into their own files](https://github.com/commercetools/nimbus/blob/main/docs/file-type-guidelines/utils-and-constants.md) within  `avatar/utils/`.  (I just did that change)

This structure is a bit verbose but also allows easy recycling. Sticking to the initial example, assuming you'd need the `getInitials` functions in another component later you can safely move function & test from the `avatar/utils` folder into the global `utils` folder, change the import in avatar, add an import to your new component and be done.

PR #1436, opened 2026-04-29
fix: notify tx team workflow

tylermorrisford:
Fixes an issue where gh CLI falls back to git remote to identify the repo, which fails with "not a git repository" when no checkout step has run. 

Adding --repo ${{ github.repository }} to both gh pr view and gh pr comment makes them target the repo explicitly


misama-ct (approved):
I APPROVE

ByronDWall (approved):

PR #1443, opened 2026-04-30
chore(claude): wire utils-and-constants into validation pipeline

misama-ct:
## Summary

Codifies the convention that pure helpers MUST live in `utils/{kebab-name}.ts` (component-scoped or global) rather than being inlined into the main component file, and wires the rule into the agent validation pipeline so it is enforced at write-time and review-time, not just documented. Also defines a mechanical merge rule for when sibling helpers belong in the same family file vs. their own files.

This change is in response to a real review miss: a recent Avatar PR inlined `getInitials` and `getFullName` in `avatar.tsx`. The convention was already documented in `docs/file-type-guidelines/utils-and-constants.md`, but no validation gate enforced it — the reviewer agent's file-type table had no row for `utils/`, and the `writing-main-component` checklist had no "no utility exports" item, so the inlining slipped through both write and review phases.

### What changes

**New skill** — `writing-utils-and-constants`
- Create / update / validate modes
- Validation checklist across 6 categories (location & inlining, file organization, purity, documentation, type safety, testing)
- Templates for util / spec / constant files and an Update Mode scenario for "move inlined helpers out of main component"

**Agent prompts** — `nimbus-coder` and `nimbus-reviewer`
- Add `utils/*.ts` row to the File Type → Skill mapping tables
- Disambiguate the "YOU code directly" guidance in `nimbus-coder` so utils cannot be misread as belonging in `{component}.tsx`
- Replace remaining "one function per file" guidance with a reference to the merge rule

**Existing skill** — `writing-main-component`
- Adds Category 1 item 7: "No utility exports" — flags any `export function` / `export const` in `{component}.tsx` whose return is not JSX

**Docs** — `main-component.md`, `utils-and-constants.md`, `component-guidelines.md`, `component-templates/index.md`, `barrel-exports.md`
- New Critical Rules subsection in `main-component.md` ("Pure Helpers Live in `utils/`, Not in the Main File") with a positive code example and cross-link
- New top-level **Structural Parity** section in `utils-and-constants.md` documenting parity between `packages/nimbus/src/utils/` and `components/{name}/utils/` so promotable helpers can move with `git mv`
- New top-level **File Organization** section in `utils-and-constants.md` defining the mechanical merge rule for solo vs. family files:
  - **Criterion 1**: shared content token ≥4 chars (after dropping connectors `By`/`For`/`From`/`To`/`With` and generic verbs `get`/`is`/`has`/`set`/`add`/`do`/`make`/`build`/`create`/`extract`/`find`/`try`)
  - **Criterion 2**: factory whose return signature matches a sibling
  - **Criterion 3**: direct dependency between siblings
  - Merged file is named for the shared token in agent-noun pluralized form (`filters.ts`, `validators.ts`); banned filenames: `helpers.ts`, `utils.ts`, `misc.ts`
  - Coupling rule prevents mixing promotable and component-coupled helpers in the same file
  - Forward-applied only — no retroactive sweep; barrel keeps file-level renames non-breaking
- Updated checklist in `utils-and-constants.md`: **one purpose per file** (per the merge rule), sibling `*.spec.ts` mandatory, `index.ts` barrel
- Clarified "When Utils/Constants Aren't Needed" so non-exported function-internal `const`s are explicitly acceptable
- Replaced banned `helpers.ts` / suffix-style examples in `component-guidelines.md`, `component-templates/index.md`, and `barrel-exports.md` with rule-conformant filenames

### Out of scope

Existing components that still inline pure helpers, or that use the older `helpers.ts` style, are not migrated in this PR — the merge rule is forward-applied only. The Avatar component was already migrated on its own branch as part of the original Avatar review.

## Test plan

- [ ] CI passes (no code changes — only docs/skills/agent prompts)
- [ ] Spot-check: open the new skill in `.claude/skills/writing-utils-and-constants/SKILL.md` and confirm the validation checklist reads cleanly
- [ ] Spot-check: confirm `docs/file-type-guidelines/main-component.md` Critical Rules section now lists the new rule as #3 and DisplayName Convention is renumbered to #4
- [ ] Spot-check: confirm `docs/file-type-guidelines/utils-and-constants.md` File Organization section's merge-rule examples table renders cleanly
- [ ] Verify the `writing-utils-and-constants` skill is discoverable (appears in the available-skills list when Claude Code starts)

### DIFF
--- a/.claude/skills/writing-utils-and-constants/SKILL.md
+++ b/.claude/skills/writing-utils-and-constants/SKILL.md
@@ -1,0 +1,338 @@
+---
+description: Create, update, or validate component-scoped utility functions and constants
+argument-hint: create|update|validate ComponentName [details]
+---
+
+# Writing Utils and Constants Skill
+
+You are a Nimbus utilities specialist. This skill helps you create, update, or
+validate component-scoped pure helper functions in `utils/` and constants in
+`constants/`.
+
+## Critical Requirements
+
+**Pure helper functions MUST NOT be inlined into the main component file
+(`{component}.tsx`).** Every component-scoped utility lives in `utils/` with
+its own file, its own spec, and is re-exported through a barrel.
+
+Patterns enforced:
+
+- One function per file (kebab-case filename matches the function name)
+- Sibling `*.spec.ts` per util (Vitest, JSDOM)
+- `utils/index.ts` barrel re-exporting every util
+- Pure functions only — no React, no JSX, no side effects
+- JSDoc on every export
+- `as const` for constants
+
+---
+
+## Mode Detection
+
+Parse the request to determine the operation:
+
+- **create** — Generate new utility/constant files for a component
+- **update** — Modify existing utilities or move inlined helpers into `utils/`
+- **validate** — Check compliance with guidelines (**PRIMARY USE CASE**)
+
+If no mode is specified, default to **validate**.
+
+---
+
+## Required Research (All Modes)
+
+Before any operation, you MUST read in parallel:
+
+1. **Utils and constants guidelines**:
+   `docs/file-type-guidelines/utils-and-constants.md`
+2. **Unit testing strategy**:
+   `docs/file-type-guidelines/unit-testing.md`
+3. **Existing patterns** (study at least one):
+   - `packages/nimbus/src/components/combobox/utils/` — multi-file utils with
+     specs (`filters.ts` + `filters.spec.ts`, `selection.ts` +
+     `selection.spec.ts`, `collection.ts` + `collection.spec.ts`)
+   - `packages/nimbus/src/components/inline-svg/utils/sanitize-svg.ts` — single
+     util with sibling spec
+   - `packages/nimbus/src/components/money-input/utils/` — utils plus an index
+     barrel re-exporting multiple helpers
+
+---
+
+## File Structure
+
+```
+packages/nimbus/src/components/{component}/
+├── {component}.tsx
+├── utils/
+│   ├── {kebab-name}.ts          # one function per file
+│   ├── {kebab-name}.spec.ts     # sibling unit test
+│   └── index.ts                 # barrel re-export
+├── constants/                   # only when constants are needed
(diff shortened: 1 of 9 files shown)


tylermorrisford (approved):

valoriecarli (approved):
noice, thanks for doing this!

.claude/skills/writing-utils-and-constants/SKILL.md
```diff
@@ -0,0 +1,338 @@
...
+
+## Critical Requirements
+
+**Pure helper functions MUST NOT be inlined into the main component file
+(`{component}.tsx`).** Every component-scoped utility lives in `utils/` with
+its own file, its own spec, and is re-exported through a barrel.
+
+Patterns enforced:
+
+- One function per file (kebab-case filename matches the function name)
```
ByronDWall: One file per method is a little more strict than I realized. I'm fine with it in theory, but up until now I think we've been colocating util methods in files in the `/utils` folder (e.g. combobox's utils).

If you'd like to enforce this, I think we should make sure that the repo follows the convention consistently in a follow up PR.

ByronDWall (approved):
This is fine as long as we go through and make sure the repo is following these rules, as they are a bit more strict than what is currently being followed in most components.

.claude/skills/writing-utils-and-constants/SKILL.md
misama-ct: I had the same ick but couldn't pin down a sensical phrasing, so I left the strict version as start. But since you have a similar feeling, I did a couple of adjustments, essentially:

- One purpose per file, not one function
- Solo files (get-initials.ts) and family files (filters.ts)
                                                            
This also means there is no retroactive sweep needed. The rule fires when a `utils/` folder is touched, and the barrel keeps file renames non-breaking. Claude invented a mechanical strategy for determining filenames and grouping, let's see if that one works out.

ByronDWall: perfect, thanks for making the update!

PR #1451, opened 2026-05-04
chore(vite): upgrade to Vite 8 + @vitejs/plugin-react 6

misama-ct:
## Summary

Upgrades the workspace from **Vite 7.3.3 → 8.0.11** and **@vitejs/plugin-react 5.2.0 → 6.0.1** (required — v6 is the first version with Vite 8 support). Also bumps `vite-plugin-dts 4.5.4 → 5.0.0` along the way.

## Why two breaking-change adaptations were needed

Vite 8 swaps **Rollup → Rolldown** and **esbuild → Oxc**. Rolldown's design preserves CommonJS semantics rather than transforming them like `@rollup/plugin-commonjs` did, so the existing nimbus library config produced a bundle that would have crashed for downstream browser consumers. Two surgical changes fix it:

### 1) Externalize React via `esmExternalRequirePlugin`, not `rollupOptions.external`

`packages/nimbus/vite.config.ts`

Under Rolldown, listing `react`/`react-dom` in `rollupOptions.external` makes them external **but** leaves `require("react")` calls inside bundled CJS deps (`use-sync-external-store/cjs/*`) emitting a runtime `__require()` helper that throws in browsers (`Calling 'require' for "react" in an environment that doesn't expose the 'require' function`). The fix is to declare the React-family externals through `esmExternalRequirePlugin` instead — the plugin both externalizes them and rewrites internal `require(externalId)` → ESM imports.

> Caveat: the two paths are mutually exclusive. If an ID appears in both `rollupOptions.external` and the plugin's `external` list, the plain externalization wins and the plugin's `require()` rewrite never runs. See [vitejs/rolldown-vite#596](https://github.com/vitejs/rolldown-vite/issues/596) for a related reproduction.

### 2) Switch `is-hotkey` to its named export

`packages/nimbus/src/components/rich-text-input/hooks/use-keyboard-shortcuts.ts`

Vite 8 standardizes CJS default-import interop. `is-hotkey@0.2.0` is babel-compiled CJS with `__esModule: true` AND `exports.default = isHotkey` AND `exports.isHotkey = isHotkey`. The stricter interop now double-wraps the default, leaving `(0, isHotkey.default)` resolving to a non-callable object. Changing `import isHotkey from "is-hotkey"` → `import { isHotkey } from "is-hotkey"` sidesteps the interop entirely (the package exports the same function under both names).

## Verification approach

Each step was committed atomically and validated end-to-end before moving to the next:

1. **[`b25f952a`](https://github.com/commercetools/nimbus/commit/b25f952a) — Switch catalog `vite` to `npm:rolldown-vite@^7.3.1`.** The Vite team's recommended intermediate step — exercises the Rolldown bundler under the Vite 7 API to isolate Rolldown issues from Vite-8 API changes.
2. **[`35077a70`](https://github.com/commercetools/nimbus/commit/35077a70) — Bump `vite-plugin-dts` 4.5.4 → 5.0.0.** v5 renamed `rollupTypes` → `bundleTypes`; we used the default value, so the option is dropped entirely (`bundleTypes` defaults to `false` too).
3. **[`7204aeba`](https://github.com/commercetools/nimbus/commit/7204aeba) — Bump `vite` 7 → 8 + `@vitejs/plugin-react` 5 → 6**, plus the two adaptations above.

Five follow-ups landed after the initial three commits:

- **[`79b73db7`](https://github.com/commercetools/nimbus/commit/79b73db7) — scope `esmExternalRequirePlugin` to the lib build.** At the top-level `plugins` array it leaked into Storybook's iframe build, externalizing `react` without an importmap and breaking the deployed bundle (`Failed to resolve module specifier "react"`). Moved into `build.rollupOptions.plugins` so it only runs for the library build.
- **[`9787a08f`](https://github.com/commercetools/nimbus/commit/9787a08f) — drop redundant rollup dep, tidy `vite.config.ts`.** With Rolldown handling bundling, the direct `rollup` import was unused — but see the next item for why the catalog/override entries had to stay.
- **[`4cbc99e8`](https://github.com/commercetools/nimbus/commit/4cbc99e8) — pin chunk extensions to `.es.js` / `.cjs.js`.** Vite 8/Rolldown changed the `[format]` placeholder from `es` → `esm`, which would have shifted chunk filenames from `*.es.js` to `*.esm.js`. The bundle-size script (`scripts/check-bundle-size.mjs`) hard-codes the pattern `**/*.es.js`, so the upgrade would have silently broken its measurement. Implemented via a tiny `outputOptions` Rollup plugin that writes a literal `chunkFileNames` per format.
- **[`c5cd53d4`](https://github.com/commercetools/nimbus/commit/c5cd53d4) — restore the `rollup` pnpm override.** Looked redundant after the upgrade (nothing imports rollup directly anymore) and was dropped, but it was still forcing `@preconstruct/cli`'s transitive `@rollup/plugin-*` peers onto rollup@4. Without it they fall back to rollup@2, preconstruct emits the secondary entrypoints' chunks duplicated _into_ `packages/tokens/dist`, the published tarball grows by ~250 KB, and CI's bundle-size check fails. Override restored.
- **[`5567f36b`](https://github.com/commercetools/nimbus/commit/5567f36b) — drop `rollup-plugin-tree-shakeable`.** The plugin annotates top-level statements with `/*@__PURE__*/`. Under Rolldown's CJS lib mode, those annotations propagate to Rolldown's synthesized `Object.defineProperty(exports, ...)` calls and the DCE pass eliminates them — every CJS chunk shipped without any exports, breaking `require("@commercetools/nimbus")` for any consumer (every named export resolved to `undefined`). Rolldown emits `/*@__PURE__*/` annotations natively in ESM output (1,680 on the current build) — enough that the extra plugin annotations weren't worth a Rolldown-incompatible code path. Re-measured downstream tree-shaking benefit at ~2.4% gz on a typical consumer bundle.

Two further commits landed during a rebase onto main:

- **[`c639a1bf`](https://github.com/commercetools/nimbus/commit/c639a1bf) — drop `vite-tsconfig-paths` in favor of native `resolve.tsconfigPaths`.** Vite 8 ships a native resolver that supersedes the plugin.
- **[`239b0519`](https://github.com/commercetools/nimbus/commit/239b0519) — regenerate `pnpm-lock.yaml` after rebase onto main.** Bookkeeping.

Verification commands run after each commit:

- `pnpm typecheck`
- `pnpm --filter @commercetools/nimbus build`
- `pnpm build:docs`
- `pnpm --filter blank-app build`
- `pnpm test:unit`
- `pnpm test:storybook`

Final state on the rebased branch: **1,180 unit tests across 85 files**, **886 storybook tests across 79 files, 0 unhandled errors**.

The bundle output was also diff-inspected: the Vite 7 output uses clean ESM `import * from "react"` for the `use-sync-external-store/cjs/*` shim. Pre-fix Vite 8 emitted broken `__require("react")` runtime helpers; post-fix output matches Vite 7 — top-level ESM imports, no CJS factory wrappers, no `__require` foot-guns. Spot-check `grep -r "__require" packages/nimbus/dist --include="*.es.js"` returns zero hits.

## Performance comparison

Apple M-series, 2 runs averaged, measured post-rebase on the final HEAD.

### Build wall-clock

| Task              | Vite 7 + Rollup | Vite 8 + Rolldown | Speedup   |
| ----------------- | --------------- | ----------------- | --------- |
| `nimbus` build    | 13.4 s          | 10.8 s            | **1.24×** |
| `docs` build      | 16.2 s          | 7.1 s             | **2.28×** |
| `blank-app` build | 3.8 s           | 2.0 s             | **1.94×** |

The big wins are on the **consumer-facing builds** (docs and blank-app) — Vite 8 + Rolldown more than halves the docs build and nearly halves blank-app. The `nimbus` library build itself improves by ~24%, dominated by `chakra typegen` and bundling cost.

### Nimbus library output

| Metric                                     | Vite 7      | Vite 8      | Δ         |
| ------------------------------------------ | ----------- | ----------- | --------- |
| `dist/` total (incl. sourcemaps, `du -sh`) | 22 MB       | 21 MB       | −4.5%     |
| Sum of all JS, uncompressed                | 2,803,204 B | 2,680,187 B | **−4.4%** |
| Sum of all JS, gzipped                     | 896,527 B   | 840,767 B   | **−6.2%** |
| JS chunk count                             | 540         | 534         | −6        |

Modest, real reductions from Rolldown's tighter wrapper/runtime overhead and slightly more aggressive chunk merging. Methodology: build twice per ref with `pnpm --filter` against pre-built workspace deps; clean `dist/` and `node_modules/.vite` between runs; sum byte sizes of `**/*.{js,cjs,mjs}` via `stat -f%z` for raw and `gzip -c | wc -c` per file for gzipped; `du -sh dist/` for total.

> **Earlier numbers in this description claimed −24% / −26%.** Those came from a build state where `rollup-plugin-tree-shakeable` was still in `rollupOptions.plugins`. That plugin proved incompatible with Rolldown's CJS lib mode — it adds `/*@__PURE__*/` annotations to top-level statements which Rolldown then propagated to its synthesized `Object.defineProperty(exports, ...)` calls and DCE'd, producing CJS chunks with **zero exports** across the board. Every CJS consumer would have seen `undefined` for every named import. The "−26% gz" was the broken bundle being smaller. The plugin has been removed (commit [`5567f36b`](https://github.com/commercetools/nimbus/commit/5567f36b)). The numbers above are from the post-fix build with all 267 CJS chunks emitting proper exports.

## Test plan

- [ ] CI passes: typecheck, lint, unit tests, storybook tests
- [ ] `pnpm start:storybook` renders components in dev mode
- [ ] `pnpm start:docs` renders the docs site in dev mode
- [ ] Smoke-test the published bundle locally by `pnpm pack`-ing nimbus and consuming it from a fresh app (optional — covered by `apps/blank-app` already)

### DIFF
--- a/.changeset/upgrade-vite-8.md
+++ b/.changeset/upgrade-vite-8.md
@@ -1,0 +1,6 @@
+---
+"@commercetools/nimbus": patch
+---
+
+Internal build toolchain upgraded to Vite 8 + Rolldown. Published JavaScript is
+6% smaller gzipped (−4.4% raw / −6.2% gzipped), with no API or behavior changes.
--- a/.changeset/upgrade-vite-8.md
+++ b/.changeset/upgrade-vite-8.md
@@ -2,5 +2,5 @@
 "@commercetools/nimbus": patch
 ---
 
-Internal build toolchain upgraded to Vite 8 + Rolldown. Published JavaScript is
-6% smaller gzipped (−4.4% raw / −6.2% gzipped), with no API or behavior changes.
+Internal build toolchain upgraded to Vite 8. Published JavaScript is 6% smaller
+gzipped (−4.4% raw / −6.2% gzipped), with no API or behavior changes.
--- a/packages/nimbus/vite.config.ts
+++ b/packages/nimbus/vite.config.ts
@@ -103,11 +103,6 @@ export default defineConfig(async () => {
       optimizeLocales.vite({
         locales: LOCALE_BCP47_CODES,
       }),
-      // Externalize react/react-dom AND rewrite internal `require("react")`
-      // calls inside bundled CJS deps (e.g. use-sync-external-store/cjs/*)
-      // into ESM imports. Replaces the top-level external entries for these.
-      // See `reactExternal` comment above.
-      esmExternalRequirePlugin({ external: reactExternal }),
     ],
     build: {
       // sourcemaps are built into separate files and should therefore be tree-shakeable
@@ -130,9 +125,19 @@ export default defineConfig(async () => {
         formats: ["es", "cjs"] satisfies LibraryFormats[],
       },
       rollupOptions: {
-        // `treeShakeable` naively adds an @__PURE__ annotation to each top-level module in our `dist`
-        // https://github.com/TomerAberbach/rollup-plugin-tree-shakeable?tab=readme-ov-file#why
-        plugins: [treeShakeable()],
+        // - `treeShakeable` naively adds an @__PURE__ annotation to each top-level module in our `dist`
+        //   https://github.com/TomerAberbach/rollup-plugin-tree-shakeable?tab=readme-ov-file#why
+        // - `esmExternalRequirePlugin` externalizes the React family for the
+        //   library build AND rewrites internal `require("react")` calls inside
+        //   bundled CJS deps (e.g. use-sync-external-store/cjs/*) into ESM
+        //   imports. Must be lib-scoped (not in top-level `plugins`) so it does
+        //   not leak into Storybook's iframe build, where externalizing react
+        //   without an importmap breaks the production deployment with
+        //   "Failed to resolve module specifier 'react'".
+        plugins: [
+          treeShakeable(),
+          esmExternalRequirePlugin({ external: reactExternal }),
+        ],
         external,
         output: {
           // Organize chunk files into chunks subfolder
--- a/packages/nimbus/vite.config.ts
+++ b/packages/nimbus/vite.config.ts
@@ -2,8 +2,7 @@ import { fileURLToPath } from "node:url";
 import { glob } from "glob";
 import optimizeLocales from "@react-aria/optimize-locales-plugin";
 import { defineConfig, esmExternalRequirePlugin } from "vite";
-import type { LibraryFormats } from "vite";
-import type { RollupLog, LoggingFunction } from "rollup";
+import type { LibraryFormats, PluginOption, Rollup } from "vite";
 import react from "@vitejs/plugin-react";
 import viteTsconfigPaths from "vite-tsconfig-paths";
 import dts from "vite-plugin-dts";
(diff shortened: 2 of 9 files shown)


packages/nimbus/vite.config.ts
```diff
@@ -119,51 +148,78 @@ export default defineConfig(async () => {
...
+            name: "nimbus-pin-chunk-extensions",
+            outputOptions(options: Rollup.OutputOptions) {
+              return {
+                ...options,
+                chunkFileNames:
+                  options.format === "es"
+                    ? "chunks/[name]-[hash].es.js"
+                    : "chunks/[name]-[hash].cjs.js",
+              };
+            },
```
ByronDWall: Suggestion: Remove the `nimbus-pin-chunk-extensions` plugin

The plugin (lines 178-189 of packages/nimbus/vite.config.ts) pins chunk extensions from Rolldown's new `.esm.js` default back to `.es.js`. I'd suggest removing it. It adds complexity to an already fairly complex build config, and the stated justification doesn't hold up on inspection.

The bundle-size script doesn't care about extensions. The comment says check-`bundle-size.mjs` hard-codes `**/*.es.js`, but the script actually just calls `sumFileSize(dir)` recursively on the entire dist/ directory. It doesn't filter by file bextension at all.

Nothing else depends on chunk extensions. The things that do depend on `.es.js` naming are the entry point filenames (`index.es.js`, `components/button.es.js`, etc.), and those are controlled by the `fileName` callback, which receives format as a runtime string parameter (still "`es`", unaffected by the `[format]` placeholder change). Chunks are internal files; Vite generates matching relative imports in the entry points regardless of what the chunk extensions end up being.

Specifically verified that none of the following are affected by chunk naming:
  - `package.json` `exports` / `module` fields (reference entry points, not chunks)
  - `postinstall` hook (`chakra typegen ./dist/index.es.js` is an entry point)
  - Consumer docs (`npx @chakra-ui/cli typegen ...index.es.js`, same)
  - `attw` / `publint` package shape checks (resolve entry points)
  - Bundle size measurement (extension-agnostic)

Removing the plugin would simplify the rollup plugins array and eliminate the `Rollup.OutputOptions` type import usage that only exists to support this hook. One less thing to maintain or explain to future contributors.

ByronDWall (approved):
Overall this PR looks really good, thanks for taking the time to update our build tooling.

I have one suggestion regarding the `nimbus-pin-chunk-extensions` plugin, but otherwise this looks solid. I would be fine with merging this as-is, or with the plugin removed, whichever you think is best.

packages/nimbus/vite.config.ts
tylermorrisford: my fault - bundle-size-monitoring.md on main still mentions the `**/*.es.js` matching. anyway I agree with this point.

.changeset/upgrade-vite-8.md
```diff
@@ -0,0 +1,6 @@
+---
+"@commercetools/nimbus": patch
+---
+
+Internal build toolchain upgraded to Vite 8 + Rolldown. Published JavaScript is
```
tylermorrisford: Claude suggested that the mention of Rolldown could be removed(per our changeset guidelines) - but I could see the explicit naming of build tools being helpful. Two cents from me and my robot friend.

tylermorrisford (approved):
nice banner by the way

packages/nimbus/vite.config.ts
misama-ct: [Fixed](https://github.com/commercetools/nimbus/pull/1451/commits/1e40f07c5fb65221bd45cc5bbb1aef80667aa205)

.changeset/upgrade-vite-8.md
misama-ct: [Fixed](https://github.com/commercetools/nimbus/pull/1451/commits/1e40f07c5fb65221bd45cc5bbb1aef80667aa205)

PR #1454, opened 2026-05-04
feat(confirmation-dialog): add ConfirmationDialog pattern (FEC-431)

misama-ct:
image

## Summary

- Adds a pre-composed `ConfirmationDialog` pattern under `patterns/dialogs/confirmation-dialog/` as the Nimbus replacement for merchant-center-application-kit's `ConfirmationDialog`. Flat API (`title`, `children`, `onConfirm`, `onCancel`, plus `isOpen`/`defaultOpen`/`onOpenChange`, `confirmLabel`/`cancelLabel`, `intent`, `isConfirmDisabled`/`isConfirmLoading`, `aria-label`) composing `Dialog.*` and two `Button`s internally.
- `intent="destructive"` maps the confirm button to `colorPalette="critical"`; `isConfirmLoading` shows a spinner and locks both action buttons plus ambient dismiss affordances (Escape, overlay, X). Cancel button, Escape, overlay, and X all route through a single `onCancel` handler.
- Internalizes open state via `useControlledState` to avoid the `slot="close"` race that double-fires `onCancel`. Default Confirm/Cancel labels localized via Nimbus i18n (`.i18n.ts` + `useLocalizedStringFormatter`, FormActionBar precedent). Also fixes `Dialog.Content` to forward `aria-label` from `Dialog.Root` context to the underlying `RaDialog`.

## Test plan

- [x] `pnpm --filter @commercetools/nimbus typecheck` passes
- [x] `pnpm --filter @commercetools/nimbus build` succeeds
- [x] `pnpm lint` passes
- [x] `pnpm test packages/nimbus/src/patterns/dialogs/confirmation-dialog/confirmation-dialog.stories.tsx` passes (default intent, destructive intent, custom labels, disabled confirm, loading lockout, ReactNode title with aria-label, long scrollable content)
- [x] `pnpm test packages/nimbus/src/patterns/dialogs/confirmation-dialog/confirmation-dialog.docs.spec.tsx` passes
- [x] Verify destructive intent renders confirm button with critical color palette
- [x] Verify `isConfirmLoading` disables both buttons and blocks Escape/overlay/X dismissal
- [x] Verify Cancel/Escape/overlay/X all route through a single `onCancel`
- [ ] Verify default labels render in en, de, es, fr-FR, pt-BR
- [x] Verify `aria-label` on `Dialog.Root` propagates to `Dialog.Content`


misama-ct:
@claude please review, be gentle

misama-ct:
@claude Thanks for the careful review — the async-confirm finding in particular was the right call. Addressed in fa1f1c7cd. Below is an explanation, have a - hopefully - last look and let me know if we're good to go now.

### What changed

**1. Async `onConfirm` (your main concern) — fixed via your option (2)**

`onConfirm` is now typed `() => void | Promise<void>`. When the consumer returns a Promise, the pattern keeps the dialog open until it settles: close on fulfillment, stay open on rejection so the consumer can surface the error and let the user retry. Sync confirms close immediately as before.

This is **not** the auto-toggle that `design.md` rejected — `isConfirmLoading` stays consumer-managed, consistent with `FormActionBar`'s `isSaveLoading`. We only delay the pattern's own `setOpen(false)` until the consumer's work is done.

Why (2) over (1) or (3):

- **(1) document a guard pattern**: the required guard needs a `useRef` to dodge a stale-closure trap (the loading flag is stale inside `onOpenChange`) — that's exactly the React-internals footgun a flat pattern is supposed to absorb. The failure mode (spinner silently doesn't render) is also hard to diagnose.
- **(3) consumer always closes**: breaks the flat-API promise (~95% of audited usages are simple sync confirms where the pattern handling close is the whole point) and creates an asymmetry with `handleCancelButton` (which still auto-closes) that's actively confusing.
- **(2)** is strictly additive at the type level (`void` widens to `void | Promise<void>`), preserves sync-confirm semantics unchanged, and makes `isConfirmLoading` finally work for the case it was designed for.

Two new stories cover the path end-to-end: `AsyncConfirm` (fulfilling promise → dialog closes after `await`) and `AsyncConfirmRejection` (rejected promise → dialog stays open with confirm button re-enabled for retry). Spec (`spec.md`), `design.md` (resolved the corresponding Open Question), `proposal.md`, `tasks.md`, and `.dev.mdx` all updated to match the new contract.

**2. Redundant `setIsOpen(false)` in `docs.spec.tsx` — fixed**

Stripped the misleading inner close calls and added an `expect(handleCancel).not.toHaveBeenCalled()` assertion so the test now also verifies the confirm path doesn't fire `onCancel`.

**6. "Ambient" comment in `confirmation-dialog.tsx` — reworded**

Now reads "dismiss paths not driven by an explicit `onPress` in this component" — your suggested phrasing is more precise.

### What I didn't change (and why)

**3. English placeholders in non-English locale files** — expected per the Transifex workflow. Translations land via the Transifex bot PR; the placeholders unblock build until then. Test-plan checkbox left honestly unchecked since it's a manual visual check.

**4. Overlay-click test coords `(2, 2)`** — the test walks up to the overlay element first (`while (overlay.parentElement !== document.body)`), so the `(2, 2)` click lands on the actual overlay, not on a fixed viewport coordinate. It's robust at any sane viewport size; if a future Dialog change makes the overlay tightly hug the modal, the same fix would apply to `Dialog`'s own stories. Leaving as-is.

**5. Spinner placement after a `ReactNode` confirm label** — the spinner sits after the label content with `ml="100"`, so for the common composed cases (`<><Icon /> Save</>`) it reads cleanly as `[icon] [text] [spinner]`. Visually fine in all current stories.

### Validation

- `pnpm --filter @commercetools/nimbus typecheck` — passes
- `pnpm test:dev` for `confirmation-dialog.docs.spec.tsx` — 9/9 pass
- `pnpm test:storybook:dev` for `confirmation-dialog.stories.tsx` — 9/9 pass (including the two new async stories)
- `pnpm openspec validate confirmation-dialog-pattern --strict` — valid
- `pnpm lint` — clean

valoriecarli:
hey... @tylermorrisford does this require a LOC ticket? im not sure.

tylermorrisford (approved):

misama-ct:
> hey... @tylermorrisford does this require a LOC ticket? im not sure.

Ehm... @valoriecarli  We did that ticket together yesterday, remember? 👵 😄

packages/nimbus/src/patterns/dialogs/confirmation-dialog/confirmation-dialog.stories.tsx
```diff
@@ -0,0 +1,732 @@
...
+ * `confirmLabel` and `cancelLabel` accept any ReactNode, so consumers
+ * can pass `intl.formatMessage(...)` results directly.
+ */
+export const CustomLabels: Story = {
+  args: {
+    onConfirm: fn(),
+    onCancel: fn(),
+  },
+  render: (args) => {
+    const [isOpen, setIsOpen] = useState(true);
```
ByronDWall: Stories that render an open dialog on mount are causing the `Docs` storybook page to show multiple modals at once. I think it would be better to have a trigger button that the test clicks to render the dialog, so that the storybook pages are usable.

packages/nimbus/src/patterns/dialogs/confirmation-dialog/confirmation-dialog.stories.tsx
```diff
@@ -0,0 +1,732 @@
...
+ * for the duration. Consumers should hold this state across an
+ * in-flight async confirm.
+ */
+export const LoadingLockout: Story = {
+  args: {
+    onConfirm: fn(),
+    onCancel: fn(),
+  },
+  render: (args) => {
+    const [isOpen, setIsOpen] = useState(true);
```
ByronDWall: This test in particular should not render the dialog on mount, as the user has no way of closing the dialog. This makes the storybook docs page completely inaccessible.

ByronDWall (approved):
Overall it looks good, besides the storybook issue. I do think all the dialog stories should have a button that must be clicked to render the dialog (see comments below).

One question (nit) - would the `FormActionBar` be useful for rendering the buttons? It's meant to handle save/cancel/delete buttons in forms/dialogs, but if there is enough of a divergence that this component legitimately needs to handle buttons differently, that is fine.

packages/nimbus/src/patterns/dialogs/confirmation-dialog/confirmation-dialog.stories.tsx
misama-ct: Fixed in 941980a9f

packages/nimbus/src/patterns/dialogs/confirmation-dialog/confirmation-dialog.stories.tsx
misama-ct: Fixed in 941980a9f

misama-ct:
Okay regarding the `FormActionBar` I consulted claude, essentially, we'd trade 12 lines of markup for 7 lines of markup, and extra DOM node and and extra dependency - but only after doing a couple of changes to FormActionBar (e.g the destructive intent doesn't fit FormActionBar's current shape). I don't think it's necessary, in any case this should be done in a follow up.

PR #1477, opened 2026-05-07
perf(combobox): fix multi-select render performance and TagGroup remount bug

ByronDWall:
## Summary

Fixes multi-select ComboBox performance with many selected tags ([FEC-908](https://commercetools.atlassian.net/browse/FEC-908)).

### Root-cause fix: TagGroup remount bug

`TagGroupTagListSlot` was calling `withContext` inside the component function, producing a new styled-component identity on every render. React unmounted and remounted the entire `<TagList>` subtree each time the parent re-rendered, and React Aria rebuilt its collection from scratch. This was catastrophic for multi-select ComboBox with many tags — every keystroke triggered a full remount.

The fix hoists that single `withContext` call to module scope, matching the pattern already used by `TagGroupRootSlot` and `TagGroupTagSlot`. This benefits all TagGroup consumers, not just ComboBox.

### ComboBox performance improvements

- **Memoize all context values individually** (input, tag group, button, popover, listbox, selectable collection) to prevent cascading re-renders when only one context changes
- **Stabilize `selectedKeys` fingerprinting** so the memo only recomputes when actual keys change, not when the parent passes a new array reference
- **Extract `singleSelectedTextValue`** to reduce `filteredCollection` dependency on the full selection set
- **Fix `handleBlur`** to check `triggerRef` + `popoverRef` instead of `e.currentTarget` (which was the input element, not the combobox wrapper — it missed focus moving to the portaled popover)
- **Move `collectionPopulatedRef`** into `useEffect` for proper React semantics
- **Set `autoFocus={false}`** on popover to prevent focus steal

### UX improvements

- **Add max-height safety net** on the content area with scrollable overflow, preventing the trigger from growing unbounded with many selected tags
- **Auto-scroll content to bottom** when adding items so the input stays visible (does not scroll on tag removal)

### i18n cleanup

- Remove dead `Nimbus.ComboBox.removeTag` message from all locale data files and compiled intl outputs (tag removal is handled entirely by TagGroup's own i18n)

## Test plan

- [x] All 112 combobox story tests pass (110 existing + 2 new)
- [x] All 7 tag-group story tests pass
- [x] TypeScript compilation passes
- [x] New `KeyboardTagGroupNavigation` story verifies arrow nav between tags, Delete/Backspace removal with focus restoration
- [x] New `ContentAreaScrollBehavior` story verifies scroll-to-bottom on selection
- [ ] Manual: `PerformanceManySelected` story (170 tags) — keystroke latency, tag removal, dropdown open
- [ ] Manual: `MultiSelectCustomOptions` — arrow-key navigation across tags, focus restoration after remove

[FEC-908]: https://commercetools.atlassian.net/browse/FEC-908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/combobox-multiselect-perf.md
+++ b/.changeset/combobox-multiselect-perf.md
@@ -1,0 +1,7 @@
+---
+"@commercetools/nimbus": patch
+---
+
+`ComboBox`: improved rendering performance in multi-select mode with many
+selected items. The tag area now has a max-height with scroll overflow to
+prevent the trigger from growing unbounded.
--- a/.changeset/taggroup-taglist-slot-perf.md
+++ b/.changeset/taggroup-taglist-slot-perf.md
@@ -1,0 +1,10 @@
+---
+"@commercetools/nimbus": patch
+---
+
+`TagGroup`: fix a render-time bug where `TagGroupTagListSlot` instantiated its
+styled component inside the component function, producing a new component
+identity on every render. This caused React to unmount and remount the entire
+tag list (and React Aria to rebuild its collection from scratch) on every parent
+render — severely impacting performance when many tags were rendered (e.g. in a
+multi-select ComboBox).
--- a/packages/i18n/data/core.json
+++ b/packages/i18n/data/core.json
@@ -35,10 +35,6 @@
     "developer_comment": "aria-label for combobox menu options",
     "string": "Options menu"
   },
-  "Nimbus.ComboBox.removeTag": {
-    "developer_comment": "aria-label for remove tag button in multi-select ComboBox",
-    "string": "Remove tag {tagName}"
-  },
   "Nimbus.ComboBox.selectedValues": {
     "developer_comment": "aria-label for selected values tag group",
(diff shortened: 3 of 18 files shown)


tylermorrisford (approved):
does the thing

valoriecarli (approved):
🙌🏻

misama-ct (changes requested):
OBSOLETE

misama-ct (approved):
Good morning! I had a look at the combobox fix and have [one suggestion](https://github.com/commercetools/nimbus/pull/1485) that I'd love to see implemented as it would fix some of the issues and preserve existing keyboard behavior.

I'm not 100% sure if an openspec proposal is necessary, as no functionality was added and the fixes are 'react'-specific fixes, but maybe you can have a look if it makes sense and add one if it does.

PR #1478, opened 2026-05-07
perf(data-table): optimize render performance for large datasets

ByronDWall:
## Summary

Performance optimizations for `DataTable` with large datasets (~170+ rows). The core strategy is reducing unnecessary React re-renders by isolating volatile state from per-row contexts, so only affected rows re-render during sort, expand, and pin interactions.

### Context splitting (3-layer architecture)

| Context | Contents | Subscribers |
|---|---|---|
| `DataTableContext` (stable) | Columns, callbacks, layout flags | Rows (via `useStableDataTableContext`) |
| `RowsDataContext` (volatile) | `sortedRows`, `filteredRows`, `sortDescriptor`, `expanded`, `pinnedRows`, `pinnedRowIds` | Body only |
| `TableSelectionContext` | `selectedKeys`, `onSelectionChange` | Table only |

### Per-row derived props

Body computes `isExpanded`, `isPinned`, `isFirstPinned`, `isLastPinned`, `isSinglePinned` per row and passes as props. `React.memo` on `DataTableRow` skips unaffected rows — only the toggled row re-renders on expand/pin.

### Sort performance

- **Pre-computed value map** — accessor called O(n) instead of O(n log n) times inside the comparator
- **`Intl.Collator` with `numeric: true`** — native natural sort ("Product 1, 2, 10" not "1, 10, 100"), faster than manual `String.toLowerCase()` comparison
- **Numeric fast-path** — pure number columns use `(a - b) * direction` subtraction

### Callback stabilization

`toggleExpand`, `togglePin`, `handleSortChange` stabilized with refs (empty `[]` deps). Previously `toggleExpand` closed over `expanded` Set, creating a new callback on every expand which invalidated the entire context.

### Other optimizations

- **Hoist `useLocalizedStringFormatter`** from per-row (170 hook calls) to root context (1 call)
- **Hoist `pinnedRowIds`** computation from per-row O(N²) to root O(N), stabilized to depend on `rows` + `pinnedRows` instead of `sortedRows`
- **`showExpandColumn` memoized** — `hasExpandableRows()` recursive walk skipped when deps unchanged
- **Replace per-row `Tooltip` with native `title`** — eliminates 170 `Tooltip.Root` mount/unmount cycles
- **Wrap sort/expand/pin in `startTransition`** — React yields to browser between row renders

### Measured perf gains (170-row dataset, Storybook interaction tests)

| Interaction | Before | After | Improvement |
|---|---|---|---|
| Sort click (CI) | ~3300ms (failing) | <2000ms (passing) | ~40% faster |
| Selection toggle | <2000ms | <2000ms | Stable, fewer re-renders |
| Expand/Pin | All 170 rows re-render | Only affected row(s) re-render | ~99% fewer re-renders |

## Test plan

- [ ] All 31 data-table story tests pass (`pnpm test:dev packages/nimbus/src/components/data-table/data-table.stories.tsx`)
- [ ] Selection toggle on ~170 rows completes under 2s
- [ ] Sort on ~170 rows completes under 2s (CI threshold)
- [ ] Natural sort order: "Product 1, 2, 10" not "1, 10, 100"
- [ ] Pin/unpin works, pinned row styling applies correctly
- [ ] Expand/collapse works for nested tables
- [ ] Double-click on cell text selects text without triggering row click
- [ ] `useDataTableContext()` public hook still returns full context (backwards compatible)

### DIFF
--- a/packages/nimbus/src/components/data-table/components/data-table.context.tsx
+++ b/packages/nimbus/src/components/data-table/components/data-table.context.tsx
@@ -7,7 +7,7 @@ import type {
   TableSelectionContextValue,
 } from "../data-table.types";
 
-type RowsDataContextValue<T extends object = Record<string, unknown>> = {
+type InteractionContextValue<T extends object = Record<string, unknown>> = {
   sortedRows: DataTableRowItem<T>[];
   filteredRows: DataTableRowItem<T>[];
   sortDescriptor?: SortDescriptor;
@@ -21,10 +21,10 @@ export const DataTableContext = createContext<DataTableContextValue<
 > | null>(null);
 DataTableContext.displayName = "DataTable.Context";
 
-export const RowsDataContext = createContext<RowsDataContextValue<
+export const InteractionContext = createContext<InteractionContextValue<
   Record<string, unknown>
 > | null>(null);
-RowsDataContext.displayName = "DataTable.RowsDataContext";
+InteractionContext.displayName = "DataTable.InteractionContext";
 
 export const TableSelectionContext =
   createContext<TableSelectionContextValue | null>(null);
@@ -40,15 +40,15 @@ export const useDataTableContext = <
   const context = useContext(
     DataTableContext
   ) as DataTableContextValue<T> | null;
-  const rowsData = useContext(
-    RowsDataContext
-  ) as RowsDataContextValue<T> | null;
+  const interactionData = useContext(
+    InteractionContext
+  ) as InteractionContextValue<T> | null;
   if (!context) {
     throw new Error("DataTable components must be used within DataTable.Root");
   }
   return useMemo(
-    () => ({ ...context, ...rowsData }),
-    [context, rowsData]
+    () => ({ ...context, ...interactionData }),
+    [context, interactionData]
   ) as DataTableContextValue<T>;
 };
 
@@ -64,10 +64,12 @@ export const useStableDataTableContext = <
   return context;
 };
 
-export const useRowsDataContext = <
+export const useInteractionContext = <
   T extends object = Record<string, unknown>,
->(): RowsDataContextValue<T> => {
-  const context = useContext(RowsDataContext) as RowsDataContextValue<T> | null;
+>(): InteractionContextValue<T> => {
+  const context = useContext(
+    InteractionContext
+  ) as InteractionContextValue<T> | null;
   if (!context) {
     throw new Error("DataTable components must be used within DataTable.Root");
   }
--- a/packages/nimbus/src/components/data-table/components/data-table.context.tsx
+++ b/packages/nimbus/src/components/data-table/components/data-table.context.tsx
@@ -2,6 +2,7 @@ import { createContext, useContext, useMemo } from "react";
 import type {
   DataTableContextValue,
   DataTableRowItem,
+  SortDescriptor,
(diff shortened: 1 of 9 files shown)


valoriecarli (approved):

tylermorrisford (approved):
I like it

packages/nimbus/src/components/data-table/data-table.types.ts
```diff
@@ -109,8 +109,6 @@ export type DataTableContextValue<T extends object = Record<string, unknown>> =
     renderEmptyState?: RaTableBodyProps<T>["renderEmptyState"];
     search?: string;
     sortDescriptor?: SortDescriptor;
-    selectedKeys?: Selection;
-    defaultSelectedKeys?: Selection;
     expanded: Set<string>;
     allowsSorting?: boolean;
     selectionMode?: "none" | "single" | "multiple";
```
misama-ct, line 116: **Public-API breaking change.** `selectedKeys`, `defaultSelectedKeys`, and `onSelectionChange` were removed from `DataTableContextValue` here (they used to sit just above `selectionMode`). Both this type and `useDataTableContext` (which returns it) are publicly exported and JSDoc'd as consumer-facing in `data-table.tsx:166-176`.

Repro that compiles on `main`, breaks here:

```tsx
const { selectedKeys, onSelectionChange } = DataTable.useDataTableContext();
// TS2339 + undefined at runtime
```

Per `docs/changeset-conventions.md`, narrowed types = `major`. Changeset is currently `patch`.

Fix on the runtime side ↓ — see paired comment on `data-table.context.tsx:50`.

packages/nimbus/src/components/data-table/components/data-table.context.tsx
```diff
@@ -1,21 +1,60 @@
...
+    DataTableContext
+  ) as DataTableContextValue<T> | null;
+  const interactionData = useContext(
+    InteractionContext
+  ) as InteractionContextValue<T> | null;
+  if (!context) {
+    throw new Error("DataTable components must be used within DataTable.Root");
+  }
+  return useMemo(
+    () => ({ ...context, ...interactionData }),
```
misama-ct: Pair with the comment on `data-table.types.ts:114`.

The merge here intentionally excludes `TableSelectionContext`, which is what causes the runtime breakage. Restoring the public contract while keeping the perf split:

```ts
const selection = useContext(TableSelectionContext);
// ...
return useMemo(
  () => ({ ...context, ...interactionData, ...selection }),
  [context, interactionData, selection],
) as DataTableContextValue<T>;
```

Internal hot-path consumers (e.g. `DataTableRow`) keep using `useStableDataTableContext` and still get the perf win. Public hook stays compatible, changeset stays `patch`.

ByronDWall: fixed in [766eeac](https://github.com/commercetools/nimbus/pull/1478/commits/766eeac3cbcb7eed0bda565c889b139936b717a7)

packages/nimbus/src/components/data-table/data-table.types.ts
ByronDWall: fixed in [766eeac](https://github.com/commercetools/nimbus/pull/1478/commits/766eeac3cbcb7eed0bda565c889b139936b717a7)

PR #1485, opened 2026-05-08
perf(tag-group): fix render-time bug causing ComboBox multi-select lag

misama-ct:
## Summary

Builds on `FEC-908-combobox-perf` ([FEC-908](https://commercetools.atlassian.net/browse/FEC-908)).

`TagGroupTagListSlot` was calling `withContext` inside the component function:

```tsx
export const TagGroupTagListSlot = <T extends object>(props) => {
  const { ref, ...restProps } = props;
  const SlotComponent = withContext<HTMLDivElement, TagGroupTagListProps<T>>(
    RaTagList,
    "tagList",
  ); // runs every render → new component identity
  return <SlotComponent {...restProps} ref={ref} />;
};
```

This produced a fresh styled-component identity on every render, so React unmounted and remounted the entire `<TagList>` subtree each time the parent re-rendered, and React Aria rebuilt its collection from scratch. For a multi-select ComboBox with many selected tags, that compounds on every keystroke. The sibling `TagGroupRootSlot` and `TagGroupTagSlot` are already defined at module scope.

The fix is hoisting that one call to module scope. Once that's in place, the ComboBox-side tag rewrite from the base branch isn't needed any more, so this PR also reverts that part — restoring RAC's `TagGroup` inside ComboBox while keeping every other perf change from the base branch.

## What the revert brings back

The lightweight-tag implementation rendered tags as `role="list"` + a plain `IconButton` per tag, which lost the WAI-ARIA `grid` pattern RAC's `useTag` implements. Two concrete keyboard improvements at the 170-tag scale this ticket targets:

- **Tabbing past the ComboBox.** RAC TagGroup makes the whole group a single Tab stop — one Tab in, one Tab out. With per-tag remove buttons each one is its own Tab stop, so a keyboard user had to press Tab 170 times to reach the next form field.
- **Deleting a specific tag.** With RAC TagGroup, arrow to the tag and press `Delete`/`Backspace`; focus moves to a neighbor afterward so the user can keep deleting. With per-tag remove buttons there was no arrow navigation between tags, `Delete` did nothing, and after a click the destroyed button left focus to fall back to `<body>`.

## Commits

1. `revert(combobox): restore RAC TagGroup, drop lightweight-tag swap` — keeps every other perf change from the base branch (handleBlur ref check, `autoFocus={false}`, `collectionPopulatedRef` in useEffect, `singleSelectedTextValue`, selectedKeys fingerprint, per-context memos, `scrollContentToBottom`, content max-height). Removes the `tag` slot, `ComboBoxTagSlot`, `removeTag` i18n entry, the stale `overflowTagCount` in `intl/de.ts`, and the `[role="listitem"]` test selectors.
2. `perf(tag-group): hoist withContext call out of TagGroupTagListSlot` — the fix, plus a comment explaining why module scope matters here, plus a changeset.

## Test plan

- [x] `pnpm test:dev packages/nimbus/src/components/combobox/combobox.stories.tsx` — 111/111 passing
- [x] `pnpm test:dev packages/nimbus/src/components/tag-group/tag-group.stories.tsx` — 7/7 passing
- [ ] Manual: `PerformanceManySelected` story (170 tags) — keystroke latency, tag removal, dropdown open
- [ ] Manual: `MultiSelectCustomOptions` — arrow-key navigation across tags, focus restoration after remove

[FEC-908]: https://commercetools.atlassian.net/browse/FEC-908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/taggroup-taglist-slot-perf.md
+++ b/.changeset/taggroup-taglist-slot-perf.md
@@ -1,0 +1,10 @@
+---
+"@commercetools/nimbus": patch
+---
+
+`TagGroup`: fix a render-time bug where `TagGroupTagListSlot` instantiated its
+styled component inside the component function, producing a new component
+identity on every render. This caused React to unmount and remount the entire
+tag list (and React Aria to rebuild its collection from scratch) on every parent
+render — severely impacting performance when many tags were rendered (e.g. in a
+multi-select ComboBox).
--- a/packages/nimbus/src/components/combobox/combobox.i18n.ts
+++ b/packages/nimbus/src/components/combobox/combobox.i18n.ts
@@ -19,9 +19,4 @@ export const messages = {
     description: "aria-label for combobox menu options",
     defaultMessage: "Options menu",
   },
-  removeTag: {
-    id: "Nimbus.ComboBox.removeTag",
-    description: "aria-label for remove tag button in multi-select ComboBox",
-    defaultMessage: "Remove tag {tagName}",
-  },
 };
--- a/packages/nimbus/src/components/combobox/combobox.recipe.ts
+++ b/packages/nimbus/src/components/combobox/combobox.recipe.ts
@@ -12,7 +12,6 @@ export const comboBoxSlotRecipe = defineSlotRecipe({
     "leadingElement",
     "content",
     "tagGroup",
-    "tag",
     "input",
     "popover",
     "listBox",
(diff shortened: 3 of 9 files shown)


ByronDWall (approved):
Definitely better ux and similarly performant. Ive also added tests to make sure the keyboard navigation when tags are rendered does not regress. Thanks for the PR!

PR #1486, opened 2026-05-08
docs: codify slot-factory call-site rule and perf-debugging discipline

misama-ct:
## Summary

Codifies four learnings from the FEC-908 ComboBox perf debug ([PR #1485](https://github.com/commercetools/nimbus/pull/1485)) into durable guidance, so future component work avoids the same caveats.

The original bug: `TagGroupTagListSlot` invoked `withContext` inside the component function body, producing a fresh styled-component identity on every render. React unmounted/remounted the entire `<TagList>` subtree on every parent render, and React Aria rebuilt its collection from scratch. Hot path for high-tag-count usage; invisible at low counts. The first response was to fork tag rendering inside ComboBox with simpler ARIA roles, which would have shipped a real keyboard-a11y regression at scale (170 Tab presses to skip past the ComboBox; `Delete` becoming a no-op).

This PR captures the principles that came out of that debug as guidance.

### `docs/file-type-guidelines/slots.md`

Two new entries under "Critical Requirements":

- **MUST Invoke Slot Factories at Module Scope** — `createRecipeContext`, `createSlotRecipeContext`, `withProvider`, `withContext` and equivalents must be called at module scope. Explains why (component identity, unmount/remount thrash, RAC collection rebuild) and includes a positive `menu.slots.tsx` example.
- **Threading Generics Through Slot Wrappers** — when a slot needs `<T extends object>`, define the slot once at module scope with the generic erased to `object`, and preserve the generic only at the wrapper's type boundary. Includes the canonical `tag-group.slots.tsx` pattern.

Validation Checklist gets two new items covering both rules.

### `docs/file-type-guidelines/architecture-decisions.md`

- **Preserving WAI-ARIA Composite Widget Patterns** (under "React Aria Integration Matrix") — explains what RAC's composite-widget primitives (`TagGroup`, `ListBox`, `GridList`, `Menu`, `Table`) provide (single Tab stop, arrow navigation, `Delete` on focused row, focus restoration) and the bar required to drop one in favor of flatter ARIA roles.
- **Performance Investigation Discipline** (new top-level section) — fix the cost at the lowest layer where it lives (slot factory, recipe, RAC primitive) before forking the consumer. Lists the trade-offs of forking (drift from upstream, hidden bug for other consumers, silent a11y regressions, ongoing maintenance).

### `.claude/skills/writing-slots/SKILL.md`

Mirrors the slots.md additions so the writing-slots skill enforces the rules on new and updated slot files. Includes the same module-scope rule, the generic-erasure pattern, and two new validation checklist items.

## Why two PRs

[PR #1485](https://github.com/commercetools/nimbus/pull/1485) is the actual fix and lives on a ComboBox-perf branch. This PR is purely documentation/process and is independently reviewable, so it lands on its own merge cycle and isn't blocked by review of the perf change.

## Test plan

- [x] All new code examples are positive (production-ready, no anti-patterns)
- [x] All claimed behaviors verified against the FEC-908 debug findings
- [x] Skill checklist updates align with new doc rules
- [ ] Reviewer sanity check: any existing slot files that violate the new rule? (None found in a quick scan, but worth a second pair of eyes)

### DIFF
--- a/.claude/skills/writing-slots/SKILL.md
+++ b/.claude/skills/writing-slots/SKILL.md
@@ -177,6 +177,84 @@ export type SlotComponent<
 >;
 ```
 
+## Factory Call Site (CRITICAL)
+
+**`createRecipeContext`, `createSlotRecipeContext`, `withProvider`,
+`withContext`, and any equivalent factory that returns a React component MUST
+be invoked at module scope.**
+
+The component identity returned by these factories is what React uses to
+reconcile the rendered subtree. Re-invoking a factory inside a component or
+function body produces a new component identity on every render, which forces
+React to unmount and remount the entire styled subtree each time the parent
+renders. When the slot wraps a React Aria collection primitive (`TagList`,
+`ListBox`, `GridList`, `Menu`), the collection is also rebuilt from scratch
+on every render. The cost scales with subtree size and only manifests at
+high item counts, so the bug commonly evades small-scale review.
+
+```typescript
+// menu.slots.tsx
+import {
+  createSlotRecipeContext,
+  type HTMLChakraProps,
+} from "@chakra-ui/react/styled-system";
+import type { SlotComponent } from "../utils/slot-types";
+
+// Factory invoked once, at module scope.
+const { withProvider, withContext } = createSlotRecipeContext({
+  key: "nimbusMenu",
+});
+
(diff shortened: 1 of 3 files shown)


misama-ct:
Not ready for primetime.

PR #1498, opened 2026-05-12
chore(ci): add package-shape baseline check (attw + publint)

misama-ct:
## Motivation

The Vite 8 migration raised a question without a confident answer: how should the project _prove_ that the published packages still work for consumers after a toolchain shift? Two adaptations were needed mid-PR — externalize React differently, switch `is-hotkey` to its named export — and both were caught only by hand-diffing the emitted bundles. `apps/docs` and `apps/blank-app` exercise Nimbus from a Vite bundler and consume via workspace symlinks, which sidesteps the entire `exports` map and is exactly the wrong consumer profile to catch CJS/ESM interop regressions.

This PR establishes a reusable baseline for future toolchain shifts. CI now packs each publishable tarball and runs two industry-standard linters against the result — what npm would actually ship, not the workspace-linked source. Next bundler swap, dependency bump, or `vite.config.ts` change runs through this filter before merge.

## The tools

**`@arethetypeswrong/cli` (attw)** validates that a package's TypeScript types resolve correctly under every module-resolution mode a consumer might realistically use:

- `node10` — legacy TS/bundler resolution; Jest defaults
- `node16` from CJS — modern Node when the consumer is CJS (Jest in JSDOM, classic Next.js pages dir)
- `node16` from ESM — modern Node when the consumer is ESM (Next.js app router server components, Remix loaders, Node 22+ scripts with `"type": "module"`)
- `bundler` — Vite, Webpack 5, esbuild, Rspack (the realistic majority of consumers)

It walks the `exports` map, traces every entry point, and flags any mode where types are missing, malformed, or describe a different shape than the runtime actually exports.

**`publint`** validates `package.json` correctness against the npm spec. It catches things the spec says are wrong but the npm CLI doesn't enforce: missing files referenced by `exports`, inconsistent `import`/`require` conditions, `.js` files described as ESM but interpreted as CJS, deprecated fields, wrong extensions for the declared `"type"`. Output is tiered — errors fail the check, warnings fail strict mode, suggestions are informational only.

Both run against the same tarball `pnpm pack` produces, which is exactly what npm would publish.

## What's added

- `pnpm check:package-shape` — packs `@commercetools/nimbus`, `@commercetools/nimbus-icons`, `@commercetools/nimbus-mcp` and runs attw plus publint against each tarball.
- Wired into `build-and-test.yml` after Typecheck. Runs identically locally — no CI-only divergence. [See the check's CI output](https://github.com/commercetools/nimbus/actions/runs/25726107125/job/75540725055?pr=1498#step:10:1).
- Strict by default. Findings fail the check; flip `REPORT_ONLY = true` in the script if you need to land something with known issues to triage separately.
- `docs/package-shape-verification.md` — reference doc covering the command, what each tool catches, when to run it locally, and how to debug failures. Linked from the Code Quality section of `CLAUDE.md`.

## What it caught on `main`

Four pre-existing findings, all fixed here so the check ships green.

**`@commercetools/nimbus`**

1. **`./setup-jsdom-polyfills` subpath had no types.** The `vite-plugin-dts` exclude pattern (`src/test/**/*`) accidentally dropped the polyfills source file even though it's a real published entry point. Narrowed the exclude to the test-only files; relocated the emitted `.d.ts` in postbuild (`vite-plugin-dts` preserves `src/` subdirs, Vite emits the runtime flat).
2. **CJS consumers got ESM-flavored types.** `package.json` `"type": "module"` caused TypeScript to read `index.d.ts` as ESM regardless of which path imported it. CJS consumers (Jest in JSDOM, classic Next.js pages dir) got types that didn't match their runtime. Postbuild now duplicates `index.d.ts` → `index.d.cts` (and the same for `setup-jsdom-polyfills`). Identical content; the extension flips how TS interprets the file. `exports[*].require.types` points at the `.d.cts`.
3. **Internal resolution error.** `vite-plugin-dts` emits bare directory re-exports (`export * from './foo'` without `/index.js`). Node16 ESM rejects extensionless relative imports — breaks consumers with `moduleResolution: "nodenext"`. Tried `rollupTypes: true` first; `@microsoft/api-extractor` stack-overflowed on the type graph depth. Settled on a postbuild script that walks every emitted `.d.ts`, regex-resolves bare relative imports against the filesystem, and rewrites them (454 files rewritten on a clean build). Runs before the `.d.cts` duplication so both flavors share the fix.

**`@commercetools/nimbus-icons`**

4. **ESM build unimportable from native Node ESM.** `tsc` emits `.js` without flipping extensions and the package has no `"type"` field, so Node defaulted both builds to CJS by extension. ESM consumers crashed parsing the `export` keyword. Marker `package.json` files now live in `dist/esm/` (`{"type":"module"}`) and `dist/cjs/` (`{"type":"commonjs"}`); same regex rewriter applied to `dist/esm/**/*.{js,d.ts}` (CJS path is left alone since CJS resolution supports extensionless imports natively).

## Impact on consumers

Worth flagging that this isn't just internal plumbing — there are real wins shipping in the published tarballs. All changes are pure additions or improvements; **no breaking changes**, no regressions for any consumer profile.

Bundler-based consumers (Vite, Webpack, Next.js — the realistic 95%+) see identical behavior. The textual changes to `.d.ts` files don't alter the type graph — same symbols, same shapes, just resolution paths that more environments can follow.

Who's better off after this lands:

- **Strict-TypeScript consumers** (`moduleResolution: "nodenext"` or `"node16"`) — type-checking now passes against Nimbus imports across the type graph.
- **CJS consumers** (Jest in JSDOM, classic Next.js pages dir) — types now correctly describe the CJS runtime shape via `index.d.cts`.
- **Anyone using `@commercetools/nimbus/setup-jsdom-polyfills`** — gets autocomplete and signature checking where there were no types at all.
- **Native-Node-ESM consumers of `@commercetools/nimbus-icons`** — package actually imports (rare today; relevant for SSR contexts that externalize node_modules and for the growing Node-22+ ESM-first ecosystem).

Changeset included in the PR.

## Build verification

Clean-room comparison of `pnpm pack` output against `main`:

- **`@commercetools/nimbus-mcp`** — byte-identical. No files in this package were touched.
- **`@commercetools/nimbus`** — after normalizing Vite's content-hash chunk filenames, 539 of 540 runtime files (`.js` + `.cjs`) match byte-for-byte. The one outlier (`chunks/DateField.es.js`) has identical imports in a different order — a semantic no-op (ES import order is irrelevant at runtime), same byte count, same set of named imports. This is Rollup module-graph non-determinism; nothing in the diff touches bundling. Type-file deltas are exactly the four fixes above: relocated `setup-jsdom-polyfills.d.ts`, two new `.d.cts` duplicates, 454 `.d.ts` files with `.js` extensions added to relative imports.
- **`@commercetools/nimbus-icons`** — all 4258 CJS files and 4256 of 4258 ESM `.js` files byte-identical. The 2 ESM outliers are the barrel index files where `.js` extensions were added to bare relative re-exports — exactly the rewrite that fixes native-Node-ESM consumers. The 2 new files are the `{"type":"module"}` / `{"type":"commonjs"}` marker `package.json` files.

## Test plan

- [x] `pnpm check:package-shape` — all three packages pass both linters in strict mode
- [x] `pnpm typecheck:strict`
- [x] `pnpm test:unit` (1174 tests)
- [x] `pnpm build:docs`
- [x] `pnpm lint`
- [ ] CI green

## Note for reviewers

The bundle-size baseline in `bundle-sizes.json` is currently ~22 KB stale on `main` (last refreshed May 6 with the Card refactor; multiple feature PRs since shrunk the bundle without triggering a baseline update because `check-bundle-size.mjs` only blocks on increases). Worth refreshing in a separate one-line housekeeping commit. Not introduced by this PR — verified that a clean build of `main` and a clean build of this branch produce byte-identical CJS output.


scripts/check-package-shape.mjs
```diff
@@ -0,0 +1,142 @@
...
+  const workDir = mkdtempSync(join(tmpdir(), "nimbus-pkg-shape-"));
+  const failures = [];
+
+  try {
+    for (const pkg of PACKAGES) {
+      header(pkg.name);
+
+      const distExists =
+        existsSync(join(pkg.dir, "dist")) ||
+        existsSync(join(pkg.dir, "package.json")); // some pkgs may not have dist/
```
ByronDWall: The `distExists` guard is unreachable. Every package directory has a `package.json`, so the `||` second clause is always true and the skip/warn branch on lines 96-98 can never execute.

Either remove the fallback clause (if the intent is "skip if no `dist/`") or change it to `existsSync(pkg.dir)` (if the intent is "skip if the package directory itself is missing"). Harmless in practice since `pnpm pack` would fail gracefully, but the dead branch + misleading comment is confusing.

docs/package-shape-verification.md
```diff
@@ -0,0 +1,155 @@
...
+
+## Quick Reference
+
+```bash
+# Run the check (CI runs this same command)
+pnpm check:package-shape
+```
+
+Requires the target packages to be built first
+(`pnpm build:packages && pnpm build:mcp`).
```
ByronDWall: The local reproduction command is missing the prerequisite `pnpm build:tokens` which must run first on a clean checkout. Should be:

```
pnpm build:tokens && pnpm build:packages && pnpm build:mcp
```

packages/nimbus/vite.config.ts
```diff
@@ -166,11 +166,15 @@ export default defineConfig(async () => {
...
+        // Note: src/test/setup-jsdom-polyfills.ts is a published entry point
+        // (see createEntries above), so it must be type-declared — only the
+        // other test-only files in src/test/ are excluded.
         exclude: [
           "src/**/*.stories.*",
           "src/**/*.spec.*",
           "src/**/*.test-*.*",
-          "src/test/**/*",
+          "src/test/unit-test-setup.ts",
+          "src/test/utils.tsx",
```
ByronDWall, line 182: The explicit exclude list here is coupled to the relocation logic in `scripts/postbuild-types.mjs`. If a new file is added to `src/test/` without updating both this list and the postbuild script, the new file's `.d.ts` will land in `dist/test/` and get silently deleted by `rmSync`.

Consider adding a cross-reference comment pointing to `scripts/postbuild-types.mjs` so future maintainers know both files must be updated together.

packages/nimbus/scripts/postbuild-types.mjs
```diff
@@ -0,0 +1,78 @@
...
+ *      moduleResolution: "nodenext". The shared rewriter lives at
+ *      scripts/lib/rewrite-relative-imports.mjs.
+ *
+ *   3. Duplicate index.d.ts and setup-jsdom-polyfills.d.ts to .d.cts variants
+ *      for the CJS exports path. Identical content; the extension flips how
+ *      TypeScript interprets the module kind (CJS regardless of `"type"`).
+ *
+ * Steps 1 & 2 run before step 3 so the .d.cts copies inherit the rewritten
+ * imports.
+ */
```
ByronDWall, line 29: Minor: this script is not idempotent — a second run without a prior `vite build` fails because `dist/test/` was deleted on the first run. That's fine for a postbuild step, but worth noting in the docblock so a future maintainer doesn't waste time debugging it.

ByronDWall (commented):
Review findings focused on accuracy, clarity, and build regression/performance risk.

scripts/check-package-shape.mjs
```diff
@@ -0,0 +1,142 @@
...
+  );
+  return passed;
+}
+
+function main() {
+  const workDir = mkdtempSync(join(tmpdir(), "nimbus-pkg-shape-"));
+  const failures = [];
+
+  try {
+    for (const pkg of PACKAGES) {
```
ByronDWall: Performance note: the sequential loop here takes ~26s locally (estimated 45–90s in CI), dominated by `attw` on the nimbus package (~920 `.d.ts` files). The three packages are independent and could be checked in parallel with `Promise.all` — swap `spawnSync` for async `spawn`, buffer each package's output, and print results in order after all finish. Expected speedup is ~2x since nimbus dominates the total time. Not blocking, but worth considering if the CI step becomes a bottleneck.

valoriecarli (approved):
Solid addition to the repo, thank you. Catching publish regressions before they reach consumers are ninjas...until they're not.

Ran through this a few times and didn't catch/notice anything Byron didn't already mention.  
But mostly it was the new label that won me over. 
🥹
Image

scripts/check-package-shape.mjs
misama-ct: Fixed [here](https://github.com/commercetools/nimbus/commit/2a28d5479).

docs/package-shape-verification.md
misama-ct: Fixed [here](https://github.com/commercetools/nimbus/commit/b8b366549).

packages/nimbus/vite.config.ts
misama-ct: Fixed [here](https://github.com/commercetools/nimbus/commit/af51e3fd7).

packages/nimbus/scripts/postbuild-types.mjs
misama-ct: Fixed [here](https://github.com/commercetools/nimbus/commit/eb43fec68).

scripts/check-package-shape.mjs
misama-ct: Fixed [here](https://github.com/commercetools/nimbus/commit/812bbd83a).

misama-ct:
@claude you think you can still find something or is this PR flawless now?

misama-ct:
Thanks for the second pass. Addressed all four items, one commit per item so each can be reviewed independently:

- **Coverage gap** — expanded `PACKAGES` from 3 to 4 (added `@commercetools/nimbus-design-token-ts-plugin`; it passes both attw and publint cleanly). Updated docs accordingly. The check ran against `nimbus-tokens` too and the reviewer's hunch was right: it fails attw, but the findings turned out to be a pre-existing CJS/`type: module` mismatch (preconstruct emits `.cjs.js` under a `"type": "module"` package, so Node treats the CJS path as ESM and `require()` is dead), not just a missing `types` condition. That's beyond what this PR set out to fix, so I excluded the package with a code comment + docs note and tracked the work in #1509. → [a14b3cb67](https://github.com/commercetools/nimbus/commit/a14b3cb67)
- **`nimbus-icons` `dev` script** — added `build:postbuild` to the chain so `pnpm dev` produces a tarball-compatible `dist/`. → [d39d83746](https://github.com/commercetools/nimbus/commit/d39d83746)
- **JSDoc `@example` regex** — added a sentence to the docblock of `rewrite-relative-imports.mjs` noting the behavior. → [fef4d276e](https://github.com/commercetools/nimbus/commit/fef4d276e)
- **Spawn-failure handling in `checkPackage`** — wrapped each `runTool` call so a binary-not-on-PATH throw becomes a synthetic failure for that tool instead of an unhandled rejection out of `Promise.all`. → [6885aaf9f](https://github.com/commercetools/nimbus/commit/6885aaf9f)

Local verification: 4 packages pass both checks, total wall time ~22s.

tylermorrisford (approved):
❤️

PR #1499, opened 2026-05-12
FEC-912: comment-chain baseline workflow with sticky PR comment

tylermorrisford:
## Summary

- Adds bundle size comment-chain workflow inline in `build-and-test.yml` — posts a sticky PR comment with per-package size deltas, using each merged PR's bot comment as the baseline for the next PR (no separate workflow or duplicate build)
- Simplifies package measurement to total `dist/` directory size per package instead of per-format entry-point patterns (esm/cjs), capturing declarations, dev/prod variants, and all build artifacts
- Refactors `scripts/check-bundle-size.mjs` to accept baseline via `BUNDLE_SIZE_BASELINE` env var, approval via `BUNDLE_SIZE_APPROVED` env var, and support `--json` output mode for the workflow to consume
- Adds label-based approval: apply `bundle-size-approved` label to bypass threshold failures (replaces the old commit-file approval flow)
- Adds `.github/workflows/bundle-size-comment.yml` — a lightweight post-merge workflow that applies the `bundle-sizes` label to merged PRs
- Bootstrap: falls back to `bundle-sizes.json` when no labeled merged PR exists yet

Closes FEC-912

## How it works

1. On PR open/push, the workflow queries the most recent merged PR with the `bundle-sizes` label
2. Extracts the machine-readable data block (`<!-- bundle-sizes-data-v1: {...} -->`) from that PR's bot comment
3. Passes the baseline to `check-bundle-size.mjs` via env var
4. Posts/updates a sticky comment with a markdown table showing per-package sizes, deltas, and statuses
5. If any package exceeds its threshold, CI fails — apply `bundle-size-approved` label to approve
6. On merge, the `bundle-sizes` label is added automatically, making this PR's sizes the baseline for the next

## Test plan

- [ ] Open a PR with no size change → comment posts with all packages at "ok"
- [ ] Push a >5% increase → comment updates (not duplicates), shows "fail" status, CI fails
- [ ] Apply `bundle-size-approved` label, re-run CI → passes
- [ ] Merge the PR → `bundle-sizes` label added automatically
- [ ] Open a second PR → workflow reads the first merged PR's comment as baseline
- [ ] On fresh repo with no labeled PRs → falls back to `bundle-sizes.json`
- [ ] Workflow YAML has no `git push`, `git commit`, or `contents: write`


ByronDWall (commented):
## Bundle Size Workflow Review

Reviewed with a focus on **effectiveness and workflow reliability**. Found 4 critical issues, 4 significant warnings, and 2 minor issues. Details in inline comments below.

---

**Issue not in diff — Stale failure message in `scripts/check-bundle-size.mjs` (minor)**

Lines 346-352 still tell users to run `node scripts/update-bundle-sizes.mjs` and commit `bundle-sizes.json`. Under the new workflow, users should apply the `bundle-size-approved` label and re-run CI. The JSON output path (line 144 in the workflow) already has the correct message, but the human-readable path does not.

.github/workflows/build-and-test.yml
```diff
@@ -20,15 +21,163 @@ jobs:
...
+            BASELINE_DATA=$(gh api \
+              "repos/${GITHUB_REPOSITORY}/issues/${BASELINE_PR}/comments" \
+              --jq '.[] | select(.body | contains("bundle-sizes-comment")) | .body' \
+              | grep -o '<!-- bundle-sizes-data-v1: .* -->' \
+              | sed 's/<!-- bundle-sizes-data-v1: //' \
+              | sed 's/ -->//')
+
+            if [ -n "$BASELINE_DATA" ]; then
+              echo "source=comment-chain" >> "$GITHUB_OUTPUT"
+              echo "data=${BASELINE_DATA}" >> "$GITHUB_OUTPUT"
```
ByronDWall: ### Critical: GITHUB_OUTPUT multiline injection

`BASELINE_DATA` is a JSON string containing `{`, `}`, `"`, `:` characters. Writing it with `echo "data=${BASELINE_DATA}"` will corrupt the `GITHUB_OUTPUT` file if the value ever contains a newline (e.g. if the source comment was line-wrapped or edited by a human).

Use the heredoc delimiter pattern instead:
```bash
echo "data<<EOF" >> "$GITHUB_OUTPUT"
echo "${BASELINE_DATA}" >> "$GITHUB_OUTPUT"
echo "EOF" >> "$GITHUB_OUTPUT"
```

.github/workflows/build-and-test.yml
```diff
@@ -20,15 +21,163 @@ jobs:
...
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          BASELINE_PR=$(gh pr list \
+            --repo "$GITHUB_REPOSITORY" \
+            --state merged \
+            --label bundle-sizes \
+            --json number \
+            --limit 1 \
+            --jq '.[0].number // empty')
```
ByronDWall: ### Critical: `gh pr list` sort order is creation-date, not merge-date

`gh pr list` defaults to sorting by `created` date, not `merged_at`. A long-lived draft PR that finally merges could be returned here, serving a stale baseline.

Switch to `gh api` with explicit sorting:
```bash
BASELINE_PR=$(gh api \
  "repos/${GITHUB_REPOSITORY}/pulls?state=closed&labels=bundle-sizes&per_page=1&sort=updated&direction=desc" \
  --jq '[.[] | select(.merged_at != null)] | .[0].number // empty')
```

.github/workflows/build-and-test.yml
```diff
@@ -20,15 +21,163 @@ jobs:
...
+
+          if [ -n "$BASELINE_PR" ]; then
+            echo "Found baseline PR: #${BASELINE_PR}"
+
+            BASELINE_DATA=$(gh api \
+              "repos/${GITHUB_REPOSITORY}/issues/${BASELINE_PR}/comments" \
+              --jq '.[] | select(.body | contains("bundle-sizes-comment")) | .body' \
+              | grep -o '<!-- bundle-sizes-data-v1: .* -->' \
+              | sed 's/<!-- bundle-sizes-data-v1: //' \
+              | sed 's/ -->//')
```
ByronDWall: ### Critical: Brittle HTML comment extraction

`grep -o '<!-- bundle-sizes-data-v1: .* -->'` has two problems:
1. `.*` is greedy — if multiple HTML comments appear on the same line, it matches from the first `<!--` to the last `-->`.
2. The `grep` + double `sed` pipeline silently returns empty on any formatting variation (extra whitespace, line wrapping), causing a silent fallback to bootstrap with no error.

Consider a more robust approach: extract the data in a single `sed` or `awk` expression, or parse via an inline Node.js script where JSON parsing gives you proper error handling.

.github/workflows/build-and-test.yml
```diff
@@ -20,15 +21,163 @@ jobs:
...
+          EXISTING_COMMENT_ID=$(gh api \
+            "repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/comments" \
+            --jq '.[] | select(.body | contains("bundle-sizes-comment")) | .id' \
+            | head -1)
+
+          if [ -n "$EXISTING_COMMENT_ID" ]; then
+            gh api \
+              --method PATCH \
+              "repos/${GITHUB_REPOSITORY}/issues/comments/${EXISTING_COMMENT_ID}" \
+              -f body="$COMMENT_BODY"
```
ByronDWall: ### Critical: Shell quoting in `gh api -f body="$COMMENT_BODY"`

The `COMMENT_BODY` variable contains backticks (from the `` `bundle-size-approved` `` label reference) and potentially other special characters. Shell interpolation inside `"$COMMENT_BODY"` can corrupt the value or cause syntax errors.

Write the comment body to a temp file and pass it via file reference:
```bash
echo "$COMMENT_BODY" > /tmp/comment-body.txt
gh api --method PATCH \
  "repos/${GITHUB_REPOSITORY}/issues/comments/${EXISTING_COMMENT_ID}" \
  -F body=@/tmp/comment-body.txt
```
Or better yet, have the Node script write the body directly to a file to avoid shell variable assignment entirely. The same issue applies to the POST path on line 171.

.github/workflows/bundle-size-comment.yml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  post-merge-label:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    steps:
+      - name: Add bundle-sizes label
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh pr edit "${{ github.event.pull_request.number }}" \
```
ByronDWall: ### Warning: Race condition on rapid successive merges

This workflow runs asynchronously after merge. If two PRs merge within seconds of each other, the second PR's CI run could start before the first merge's labeling workflow completes. The baseline fetch won't find the first PR's label and falls back to bootstrap/git.

This degrades gracefully (uses stale-but-valid baseline), but could produce noisy false positives on size delta if the bootstrap baseline is significantly outdated. If this becomes a problem, consider adding a short retry/delay in the baseline fetch step.

.github/workflows/build-and-test.yml
```diff
@@ -20,15 +21,163 @@ jobs:
...
+              --method POST \
+              "repos/${GITHUB_REPOSITORY}/issues/${PR_NUMBER}/comments" \
+              -f body="$COMMENT_BODY"
+            echo "Posted new comment"
+          fi
 
-      - name: Check bundle size
-        run: node scripts/check-bundle-size.mjs
+      - name: Fail if bundle size thresholds exceeded
+        if: github.event_name == 'pull_request' && steps.bundle-size-check.outputs.exit_code != '0' && steps.approval.outputs.approved != 'true'
```
ByronDWall: ### Warning: Fail step fires when `exit_code` output is unset

If the "Measure and compare bundle sizes" step crashes mid-execution (Node segfault, OOM kill), the `exit_code` output is never written to `GITHUB_OUTPUT`. Then `steps.bundle-size-check.outputs.exit_code` is empty string, and `'' != '0'` evaluates to **true**, triggering this failure step with a misleading "Apply the label" error message.

Check for the explicit failure code instead:
```yaml
if: >-
  github.event_name == 'pull_request'
  && steps.bundle-size-check.outputs.exit_code == '1'
  && steps.approval.outputs.approved != 'true'
```

.github/workflows/build-and-test.yml
```diff
@@ -20,15 +21,163 @@ jobs:
...
+            --json number \
+            --limit 1 \
+            --jq '.[0].number // empty')
+
+          if [ -n "$BASELINE_PR" ]; then
+            echo "Found baseline PR: #${BASELINE_PR}"
+
+            BASELINE_DATA=$(gh api \
+              "repos/${GITHUB_REPOSITORY}/issues/${BASELINE_PR}/comments" \
+              --jq '.[] | select(.body | contains("bundle-sizes-comment")) | .body' \
```
ByronDWall: ### Warning: Multiple comments matching the magic string

The jq filter `select(.body | contains("bundle-sizes-comment"))` matches *any* comment containing the marker — including human comments that quote or reference the bot comment. The baseline data extraction doesn't take only the first match, so `grep -o` could return concatenated results from multiple comments.

Filter by the GitHub Actions bot login: `select(.user.login == "github-actions[bot]")`, or pipe the `gh api` output through `| head -1` before the `grep` to ensure only the first matching comment body is processed.

.github/workflows/build-and-test.yml
```diff
@@ -20,15 +21,163 @@ jobs:
...
+        if: github.event_name == 'pull_request'
+        run: |
+          (type -p wget >/dev/null || (apt-get update && apt-get install -y wget)) \
+          && mkdir -p -m 755 /etc/apt/keyrings \
+          && out=$(mktemp) && wget -nv -O "$out" https://cli.github.com/packages/githubcli-archive-keyring.gpg \
+          && cat "$out" | tee /etc/apt/keyrings/githubcli-archive-keyring.gpg > /dev/null \
+          && chmod go+r /etc/apt/keyrings/githubcli-archive-keyring.gpg \
+          && echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" | tee /etc/apt/sources.list.d/github-cli.list > /dev/null \
+          && apt-get update \
+          && apt-get install -y gh
```
ByronDWall: ### Warning: apt-based gh CLI install adds ~20-30s and a network dependency to every PR build

Every PR build runs `apt-get update` twice and downloads from `cli.github.com`. If that endpoint is slow or rate-limited, the entire build fails at this step.

A static binary download is faster and more reliable:
```bash
curl -sL https://github.com/cli/cli/releases/download/v2.74.0/gh_2.74.0_linux_amd64.tar.gz | tar xz
export PATH="$PWD/gh_2.74.0_linux_amd64/bin:$PATH"
```
Or include `gh` in a custom Docker image to eliminate the install step entirely.

scripts/check-bundle-size.spec.mjs
```diff
@@ -101,16 +101,13 @@ describe("check-bundle-size failure detection", () => {
   withBaselineGuard();
 
   it("exits 1 when baseline sizes are much smaller than actual", () => {
```
ByronDWall: ### Minor: No test coverage for the new `BUNDLE_SIZE_BASELINE` env var path

The new primary baseline source (`BUNDLE_SIZE_BASELINE` env var → `resolveBaseline()` → `"comment-chain"` source) has no test coverage. Consider adding tests that:
1. Set `BUNDLE_SIZE_BASELINE` to a valid JSON string and verify the script uses it as the baseline
2. Set `BUNDLE_SIZE_BASELINE` to malformed JSON and verify exit code 1
3. Verify `BUNDLE_SIZE_APPROVED=true` causes threshold exceedances to report as `approved` rather than `fail`

.github/workflows/build-and-test.yml
tylermorrisford: commit [cf87584](https://github.com/commercetools/nimbus/pull/1499/commits/cf87584129df598cc79331a8404e2365ad0e956e) addresses all comments other than the race condition; I'd like to see if that becomes a common issue before adding a delay.

valoriecarli (approved):

ByronDWall (approved):

PR #1536, opened 2026-05-20
fix: rm bundle fallback file, clarify docs

tylermorrisford:
This PR removes the `bundle-sizes.json` file, updates scripts/docs appropriately, and clarifies that this file is only meant to "seed" the comment chain, and then be removed. Also colocates the tests with the scripts in the `.github/actions/bundle-size` directory.


valoriecarli (approved):
🥲

ByronDWall (approved):

PR #1556, opened 2026-05-29
feat(drag-and-drop): cross-list DnD, useDragAndDrop hook, and DataTable integration

### LINKED ISSUE #1554: DraggableList: no way to opt out of cross-list drops
## Summary

`DraggableList.Root` accepts drops from every other `DraggableList` in the page, with no public API to opt out. Consumers rendering two (or more) lists that should be independent have no way to disable cross-list dropping.

cc @ByronDWall

## Where it's hardcoded

In `packages/nimbus/src/components/draggable-list/components/draggable-list.root.tsx`:

- A single module-level constant `DRAGGABLE_LIST_DATA_FORMAT = "nimbus-draggable-list-item"` is used as both the drag payload type and `acceptedDragTypes` — so every Nimbus DraggableList implicitly accepts every other Nimbus DraggableList.
- `getDropOperation: () => "move"` is hardcoded — there's no path to return `"cancel"` for foreign sources.
- The component constructs its own `dragAndDropHooks` internally and assigns it after spreading consumer props, so a consumer-supplied `dragAndDropHooks` is silently overridden.
- `DraggableListRootProps` exposes none of these knobs.

The `CrossListDragAndDrop` story confirms cross-list movement is the intended default — but there's no way to opt out.

## Repro

```tsx
<>
  <DraggableList.Root aria-label="A" items={a} onUpdateItems={setA} />
  <DraggableList.Root aria-label="B" items={b} onUpdateItems={setB} />
</>
```

Items can be dragged from A into B and vice versa, with no consumer-controlled way to prevent it.

## Proposed API

Any one of the following would close the gap; my preference is the first:

1. **`dragNamespace?: string`** prop — when set, the list uses `\`nimbus-draggable-list-item:${dragNamespace}\`` as its drag type and only accepts that exact type. Lists with different namespaces won't accept each other's items. Backward-compatible: unset preserves today's behavior.
2. **`acceptedDragTypes?: string[]`** + a way to set the outgoing drag type — direct passthrough to `useDragAndDrop`. More flexible, more surface area.
3. **`getDropOperation?: (...) => DropOperation`** override — lets consumers return `"cancel"` based on source/target inspection.

(1) covers the "two independent lists on one page" case with one prop and zero new concepts for consumers. (2)/(3) are useful for more advanced cases (custom drag sources, conditional acceptance) and could come later.

## Workaround today

There isn't a clean one. App-level revert in `onUpdateItems` (detect foreign items, restore prior state) works but flashes the wrong state briefly. Forking the component is the only way to get correct UX.

ByronDWall:
Closes #1554

## Motivation

`DraggableList.Root` uses a single hardcoded drag type (`nimbus-draggable-list-item`) for all instances on a page. Every list accepts every other list's items, with no public API to opt out. Consumers rendering two independent lists — say, a "selected columns" list and an "available filters" list — have no way to prevent items from leaking between them. The `getDropOperation` callback is hardcoded to `"move"` and the internally-constructed `dragAndDropHooks` silently overrides any consumer-supplied value.

The only workaround today is reverting foreign items inside `onUpdateItems` (which flashes the wrong state) or forking the component entirely.

This PR solves the problem with a `dragNamespace` prop (option 1 from the issue) and, while doing so, extracts the drag-and-drop wiring into a shared `useDragAndDrop` hook that also works with `DataTable` and any other React Aria collection component.

### New `useDragAndDrop` hook

Rather than fixing `DraggableList` in isolation, the DnD configuration is extracted into a reusable hook. It wraps React Aria's `useDragAndDrop` with:

- **Namespace isolation** — `dragNamespace` appends a suffix to the internal drag type, so only collections sharing the same namespace can exchange items. Omitting it preserves backward-compatible behavior where all lists interoperate.
- **External drop support** — `onExternalDrop` + `acceptExternalTypes` let a collection accept drops from outside the app (text, files, etc.). `resolveDropOperation` validates incoming types against `acceptExternalTypes` before returning the configured operation.
- **Outgoing format serialization** — `serializeDragItem` adds custom MIME representations alongside the internal format when items are dragged out.
- **Auto-keying** — items returned from `onExternalDrop` without `key` or `id` get a `crypto.randomUUID()` key so React Aria's collection can track them.

### State handler factories

Consumers need to wire up insert/append/reorder/remove callbacks. Two factories make this a one-liner:

- **`createArrayHandlers(setItems, getKey?)`** — for `useState` arrays
- **`createListDataHandlers(list)`** — for React Aria's `useListData`

Both return an object meant to be spread into the hook options.

### External drop helpers

Six composable utilities for converting raw `DropItem[]` inside `onExternalDrop`:

| Helper | Accepts | Produces |
|--------|---------|----------|
| `createItemsFromTextDrop` | `text/plain` (or custom) | `{ key, label }` |
| `createItemsFromFileDrop` | File drops | `{ key, label, fileName, fileType, file }` |
| `createItemsFromDirectoryDrop` | Directory drops | Flattened files with `{ directory }` |
| `createItemsFromJsonDrop` | `application/json` | `{ key, label, data }` |
| `createItemsFromCsvDrop` | `text/csv` | `{ key, label, headers, row }` |
| `createItemsFromImageDrop` | Image file drops | `{ key, label, fileType, objectUrl, file }` |

### DraggableList changes

- New props: `dragNamespace`, `onExternalDrop`, `acceptExternalTypes`, `externalDropOperation`, `serializeDragItem`
- Internal implementation now delegates to the shared hook instead of calling React Aria directly
- `DraggableList.Root` continues to manage its own `useListData` state; the hook handles the drag-and-drop configuration

### DataTable integration

The DataTable previously had no drag-and-drop support. This PR adds:

- Optional `dragAndDropHooks` prop on both `DataTable` and `DataTable.Table`
- When provided, a sticky drag handle column (`DragIndicator` icon) is automatically prepended
- Sticky column offset cascade handles all combinations of drag + selection + expand columns in base, hover, and pinned row contexts
- `draggable="true"` on rows suppresses native double-click text selection; a programmatic `selection.modify()` workaround restores it using `caretPositionFromPoint` (Chrome/Firefox) with `caretRangeFromPoint` fallback (Safari)
- Row click handler uses native DOM capture-phase listeners to coexist with React Aria's press handling

## Architecture

```
hooks/use-drag-and-drop/
├── use-drag-and-drop.ts        # Main hook + resolveDropOperation
├── use-drag-and-drop.types.ts  # DragAndDropProps, UseDragAndDropOptions, DragAndDropItemData
├── state-handlers.ts           # createArrayHandlers, createListDataHandlers
├── process-drop-items.ts       # Internal/external item separation + auto-keying
├── external-drop-helpers.ts    # Six composable drop converters
├── index.ts                    # Public barrel export
├── use-drag-and-drop.mdx       # Hook documentation
├── use-drag-and-drop.spec.ts   # Unit tests (63 tests)
├── use-drag-and-drop.stories.tsx
└── utils/
    └── use-drag-and-drop.test-utils.ts
```

`resolveDropOperation` is extracted as a pure function so the drop-operation decision logic (internal vs. external type validation vs. cancel) is unit-testable independent of React.

## Testing

**Unit tests** (63 tests in `use-drag-and-drop.spec.ts`):
- `createArrayHandlers` — insert, append, reorder, remove, default key extraction, missing key error
- `createListDataHandlers` — all delegation paths
- `processDropItems` — internal deserialization, external routing, auto-keying, corrupted JSON
- All six external drop helpers — type filtering, field extraction, edge cases
- `resolveDropOperation` — internal move/cancel, external type accepted/rejected, operation allowed/denied

**Storybook interaction tests**:
- Reorder within a single list
- Namespace isolation (reorder alpha, verify beta unchanged)
- Cross-list transfer (drag from source to target)
- DataTable integration (drag handle rendering, keyboard reorder, row click coexistence)

## Test plan

- [x] Unit tests pass (`pnpm test:dev` — 110 tests across 4 files)
- [x] TypeScript compiles cleanly (`pnpm typecheck`)
- [ ] Storybook stories render for `useDragAndDrop`, `DraggableList`, and `DataTable` DnD variants
- [ ] Keyboard drag-and-drop works (Enter to grab, Arrow to move, Enter to drop)
- [ ] Cross-list transfer moves items between lists sharing a namespace
- [ ] Namespace isolation prevents drops between lists with different namespaces
- [ ] DataTable drag handles appear when `dragAndDropHooks` is provided
- [ ] Pinned rows + drag column sticky offsets render correctly
- [ ] Double-click text selection works on draggable DataTable rows


packages/nimbus/src/components/data-table/components/data-table.row.tsx
```diff
@@ -145,12 +149,7 @@ const DataTableRowInner = <T extends DataTableRowItem = DataTableRowItem>({
       if (!onRowClick) return;
       // Prevent row click when clicking on interactive elements to avoid conflicts
       const isInteractiveElement = getIsTableRowChildElementInteractive(e);
-      // Prevent row click when text is selected
-      const hasSelectedText =
```
tylermorrisford, line 149: from claude: 

hasSelectedText guard removed unconditionally — regression for non-draggable rows

In data-table.row.tsx, the old handleRowClick path skipped navigation if the user had selected text (!hasSelectedText). That guard is gone now. On any row with onRowClick — draggable or not — a mouse-up after text selection will trigger navigation. The dblclick programmatic-selection workaround only applies to [draggable='true'] rows, so non-draggable rows lose the protection.

But honestly if the row has an onRowClick, shouldn't the navigation or other onRowClick operations take precedence over selecting text?

tylermorrisford (approved):
one Q, otherwise 💰

misama-ct (approved):

packages/nimbus/src/components/data-table/components/data-table.row.tsx
ByronDWall: good catch,  Tylaude!

PR #1560, opened 2026-05-29
feat(splitter): add resizable two-pane Splitter component + useResponsiveSplitterSizes hook

misama-ct:
## Summary

Adds `Splitter`, a WCAG 2.1 AA compliant compound component for a user-resizable
two-pane layout: a configurable `Splitter.Aside` and a `Splitter.Main` that fills
the remaining space, with a draggable, keyboard-operable handle between them — plus
`useResponsiveSplitterSizes`, a companion hook that lets consumers drive pane sizes
in pixels, size tokens, or per-container-width breakpoints.

> **Note:** this PR previously consisted of two stacked PRs (the component and the
> hook). They were combined: the Splitter is unreleased and our first consumer needs
> the hook alongside it, so component + hook now ship as one PR, one changeset, and
> two OpenSpec changes. The former hook PR (#1587) is folded in here.

## Motivation

Nimbus has no primitive for user-resizable panes. Consumers building IDE-like or
sidebar-plus-content layouts otherwise hand-roll the drag math, keyboard handling,
and the W3C window-splitter ARIA contract — error-prone and rarely accessible.
`Splitter` does that one job well, with clean seams for persistence and collapse.

The component is deliberately **percentage-native** (the boundary is one number,
0–100). But real layouts are often specified in pixels ("a 320px sidebar") or per
breakpoint. `useResponsiveSplitterSizes` is the pixel/token → percentage translator
that keeps the component simple while giving consumers the units they actually design in.

### Component

New component at `packages/nimbus/src/components/splitter/` (registered in the
component barrel and the theme slot-recipes), built on React Aria primitives
(`useSeparator`, `useMove`, `useFocusRing`) and a Chakra slot recipe.

**Compound API:** `Splitter.Root`, `Splitter.Aside`, `Splitter.Main`,
`Splitter.Handle`.

- **One dimension to configure.** A 2-pane splitter has a single boundary, so the
  size is one number — the aside's percentage (`defaultSize`, or controlled
  `size`). The main pane is always the remainder. The role is designated by the
  component type, so there are no ids to wire and no pane-keyed config.
- **Either-side placement.** The aside may be rendered before or after the main
  pane (a left/top nav or a right/bottom panel); `size` always refers to the
  aside, and the handle's ARIA value tracks the leading pane.
- **Flat constraints on `Root`.** `minSize` / `maxSize` bound the aside; `maxSize`
  also fixes the main pane's floor (`100 − maxSize`).
- **Collapsible aside.** `collapsible` + `collapsedSize`, controllable via the
  `collapsed` boolean (or `defaultCollapsed`); Enter on the focused handle toggles
  it, double-click restores the mount-time size.
- **Persistence is consumer-wired.** Hydrate `defaultSize` from any storage and
  write back the single number from `onSizeChangeEnd` (settle-only; `onSizeChange`
  is the live, per-tick channel). No bundled hook, no `autoSaveId`.
- **Orientation** (`horizontal` / `vertical`), `keyboardStep`,
  `isDoubleClickDisabled`, and `isDisabled` round out `Root`.
- Three-or-more regions are composed by nesting a `Splitter` inside a pane.

### `useResponsiveSplitterSizes` hook

An opt-in companion hook that returns `rootProps` to spread onto `Splitter.Root`. It
measures the container with a `ResizeObserver`, translates the consumer's config into
the percentage the component consumes, clamps to `minSize` / `maxSize`, drives the
settle-only controlled `size` channel (no snap-back), and can persist the user's
settled size across reloads (pixel-first, per-band, versioned).

- **`number` = pixels. Period.** A bare number is always a pixel value; a `"30%"`
  string is a passthrough percentage; a size token (`3xs`–`8xl`, `breakpoint-*`)
  resolves against the theme.
- **Per-container-width breakpoints** via object notation, e.g. `{ 0: 320, 768: "30%" }`
  — same notation supported for `minSize` / `maxSize` / `collapsedSize`.
- **Synchronous first paint** for configs that need no measurement (`%` values), so
  there's no uncontrolled-default flash.
- Degrades gracefully without `ResizeObserver` / `localStorage`.

```tsx
import { Splitter, useResponsiveSplitterSizes } from "@commercetools/nimbus";

function AppLayout() {
  // 320px sidebar below a 768px container width, 30% above; persisted across reloads.
  const { rootProps } = useResponsiveSplitterSizes({
    persistKey: "app:main-splitter",
    size: { 0: 320, 768: "30%" },
    minSize: 240,
    maxSize: "50%",
  });

  return (
    <Splitter.Root {...rootProps} collapsible>
      <Splitter.Aside>{/* nav */}</Splitter.Aside>
      <Splitter.Handle />
      <Splitter.Main>{/* content */}</Splitter.Main>
    </Splitter.Root>
  );
}
```

## How it works

`Splitter.Root` owns the aside-size state machine (synchronous mount-time seeding,
live vs. settled change channels, controlled/uncontrolled size and collapse with
reconciliation) and exposes it via context. The panes register their role + DOM id;
the handle resolves the leading/trailing siblings, applies clamped resize, and renders
the W3C separator ARIA model (`role="separator"`, `aria-valuenow/min/max/text`,
`aria-orientation`, `aria-controls`). Sizes carry full float precision end-to-end;
only `aria-valuenow` is rounded for AT.

The hook sits entirely outside the component as a pure translator: all pixel math,
container-width resolution, hook-side clamping, and per-band persistence live in
`hooks/use-responsive-splitter-sizes.ts` + `utils/` (pure resolvers, size-token
lookup, storage adapter). The component never learns about pixels.

## Test plan

- [x] `pnpm --filter @commercetools/nimbus typecheck` — clean
- [x] `pnpm --filter @commercetools/nimbus build` (incl. theme typings) — clean
- [x] `pnpm test:dev packages/nimbus/src/components/splitter/` — **103 passing across 12 files**
      (util/hook specs + Storybook play functions + consumer docs.spec)
- [x] `pnpm lint` — clean
- [x] `openspec validate add-splitter-component --strict` — valid
- [x] `openspec validate add-responsive-splitter-sizes-hook --strict` — valid

Component play functions cover default/vertical layouts, aside-trailing (either-side),
pointer drag, keyboard (arrows + Home/End), min/max clamping, collapse (keyboard +
controlled), resize-lock while collapsed, double-click restore (incl. 0%), controlled
size, persistence, nesting, float precision, and disabled states.

Hook coverage: synchronous `%` first paint, px→% conversion, container-width band
selection + hysteresis, object-notation `minSize`/`maxSize`, vertical (height) axis,
emit gating, pixel-first persistence round-trip across remount + resize, collapse-driven
settle suppression, and graceful degradation without `ResizeObserver`/storage.

### Interactive Storybook stories

In this PR's Storybook preview (Vercel check below), under **Components › Splitter**:

- `ResponsivePixelSizesHook` — fixed-width container, px config.
- `ResponsiveByContainerWidth` — two static containers (640 vs 1000px) showing band switch.
- `NestedResponsiveSplitter` — **interactive**: drag the outer handle to resize the inner
  container across the 768px threshold and watch the inner aside switch bands live.

## Feedback wanted

- The aside is the single configured pane and `maxSize` expresses the main pane's
  floor as its complement — does that read clearly at the call site, or would a
  named main-floor prop be worth the extra surface?
- The hook treats a bare `number` as pixels unconditionally — clear enough given the
  component is percentage-native, or worth a more explicit `{ px: n }` form?

Closes FEC-977


misama-ct:
@claude review this branch

misama-ct:
Thanks for the review @claude — addressed in 1f99806aa:

- **🟠 aria-valuenow out of range on collapse** — fixed, but by widening the bounds rather than clamping the value. `aria-expanded` isn't valid on `role="separator"` and clamping `valuenow` would misreport the collapsed position (collapse is conveyed *only* through the value). So `aria-valuemin`/`valuemax` now reflect the true minimum reachable boundary position (`collapsedSize` for a collapsible pane), keeping `valuenow` truthful and in-range. Added an in-range regression assertion to `CollapsibleByKeyboard`.
- **🟡 controlled-collapse + restore** — `restoreDefaults` now only resets internal collapse refs when uncontrolled; in controlled mode it just fires `onCollapsedPaneChange(null)` and lets the prop drive.
- **🟡 paneCount warning during render** — moved into a `useEffect` so transient 1-pane commits don't trigger a false warning.
- **Nits** — removed the redundant `role="separator"` (`useSeparator` already supplies it) and fixed the drag-accumulation comment.
- **Pointer-drag coverage** — added a `useMove` drag play function (driven via `fireEvent` pointer events with a stable `pointerId`).

i18n/LOC ticket is being handled separately. Typecheck, lint, and the full splitter suite (39 tests) are green.

misama-ct:
Girl, I traced it — the controlled-restore follow-up is already handled. 🟡

Walking your exact scenario (controlled collapse → drag to 50/50 → collapse → double-click) it lands on the **mount-time 30/70, not the pre-collapse 50/50**. The piece the trace misses: when `restoreDefaults` calls `commitSizes(initial)`, the drag-expand detection in `writeSizes` (`use-splitter-state.ts:125-134`) sees the collapsed pane grow back above its `collapsedSize` and **clears `appliedCollapseRef`/`preCollapseSizesRef`**. So when the consumer then sets `collapsedPane = null`, the reconcile effect sees `cur === prev === null` and no-ops — it never reaches the `cur === null` branch that would re-apply pre-collapse sizes. It only diverges in a degenerate config where a pane's mount size already equals its `collapsedSize`.

Added a regression story `ControlledCollapseRestore` (7f6e2c6bf) that reproduces the scenario and asserts the restore lands on 30 — also fills the controlled-collapse + double-click coverage gap you flagged.

🔵 On `aria-valuemin` understating the draggable floor: agreed, the reachable set is `{0} ∪ [10, 80]` announced as `[0, 80]`. That's the inherent cost of mapping a discrete collapse onto a continuous `separator` value, and keeping `valuenow` truthful is the right priority — noted as a known AT nuance.

Splitter suite green (16 stories + specs).

packages/nimbus/src/components/splitter/components/splitter.pane.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  useEffect(() => {
+    if (process.env.NODE_ENV !== "production") {
+      if (!id) {
+        console.warn(
+          "[Splitter] <Splitter.Pane> is missing the required `id` prop. Each pane must have a unique id; the pane will be excluded from state."
+        );
+        return;
+      }
+    }
```
misama-ct: I'll prepare a devWarn util, as I have seen this before in other places. Didn't want to bloat this PR any further.

misama-ct:
@claude I needed to add a slight change to the plan. When a pane is collapsed, it can not be resized anymore by dragging as minSize and collapsedSize may be different. Please have a look, all other issues should be resolved or mitigated by now.

misama-ct:
@claude in the beginning we said we do not want controlled sizes due to potential re-render issues, however, to be able to react viewport- or container-query-changes, there needs to be a way to control the component from the outside, thus we introduced the `sizes` property. Have a look, let me know your thoughts.

packages/nimbus/src/components/splitter/hooks/use-handle-resize.ts
```diff
@@ -0,0 +1,99 @@
...
+      const parent = handleRef.current?.parentElement;
+      if (!parent) return;
+      const containerSize =
+        orientation === "horizontal" ? parent.offsetWidth : parent.offsetHeight;
+      if (containerSize <= 0) return;
+      const deltaPx = orientation === "horizontal" ? e.deltaX : e.deltaY;
+      dragAccumRef.current += (deltaPx / containerSize) * 100;
+      const wholeDelta = dragAccumRef.current;
+      if (Math.abs(wholeDelta) < MOVE_TOLERANCE) return;
+      dragAccumRef.current = 0;
```
tylermorrisford: [critical] Drag accumulator discards sub-tolerance remainder on every tick.

`dragAccumRef.current = 0` zeroes the accumulator after applying a delta instead of carrying the remainder forward. The classic carry-forward pattern is:

```ts
dragAccumRef.current -= wholeDelta;
```

As written, sub-pixel remainders never accumulate across pointer-move events. The comment on line 70 describes this as intended to "build up until they clear the gate" — but the gate resets to zero instead of to the leftover fraction, so slow or noisy input devices see no accumulation between events.

packages/nimbus/src/components/splitter/splitter.types.ts
```diff
@@ -0,0 +1,401 @@
...
+/**
+ * Props for `<Splitter.Handle>` — the interactive separator between the aside
+ * and main panes. The handle takes no per-handle *behaviour* configuration:
+ * resize and collapse behaviour is configured on `<Splitter.Root>`
+ * (`keyboardStep`, `isDoubleClickDisabled`, `isDisabled`). It still accepts
+ * standard DOM and style props via the slot — notably an `id` (e.g. a persistent
+ * id for analytics) and `aria-label` / `aria-labelledby` to override the
+ * localized default ("Resize panes").
+ */
+export type SplitterHandleProps = OmitInternalProps<SplitterHandleSlotProps> & {
```
tylermorrisford, line 266: [major] `SplitterHandleProps` accepts `id`, contradicting the Anonymous Handle requirement.

Spec 1.9 states: _"TypeScript SHALL emit a compile error when consumer attempts `<Splitter.Handle id="…">`"_. Because `SplitterHandleSlotProps = HTMLChakraProps<"div">`, `id` is fully accepted — and the JSDoc comment above (lines 263–264) actively invites passing it for analytics.

Two options:
1. **Enforce anonymity (spec-compliant):** Apply `Omit<SplitterHandleSlotProps, "id">` and remove the analytics reference from the JSDoc.
2. **Relax the spec:** Update Spec 1.9 to permit `id` on the Handle DOM element and clarify it is distinct from the pane ids used by `aria-controls`.

The current state (type permits `id`, docs encourage it, spec forbids it) is inconsistent and needs a decision.

packages/nimbus/src/components/splitter/splitter.stories.tsx
```diff
@@ -0,0 +1,1328 @@
...
+      expect(Number(handle.getAttribute("aria-valuenow"))).toBe(0);
+    });
+  },
+};
+
+// ============================================================
+// Float-precision — fractional percentages are applied unrounded.
+// ============================================================
+
+export const FloatPrecision: Story = {
```
tylermorrisford, line 524: This test finds `null` instead of `'31'` for some reason

ByronDWall (approved):
Some minor nits that I took the liberty of updating around using design tokens and consistent names for the panels in the docs, along with fixing the issue @tylermorrisford mentioned. As always, lmk if youd prefer I pr changes in a child branch or just leave comments.

This is a phenomenal component, thanks for building it!

tylermorrisford (approved):
🍌

PR #1596, opened 2026-06-11
feat(region): headless primitive for named content projection

misama-ct:
## The story

The goal was to **display arbitrary content inside a Splitter pane** — an agent panel in the aside, say — from a component that lives somewhere else entirely. That's really just **portalling**, and it has nothing to do with splitting a view: the Splitter is already a dense component (size state machine, controlled/uncontrolled collapse, settle-vs-live channels, responsive sizing, ARIA), and the capability is generic enough to reuse elsewhere (a drawer, a tab panel, the next component). So I built it as a standalone, headless **`Region`** primitive — and the Splitter doesn't even need to know it exists.

The Splitter already exposes collapse through its controlled `collapsed` / `onCollapsedChange` props. So instead of teaching the Splitter about regions, a `Region` **target** publishes a **value** (the toggle callbacks + state), and the **app shell** composes the two. The Splitter stays pure; `Region` stays generic; from one `useRegion(name)` hook the consumer both fills the pane and reads its controls.

> Regions are addressed by **flat name**, not tree position — so nesting (an agent splitter wrapping a main-menu splitter) is a non-issue. Two regions are just two names.

**Stacked on the splitter PR.** The Region primitive doesn't depend on Splitter — only the `ShellOwnedSidePanel` demo and a doc link use it — but those need Splitter present to build, so this is based on the splitter component branch (`FEC-977-nimbus-create-splitter-component-for-resizable-two-pane-layouts`) rather than `main`. When the splitter PR merges, GitHub auto-retargets this to `main` and the diff stays region-only. **No Splitter API changes.**

---

## The consumer surface — three things

1. **`<Region name="…" />`** — the **target**: marks where a region renders. It's layout-transparent (`display: contents`), so projected content lays out as a _direct child of the target's parent_ — no wrapper box. (A portal needs a real container node, so an element must exist; `display: contents` keeps it out of layout.) It's a `chakra.div`, so it takes style props. Optionally publishes a `value`.
2. **`useRegion(name)`** → `{ Region, value }` — a stable portal to **fill** the target, plus the published value.
3. **`Region.Provider`** — the scope, mounted **ambiently inside `NimbusProvider`**. So in a Nimbus app you place targets and call `useRegion` **without wrapping anything**.

---

## The mechanism in 30 seconds

**1. The app shell composes Splitter + Region** — the target sits in the aside, publishes the collapse controls, and the Splitter is driven by its own controlled props:

```tsx
const ShellSidePanel = ({ children }) => {
  const [open, setOpen] = useState(false);

  // ① stable callbacks; value memoized on the reactive bit
  const commands = useRef({
    expand: () => setOpen(true),
    collapse: () => setOpen(false),
    toggle: () => setOpen((o) => !o),
  }).current;
  const controller = useMemo(
    () => ({ isCollapsed: !open, ...commands }),
    [open, commands],
  );

  // No <Region.Provider> — NimbusProvider supplies it ambiently.
  return (
    <Splitter.Root
      collapsible
      collapsed={!open}
      onCollapsedChange={(c) => setOpen(!c)}
    >
      <Splitter.Main>{children}</Splitter.Main> {/* ② stable children */}
      <Splitter.Handle />
      <Splitter.Aside>
        <Region name="app-side-panel" value={controller} />
      </Splitter.Aside>
    </Splitter.Root>
  );
};
```

**2. A consumer that owns no splitter markup** fills the pane and drives it:

```tsx
const SidePanel = ({ panel }) => {
  const { Region: Aside, value } = useRegion("app-side-panel");
  const expand = value?.expand;
  const collapse = value?.collapse;

  // ③ depend on the STABLE callbacks, not the value object
  useEffect(() => {
    expand?.();
    return () => collapse?.();
  }, [expand, collapse]);

  return <Aside>{panel}</Aside>; // paints in the aside, stays in this tree
};
```

The `Region` portal has stable identity, so the projected subtree never remounts. Targets resolve by name regardless of nesting, and everything is `null`-safe before the target mounts.

**Standalone, no Splitter** — the same primitive for any "render over there" need:

```tsx
<SomethingDeeplyNested />   {/* authors content */}
<Region name="sidebar" />   {/* content paints here */}

const { Region: Sidebar } = useRegion("sidebar");
return <Sidebar>{content}</Sidebar>;
```

---

## Why the target renders `display: contents`

A portal needs a real DOM node as its container — there's no API to portal "into a position." So the target must render _an_ element. A normal wrapper box would insert a layout node between the parent and the projected content (in a flex/grid parent the projected root becomes a grandchild, not a direct item). `display: contents` makes the element generate **no box** while still existing as the portal container — so projected content lays out exactly where the target sits. Override `display` to make it a real box when you want one. (Documented in the spec + dev docs, since it's a non-obvious technique.)

---

## Performance: why a panel toggle doesn't re-render the app

The provider wraps the entire app (it's inside `NimbusProvider`), so filling a region must touch as little as possible. Four things make that true:

1. **The registry is an external store** (`useSyncExternalStore`), not React state — so `setNode` / `setValue` notify only that name's consumers; **`Region.Provider` never re-renders**. This is the load-bearing decision.
2. **The app flows through as a stable `children` prop** (② above), so the shell's collapse-state re-renders bail out of the app subtree.
3. **Stable callbacks at the registration site** (① above), with the value memoized on the reactive field — so registering doesn't churn.
4. **Consumers depend on the stable callbacks, not the value object** (③ above), so an expand-on-mount effect runs once instead of ping-ponging on every toggle.

Net: toggling re-renders the shell, `Splitter.Root`, its panes, the target, and the one consumer — **never the app body**. Dragging re-renders only the panes.

---

## What's in the box

**`Region` primitive** (`components/region/`) — headless, component-agnostic:

- `<Region name="…" value={…} />` — the target: a layout-transparent `chakra.div` (`display: contents`) that registers its DOM node and optionally publishes a value.
- `useRegion(name)` → `{ Region, value }`; stable portal, `null`-safe before mount.
- `Region.Provider` — DOM-less scope; reuses an ancestor's registry when nested.
- External-store backed; `createPortal` is encapsulated (no consumer-side portal code).

**`NimbusProvider`:** mounts an ambient `Region.Provider` (alongside the i18n provider / toast outlet), so `useRegion` / `<Region>` work app-wide. Additive — existing apps are unaffected.

**Splitter:** unchanged. It composes with `Region` via its existing controlled `collapsed` / `onCollapsedChange`.

---

## Specs, docs & guideline compliance

- **OpenSpec proposal** at `openspec/changes/add-region-primitive` (proposal, design, `nimbus-region` spec, tasks); passes `openspec validate --strict`. New requirements capture the **layout-transparent target** (`display: contents`) and the **ambient scope in `NimbusProvider`**.
- **Docs:** `region.mdx` (overview, _Components › Utilities_) and `region.dev.mdx` (engineering guide — model, the `display: contents` technique, value publishing, the Splitter composition, the performance/memoization contract, SSR).
- **Naming convention:** the entry point is **`Region.Provider`**, not `.Root`, because it mounts no DOM element of its own. Now documented in `docs/file-type-guidelines/compound-components.md` (`.Root` = DOM-mounting root; `.Provider` = headless context scope), with `Region` as the reference example.
- Recipe / slots / design-tokens / i18n are intentionally **N/A** for a headless primitive (documented).

---

## Testing

- **Unit (10 region specs):** node + value records, stable snapshots, per-name notification isolation, reference-equal no-ops, cleanup when emptied, value-survives-node-clear, unsubscribe, flat namespace.
- **Unit (NimbusProvider):** new case asserting `useRegion` / `<Region>` resolve under a bare `NimbusProvider` (ambient scope).
- **Stories (3, with `step()` play functions):** `ProjectIntoNamedRegion`, `NullBeforeTargetMounts`, and `ShellOwnedSidePanel` (Splitter + Region composition — fills the aside, opens on mount, collapses via the published `toggle()`; passes the a11y check). They rely on the ambient provider supplied by Storybook's `NimbusProvider` decorator.
- Splitter stories unchanged and passing. Typecheck + lint clean.

## Notes

- Changeset included (`minor`).


packages/nimbus/src/components/region/region.target.tsx
```diff
@@ -0,0 +1,65 @@
...
+    (node: HTMLDivElement | null) => {
+      const r = consumerRef.current;
+      if (typeof r === "function") r(node);
+      else if (r != null)
+        (r as { current: HTMLDivElement | null }).current = node;
+      if (node) registry?.setNode(name, node);
+      else registry?.clearNode(name, ownNodeRef.current);
+      ownNodeRef.current = node;
+    },
+    [registry, name]
```
tylermorrisford, line 44: [major] Name change leaves a zombie node in the old registry slot.

When  changes (e.g.  → ), React recreates  with the new name *before* calling  to detach the old node. So the detach executes  — a no-op because  has no record yet — and slot  keeps its live DOM node permanently. Any  consumer retains a valid portal container and keeps rendering content into an unexpected location.

Fix: capture  in a stable ref (like / already do) so the detach always reads the name that was current *at attach time*:

```ts
const nameRef = useRef(name);
nameRef.current = name;

const setRef = useCallback(
  (node: HTMLDivElement | null) => {
    // ... consumer ref forwarding ...
    if (node) registry?.setNode(nameRef.current, node);
    else registry?.clearNode(nameRef.current, ownNodeRef.current);
    ownNodeRef.current = node;
  },
  [registry] // name removed from deps; nameRef.current is always up to date
);
```

You'll also want a  to actively evict the old slot when  changes, before the new name is registered.

packages/nimbus/src/components/region/region.target.tsx
```diff
@@ -0,0 +1,65 @@
...
+  // Declare this target as the region's single occupant for the mount's lifetime;
+  // a second concurrent target for the same name warns in development.
+  useEffect(() => registry?.claim(name, "target"), [registry, name]);
+
+  // Publish the value (when provided). Memoize `value` at the call site — its
+  // identity changing is exactly what notifies consumers.
+  useEffect(() => {
+    if (value === undefined) return;
+    registry?.setValue(name, value);
+    return () => registry?.setValue(name, null);
```
tylermorrisford: [major] `setValue` cleanup has no owner-check; stale target unmount can null a live target's published value.

The cleanup on this line calls `setValue(name, null)` unconditionally. Compare with `clearNode`, which is owner-checked: only the holder of the slot may evict it. When two targets for the same name coexist (a logged dev error, but not a hard block), whichever unmounts first wipes the value the still-live target published. This can also surface in the name-change scenario: the cleanup closes over the old `name` (correct), but if a competing target already claimed that slot its value gets clobbered.

Fix: mirror the `clearNode` pattern and guard the cleanup:

```ts
return () => {
  // Only clear our value if we still hold the slot.
  if (registry?.get(name)?.node === ownNodeRef.current || ownNodeRef.current === null) {
    registry?.setValue(name, null);
  }
};
```

packages/nimbus/src/components/region/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from "./region";
+export * from "./region.types";
+export { useRegion } from "./use-region";
+export { createRegionRegistry, RegionRegistryContext } from "./region.registry";
```
tylermorrisford: [minor] Internal registry APIs leaking from the public barrel.

`createRegionRegistry` and `RegionRegistryContext` are implementation details. Exporting them lets consumers create detached registries or inject custom contexts, bypassing `Region.Provider` / `NimbusProvider` entirely. `RegionRegistryContext` is the more dangerous one — a consumer can call `useContext(RegionRegistryContext)` and get direct access to the raw registry with no API guardrails.

If these are intentionally exposed for advanced composition, annotate them `@internal` and document them as unsupported. Otherwise, remove them from this export.

packages/nimbus/src/components/region/region.target.tsx
tylermorrisford, line 44: [major] Name change leaves a zombie node in the old registry slot.

When `name` changes (e.g. `"A"` → `"B"`), React recreates `setRef` with the new name *before* calling `setRef(null)` to detach the old node. So the detach executes `clearNode("B", oldNode)` — a no-op because `"B"` has no record yet — and slot `"A"` keeps its live DOM node permanently. Any `useRegion("A")` consumer retains a valid portal container and keeps rendering content into an unexpected location.

Fix: capture `name` in a stable ref (like `consumerRef`/`ownNodeRef` already do) so the detach always reads the name that was current *at attach time*:

```ts
const nameRef = useRef(name);
nameRef.current = name;

const setRef = useCallback(
  (node: HTMLDivElement | null) => {
    // ... consumer ref forwarding ...
    if (node) registry?.setNode(nameRef.current, node);
    else registry?.clearNode(nameRef.current, ownNodeRef.current);
    ownNodeRef.current = node;
  },
  [registry] // name removed from deps; nameRef.current is always up to date
);
```

You'll also want a `useEffect` to actively evict the old slot when `name` changes, before the new name is registered.

packages/nimbus/src/components/region/region.target.tsx
tylermorrisford: [major] `setValue` cleanup has no owner-check; stale target unmount can null a live target's published value.

The cleanup on this line calls `setValue(name, null)` unconditionally. Compare with `clearNode`, which is owner-checked: only the holder of the slot may evict it. When two targets for the same name coexist (a logged dev error, but not a hard block), whichever unmounts first wipes the value the still-live target published.

Fix: mirror the `clearNode` pattern and guard the cleanup:

```ts
return () => {
  // Only clear our value if we still hold the slot.
  if (registry?.get(name)?.node === ownNodeRef.current || ownNodeRef.current === null) {
    registry?.setValue(name, null);
  }
};
```

packages/nimbus/src/components/region/index.ts
tylermorrisford: [minor] Internal registry APIs leaking from the public barrel.

`createRegionRegistry` and `RegionRegistryContext` are implementation details. Exporting them lets consumers create detached registries or inject custom contexts, bypassing `Region.Provider` / `NimbusProvider` entirely. `RegionRegistryContext` is the more dangerous one — a consumer can call `useContext(RegionRegistryContext)` and get direct access to the raw registry with no API guardrails.

If intentionally exposed for advanced composition, annotate with `@internal` and document as unsupported. Otherwise, remove from this export.

packages/nimbus/src/components/region/region.target.tsx
tylermorrisford, line 44: [major] Name change leaves a zombie node in the old registry slot.

When `name` changes (e.g. `"A"` → `"B"`), React recreates `setRef` with the new name *before* calling `setRef(null)` to detach the old node. So the detach executes `clearNode("B", oldNode)` — a no-op because `"B"` has no record yet — and slot `"A"` keeps its live DOM node permanently. Any `useRegion("A")` consumer retains a valid portal container and keeps rendering content into an unexpected location.

Fix: capture `name` in a stable ref (like `consumerRef`/`ownNodeRef` already do) so the detach always reads the name that was current *at attach time*:

```ts
const nameRef = useRef(name);
nameRef.current = name;

const setRef = useCallback(
  (node: HTMLDivElement | null) => {
    // ... consumer ref forwarding ...
    if (node) registry?.setNode(nameRef.current, node);
    else registry?.clearNode(nameRef.current, ownNodeRef.current);
    ownNodeRef.current = node;
  },
  [registry] // name removed from deps; nameRef.current is always up to date
);
```

You'll also want a `useEffect` to actively evict the old slot when `name` changes, before the new name is registered.

packages/nimbus/src/components/region/region.target.tsx
tylermorrisford: [major] `setValue` cleanup has no owner-check; stale target unmount can null a live target's published value.

The cleanup on this line calls `setValue(name, null)` unconditionally. Compare with `clearNode`, which is owner-checked: only the holder of the slot may evict it. When two targets for the same name coexist (a logged dev error, but not a hard block), whichever unmounts first wipes the value the still-live target published.

Fix: mirror the `clearNode` pattern and guard the cleanup:

```ts
return () => {
  // Only clear our value if we still hold the slot.
  if (registry?.get(name)?.node === ownNodeRef.current || ownNodeRef.current === null) {
    registry?.setValue(name, null);
  }
};
```

packages/nimbus/src/components/region/index.ts
tylermorrisford: [minor] Internal registry APIs leaking from the public barrel.

`createRegionRegistry` and `RegionRegistryContext` are implementation details. Exporting them lets consumers create detached registries or inject custom contexts, bypassing `Region.Provider` / `NimbusProvider` entirely. `RegionRegistryContext` is the more dangerous one — a consumer can call `useContext(RegionRegistryContext)` and get direct access to the raw registry with no API guardrails.

If intentionally exposed for advanced composition, annotate with `@internal` and document as unsupported. Otherwise, remove from this export.

packages/nimbus/src/components/region/region.target.tsx
ByronDWall: @tylermorrisford I traced through this scenario carefully and I believe the current code is actually correct — no zombie node occurs. Here's why:

`setRef` is created via `useCallback(..., [registry, name])`. When `name` changes from `"A"` → `"B"`, `useCallback` returns a **new** function (`setRef_B`) whose closure captures `name = "B"`. But the **previous** function (`setRef_A`, with `name = "A"` in its closure) still exists as the ref React committed on the last render.

React's ref callback contract guarantees: *"If you pass a different ref callback, React will call the **previous** callback with `null` and the next callback with the DOM node."* So the commit-phase sequence is:

1. `setRef_A(null)` — closure has `name = "A"` → calls `clearNode("A", ownNodeRef.current)` → **correctly clears slot A** ✓
2. `setRef_B(domNode)` — closure has `name = "B"` → calls `setNode("B", domNode)` → **registers slot B** ✓

The key distinction is that function *creation* (render phase) and function *invocation* (commit phase) are separate. React invokes the old function reference — which retains its old closure — not the new one.

The proposed fix (removing `name` from deps and using `nameRef`) would actually **introduce** the zombie: with a stable `setRef` identity, React wouldn't detect a ref change, so neither `setRef(null)` nor `setRef(node)` would fire on a name change. The additional `useEffect` mentioned as a companion would then be needed to compensate — equivalent work but more complex code.

Happy to walk through a more detailed trace if helpful.

packages/nimbus/src/components/region/region.target.tsx
ByronDWall: Agreed — the asymmetry with the owner-checked `clearNode` is real. However, the proposed fix (`registry?.get(name)?.node === ownNodeRef.current`) would break the name-change case: by the time the effect cleanup runs, the ref callback has already cleared the old name's node and updated `ownNodeRef.current` to point at the new slot, so the guard would fail and the old name's value would never be cleaned up.

Instead, I added a value-based owner-check that mirrors the `clearNode` pattern more directly:

**Registry:** new `clearValue(name, expectedValue)` — clears only if the current value is the one we published.

**Target cleanup:** `return () => registry?.clearValue(name, value)` — the effect closure captures the `value` it published, so:
- **Normal unmount:** `clearValue("x", ourValue)` — matches, clears ✓
- **Competing target:** `clearValue("x", ourValue)` — the live target's value is different, no-op ✓  
- **Name change:** `clearValue("A", ourValue)` — matches the old slot, clears ✓ (ref callbacks already handled the node transfer)

Fix + 3 tests: https://github.com/commercetools/nimbus/commit/7078255f6

packages/nimbus/src/components/region/index.ts
ByronDWall: Agreed — these are implementation details that don't belong in the public surface. Removed both `createRegionRegistry` and `RegionRegistryContext` from the barrel. All internal consumers (`region.provider.tsx`, `region.portal.tsx`, `region.target.tsx`, `use-region.ts`) import directly from `./region.registry`, so nothing breaks.

Fix: https://github.com/commercetools/nimbus/commit/7078255f6

packages/nimbus/src/components/region/region.target.tsx
ByronDWall: Duplicate of the first review thread — see reply at https://github.com/commercetools/nimbus/pull/1596#discussion_r3398417471

ByronDWall: Duplicate of the first review thread — see reply at https://github.com/commercetools/nimbus/pull/1596#discussion_r3398417471

packages/nimbus/src/components/region/region.target.tsx
ByronDWall: Duplicate — see reply at https://github.com/commercetools/nimbus/pull/1596#discussion_r3398460029

ByronDWall: Duplicate — see reply at https://github.com/commercetools/nimbus/pull/1596#discussion_r3398460029

packages/nimbus/src/components/region/index.ts
ByronDWall: Duplicate — see reply at https://github.com/commercetools/nimbus/pull/1596#discussion_r3398460876

ByronDWall: Duplicate — see reply at https://github.com/commercetools/nimbus/pull/1596#discussion_r3398460876

packages/nimbus/src/components/region/region.target.tsx
tylermorrisford: ohh - thank you for clarifying!

tylermorrisford (approved):
wunderbar

ByronDWall (approved):
Yet another great abstraction, I can see a lot of use-cases for this and it handles portalling and state management so so well. thx for all the careful consideration of render performance, it really makes a difference!

PR #1581, opened 2026-06-09
FEC-983: Create Floating Action Button (FAB)

tylermorrisford:
## Summary

- Adds `FloatingActionButton` — a circular, elevated icon button for a single
  prominent action (e.g. the MC Agent Panel trigger)
- Implemented as a thin wrapper around `IconButton` with fixed styling defaults:
  `variant="solid"`, `colorPalette="primary"`, `size="xl"`,
  `borderRadius="full"`, `boxShadow="3"`, `zIndex="fab"`
- All button behavior — press/keyboard/focus handling, disabled state, ref
  forwarding, and accessible-label enforcement — is inherited from `IconButton`
  with no duplication
- `FloatingActionButtonProps` is a type alias for `IconButtonProps`, existing
  only for naming consistency at the consumer's import site
- Lives in `src/patterns/buttons/` rather than `src/components/` because it is
  a composed, use-case-specific component rather than a primitive; this follows
  the established pattern for components with specific requirements (e.g.
  `patterns/dialogs/`, `patterns/fields/`)

## Files

```
packages/nimbus/src/patterns/buttons/floating-action-button/
  floating-action-button.tsx          # wraps IconButton with FAB defaults
  floating-action-button.types.ts     # type alias for IconButtonProps
  floating-action-button.stories.tsx  # smoke test + ColorPalettes story
  floating-action-button.figma.tsx    # Figma code connect
  floating-action-button.mdx          # designer documentation
  floating-action-button.dev.mdx      # developer documentation
  floating-action-button.docs.spec.tsx # 7 consumer integration tests
  index.ts
packages/nimbus/src/patterns/buttons/index.ts   (new category barrel)
packages/nimbus/src/patterns/index.ts           (adds ./buttons export)
packages/nimbus/src/components/index.ts         (removed old export)
```

## Design decisions

**`IconButton` wrapper, no recipe** — The FAB has no sub-parts and no
dimensions of its own. Using `defineRecipe` would duplicate `IconButton`'s
styling logic for no benefit. Wrapping `IconButton` directly means the FAB
automatically inherits future `IconButton` improvements (animation, dark mode,
focus ring updates) at zero maintenance cost.

**`patterns/buttons/` location** — `src/components/` is reserved for
primitives. The FAB is a composed, context-specific component (the canonical use
case is a single fixed-position trigger per page), so it belongs in `patterns/`.
The `buttons/` sub-directory mirrors the `dialogs/` and `fields/` pattern of
grouping related composed components.

[...]

PR #1614, opened 2026-06-15
refactor(fab): wrap IconButton instead of duplicating button mechanics

misama-ct:
## What

Reworks the `FloatingActionButton` (from #1581) so it **wraps `IconButton`** instead of implementing the button interaction model on its own.

This PR targets the FEC-983 branch, so the diff here is *only* the rework on top of the existing FAB work.

```tsx
// before: a standalone component with its own useButton/refs/slots/recipe
// after:
const FloatingActionButton = (props) => (
  <IconButton
    variant="solid"
    colorPalette="primary"
    size="xl"
    borderRadius="full"
    boxShadow="3"
    zIndex="fab"
    {...props}
  />
);
```

## Why wrap `IconButton`?

A FAB is conceptually **a circular, elevated icon button** — every visual difference (circular, shadow, size, stacking) is expressible through existing button props, and *placement is the consumer's job* via standard style props (`position`, `insetBlockEnd`, …). It adds no interaction behavior of its own.

As a standalone component it carries its own copy of the button stack:

- `useButton` + `useObjectRef` + `mergeRefs` + `data-pressed` wiring, mirroring `button.tsx`
- a slot with the React-Aria event-prop `shouldForwardProp` filter
- types re-declaring the deprecated-native-props block (`onClick`/`disabled`/`tabIndex`/`onMouseEnter`/`onMouseLeave`/`aria-disabled`)
- a recipe with hardcoded width/height

`IconButton` already establishes this pattern in the codebase: a thin wrapper over `Button`. Wrapping `IconButton` lets the FAB inherit all button behavior, accessibility, and prop handling from a single source of truth, so there's only one copy of the interaction/accessibility model to maintain.

### Concretely

- **Deletes the FAB's `recipe` and `slots`**, and removes the recipe registration from the theme.
- **`FloatingActionButtonProps` now mirrors `IconButtonProps`** — no hand-maintained prop duplication. `aria-label` is inherited from `IconButton` (optional at the type level, enforced at runtime), which keeps it consistent with IconButton's slot/context usage.
- **Sizing reuses the Button `xl` size variant** (`h`/`minW` 1200), which already existed in the recipe (commented out), rather than hardcoding width/height on the FAB. The variant sizes both the button and the icon inside it. Its icon value is set to `700` (the commented value was `500`, smaller than `md`'s `600`), matching `md`'s ~60% icon-to-button ratio. The reactivated size also gets Button story coverage.
- **Adds a named `fab` z-index token** (`1150`, between `sticky` and `banner`) so the default stacking is token-driven instead of a magic value. It sits above sticky page chrome but below banners, overlays, modals, and the menus/tooltips a FAB spawns — so it never covers its own popover or a modal.
- **Condenses the stories to what the wrapper adds** (see below).
- **Net code change to the FAB: ~−340 lines.**

## ⚠️ Cross-cutting change to call out

Sizing the FAB via the `xl` variant means **`xl` is now a public Button size** (`<Button size="xl">`), not just an internal FAB detail — it was previously commented out in `button.recipe.ts`. This is intentional (it avoids hardcoding dimensions on the FAB and keeps sizing token-driven through the recipe), but it does widen the Button API, so it's worth a conscious sign-off. The same change sets `xl`'s icon size to `700` and adds `xl` to the Button story's size list for coverage.

> Note: `lg`/`xl`/`2xl` were previously deactivated in `button.recipe.ts`. The FAB needs `xl`, so it's reactivated here; `lg` and `2xl` stay deactivated until a use case calls for them.

## Stories: test what the wrapper adds, not the base

Because the FAB introduces **no behavior of its own**, story coverage for press / keyboard / focus / disabled / ref-forwarding / DOM-prop-filtering would only duplicate the `Button` and `IconButton` suites, which already cover all of it.

The stories are condensed to what the FAB itself contributes:

- **`Base`** — renders with a real `aria-label`, plus a one-line smoke test that the wrapper delegates to an actual `<button>`.
- **`ColorPalettes`** — visual showcase of `colorPalette`, the FAB's one meaningful styling axis.

A header comment notes why the file is intentionally thin, so the inherited-behavior tests aren't re-added later. Result: 6 stories / 175 lines → 2 stories / ~60 lines.

> **Note:** the consumer-facing `floating-action-button.docs.spec.tsx` is **deliberately left self-contained** — it overlaps with the Button docs, but a consumer reading the FAB docs shouldn't be bounced elsewhere to find a working example.

## Behavior preserved

Inherited behavior is still fully exercised by the Button and IconButton story suites; the FAB's own stories pass. Button stories also pass with the reactivated `xl` size.

```
FAB stories     → 2 passed (2)
Button stories  → 15 passed (15)
```

## Test plan

- [x] `pnpm test:storybook:dev` for FAB + Button stories — all pass
- [x] `pnpm --filter @commercetools/nimbus typecheck` — no FAB/Button/`xl` errors (only pre-existing `rating` errors remain)
- [x] `pnpm build:tokens` regenerates the `fab` token; `build-theme-typings` picks up the `xl` size


packages/nimbus/src/components/floating-action-button/floating-action-button.types.ts
```diff
@@ -1,104 +1,13 @@
...
-export type FloatingActionButtonProps = Omit<
-  FloatingActionButtonRootSlotProps,
-  keyof RaButtonProps | keyof NativePropsWithAriaEquivalents
-> &
-  RaButtonProps &
-  NativePropsWithAriaEquivalents & {
-    /**
-     * Accessible label for the button. Required since there is no visible text.
-     */
-    "aria-label": string;
```
tylermorrisford, line 87: This is one aspect I think we should enforce, that the label should _not_ be optional

packages/nimbus/src/components/floating-action-button/floating-action-button.tsx
tylermorrisford, line 1: Does it make sense to place everything in the /patterns directory?

tylermorrisford (approved):
I approve with two questions ❤️

packages/nimbus/src/components/floating-action-button/floating-action-button.types.ts
misama-ct: I thought so too, but then my terminal said no. Reason being that the Button can receive the aria-label through react-aria's slot-mechanism. So technically if we enforce it via typescript we'd have cases where the button actually has a valid label, but typescript would complain. So no typescript enforcement. But the IconButton has a runtime-check that `console.warn`'s the developer if it did not receive a label, so we're good.

packages/nimbus/src/components/floating-action-button/floating-action-button.tsx
misama-ct: That would indeed make sense. I forgot about that one.

PR #1632, opened 2026-06-17
feat(activity-indicator): add ActivityIndicator component (FEC-981)

misama-ct:
## Summary

Adds a new **`ActivityIndicator`** component — an animated three-dot indicator for agent/system activity ("Thinking…", "Processing request", "Agent is typing") in chat and AI surfaces.

This implements [FEC-981](https://commercetools.atlassian.net/browse/FEC-981), but **deviates from the original ticket scope**: rather than adding a `variant="dots"` to `LoadingSpinner`, it ships a dedicated sibling component.

image

### Why a separate component

The decision rests on the **accessibility / semantics contract**:

- **Semantics / a11y (primary reason).** `LoadingSpinner` is _always_ a `role="progressbar"` (via React Aria's `useProgressBar`) with a default label — indeterminate progress, "wait for completion", always announced. `ActivityIndicator` is decorative by default (`aria-hidden`) and upgrades to `role="status"` / `aria-live="polite"` only when given an `aria-label`. Folding these into one `variant` would make a styling prop silently flip the ARIA role and default announcement behavior — a confusing contract.
- **Discoverability.** Given the `nimbus-chat` direction, a chat builder is unlikely to look under `LoadingSpinner` for a typing/thinking hint.
- **Reversibility.** A separate component can be merged/aliased later; a public `variant` can't be retracted without a breaking change.

`ActivityIndicator` is drawn as a single square `viewBox="0 0 24 24"` SVG and reuses `LoadingSpinner`'s exact `size` tokens (2xs=350 … lg=1000), so the two are sizing-identical and interchangeable in icon slots.

`LoadingSpinner` is **unchanged** — non-breaking.

## What's included

- `ActivityIndicator` component: types, recipe, slots, i18n, stories (play functions), docs (`.mdx`/`.dev.mdx`/`.guidelines.mdx`), consumer docs spec, and Figma Code Connect (node `10603-16574`).
- Recipe registered in the theme; exported from the package barrel.
- Changeset (`minor`).
- OpenSpec change `add-activity-indicator` (proposal/design/specs/tasks).

## Key behaviors

- **`size="inherit"` (default):** em-relative; dots scale with surrounding `font-size` and flow inline next to text.
- **Fixed sizes (`2xs`–`lg`):** square icon-box footprint using the same scale points as `LoadingSpinner`, so it drops into input start/end icon slots interchangeably with a spinner.
- **Accessibility:** decorative by default (`aria-hidden`); provide an `aria-label` to upgrade it to a polite live region. No React Aria; not a progressbar.
- **`colorPalette`:** accepts any Nimbus palette (default `primary`); dots fill from `colorPalette.11`.
- **`variant`:** `plain` (default → `colorPalette.11`) for neutral backgrounds, `contrast` (→ `colorPalette.contrast`) for sitting on a solid colored surface — the dots take the palette's contrast color (black/white) automatically.
- **Reduced motion:** bounce → gentle opacity pulse under `prefers-reduced-motion`.

## Dot geometry

The three dots are drawn on the 24×24 icon grid at `r=3` (⌀6), `cy=13`, with `cx` at **4 / 12 / 20** (~1px outer padding, 2px gaps). This is a deliberate hybrid of two competing constraints rather than a clean adoption of either:

- **Honor the icon family.** Every other Nimbus icon reserves outer "safe space" inside the 24px canvas, so the indicator keeps ~1px of outer padding and doesn't read as an outlier sitting next to other icons.
- **Honor the Figma gap.** Figma specs a wider gap between the dots than our strict icon safe-space would allow.

Neither extreme works: the literal Figma three-dots-plus-gaps width is **26px** and overflows the 24px canvas, while strict icon padding squeezes the dots together in the smaller size variants. `cx 4/12/20` splits the difference. The dots rest at `cy=13` (1px below center) because the bounce only travels upward — nudging the resting row down balances the motion around the grid's vertical midpoint.

## Verification

| Check                                       | Result      |
| ------------------------------------------- | ----------- |
| Story play functions (source)               | ✅ 7/7      |
| Consumer docs spec (jsdom)                  | ✅ 3/3      |
| ESLint (workspace source)                   | ✅ 0 errors |
| Typecheck — ActivityIndicator               | ✅ clean    |
| `pnpm --filter @commercetools/nimbus build` | ✅          |

## Notes

- Locale strings for de/es/fr/pt currently fall back to the English source ("Agent is typing"); real translations arrive via Transifex sync, as with every component.

Refs FEC-981

[FEC-981]: https://commercetools.atlassian.net/browse/FEC-981?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


misama-ct:
@claude review this branch

misama-ct:
@claude — thanks for the review. We dug into the a11y feedback (suggestions #1/#2) and ended up reworking the contract rather than patching it, plus cleared the OpenSpec drift (#4). Pushed in `53a47ee52`. Here's the rundown.

### The live-region path was unreliable, so we removed it

Your #1 (a static `aria-label` on a region with no text content) is real, but there's a more fundamental problem underneath it: **mount timing**. This indicator is mounted only while activity is in progress (`{isThinking && …}`), and a `role="status"`/`aria-live` region that mounts _together with_ its content is not announced reliably across NVDA/JAWS/VoiceOver — the region has to already exist in the DOM before its content changes (W3C ARIA22, MDN, TetraLogical). Moving the label into visually-hidden text content (the suggested fix) solves the empty-content issue but **not** the mount-timing one.

So the labeled path was effectively non-functional as built. Rather than keep a contract that doesn't announce, we made the indicator **purely decorative**:

- It is **always `aria-hidden`**; the `aria-label` prop and the `role`/`aria-live` switch are gone. (This also makes #2 moot — no more redundant `aria-live` alongside `role="status"`.)
- Announcing the activity is the **consumer's** responsibility — via adjacent visible text ("Thinking…") or a persistent live region owned by the chat/turn container (e.g. `@react-aria/live-announcer`). That's the React Aria guidance for a reusable, conditionally-mounted leaf, and it matches the component's own "presentational, not a progress indicator" identity.
- For #3: `aria-hidden` is now spread **after** `{...restProps}`, so the decorative contract is authoritative and a consumer can't accidentally un-hide it.

We considered mirroring `LoadingSpinner` (`role="progressbar"` + `aria-valuetext`, where a widget role announces its name on mount and sidesteps the timing problem) but rejected it — it re-introduces the "indeterminate progress, please wait" semantics this component explicitly disclaims.

### Cleanup that followed from the decision

- Removed the now-dead i18n machinery (`.i18n.ts` + generated messages/intl) and the `Nimbus.ActivityIndicator.default` key from the i18n data sets — the component no longer exposes an accessible name of its own.
- Updated stories (dropped the `Labeled` story; `Base` asserts always-decorative), the consumer docs spec, the dev docs, and the guidelines.

### OpenSpec sync (#4)

Updated `proposal.md` / `design.md` / `spec.md` / `tasks.md` — the design doc now records _why_ the live-region approach was abandoned. Also fixed the drift you flagged, after checking each claim against the actual recipe:

- The `primary`/`white` palette **remap** no longer exists — any palette is accepted as-is (`colorPalette.11` / `colorPalette.contrast`).
- The recipe never sets the SVG `width`/`height: 100%`, `overflow: visible`, or `vertical-align: middle`; it relies on `inline-flex` centering, a root-sized box, and the dimensionless SVG filling it. Spec/tasks now match.

### Verification

Story play functions pass against source (6/6) and lint is clean. Your other verified-as-correct points (generated-messages import, recipe registration, keyframes, reduced-motion pulse) are unaffected.

tylermorrisford (approved):
"So much code, for only three dots" - a wise man, probably

valoriecarli (approved):
ngl. i am entranced by the Color Palettes story. i can't stop looking at it.

PR #1638, opened 2026-06-18
fix(splitter): collapse/expand no longer fire the size callbacks

misama-ct:
## Summary

A controlled `Splitter` (you drive `size` and feed `onSizeChangeEnd` back) broke on a programmatic collapse: the aside reopened at the **collapsed width** (e.g. `0`) instead of its previous size. Collapse and expand now change only the aside's layout and are reported solely via `onCollapsedChange` — the size callbacks (`onSizeChange`, `onSizeChangeEnd`) no longer fire on them, so there's nothing for a controlled consumer to feed back and corrupt.

## Root cause

`onSizeChangeEnd` is the seam consumers wire persistence/controlled-`size` to, and the contract says to feed the settled value back into `size`. But collapse committed the size through the normal channel, firing `onSizeChange`/`onSizeChangeEnd` with `collapsedSize`. A consumer following the contract fed that `0` back into `size`; on expand the component restored the (now-`0`) value. It reproduces with **no hook involved** — any controlled-size + controlled-collapse consumer hits it:

```tsx
const [size, setSize] = useState(30);
const [collapsed, setCollapsed] = useState(false);
<Splitter.Root size={size} onSizeChangeEnd={setSize}
  collapsible collapsed={collapsed} onCollapsedChange={setCollapsed} />
// programmatic collapse → onSizeChangeEnd(0) → setSize(0) → expand restores to 0
```

So the defect is in the component, not in `useResponsiveSplitterSizes` (which was simply the first consumer to surface it).

## Fix

The collapse/expand reconcile in `use-splitter-state.ts` now writes the size **silently** — it updates the rendered layout but emits neither `onSizeChange` nor `onSizeChangeEnd` (the same silent-write the controlled-`size` reconcile already uses). Collapse state reaches consumers through `onCollapsedChange`; the size callbacks stay resize-only (drag, keyboard, double-click restore, drag-to-expand). `preCollapseSizeRef` expand-restore and the controlled-`size` loop are unchanged. Fixed at the source, so it holds for every consumer and every wiring — no hook collapse-awareness needed.

## Behavior change / compatibility

This changes the public contract of the size callbacks (hence the `minor` changeset): `onSizeChange` / `onSizeChangeEnd` no longer fire on collapse/expand. A consumer that relied on either firing during collapse should read collapse from `onCollapsedChange` instead. Persistence/controlled-size consumers need no change — the previously-broken case now just works.

## Regression test

`ControlledSizeCollapseRepro` (`splitter.stories.tsx`) — a plain controlled-size + controlled-collapse story with **no hook**: collapse → expand must restore the real width. It fails `expected 30, received 0` against the unfixed component and passes with the fix. `ResizeLockedWhileCollapsed` continues to guard that resize stays locked while collapsed.

## Test plan

- [x] `pnpm test:storybook:dev .../splitter.stories.tsx` — 24/24 stories green (incl. the new repro)
- [x] `pnpm exec vitest --run .../use-splitter-state.spec.tsx .../use-responsive-splitter-sizes.spec.tsx` — 19/19
- [x] `openspec validate add-splitter-component --strict` passes (size-channel scenarios updated + a negative scenario added)
- [x] no splitter typecheck errors

## Risk

Medium — a public behavior change to the size callbacks on collapse, but narrowly scoped to the splitter and covered by stories + unit tests. No migration, infra, or dependency impact.

## Docs / spec

- `splitter.types.ts` + `splitter.dev.mdx`: `onSizeChange`/`onSizeChangeEnd` docs no longer list collapse/expand.
- `add-splitter-component` OpenSpec spec: size-channel scenarios corrected + a negative scenario for collapse/expand.

### DIFF
--- a/.changeset/splitter-collapse-no-size-settle.md
+++ b/.changeset/splitter-collapse-no-size-settle.md
@@ -1,0 +1,13 @@
+---
+"@commercetools/nimbus": patch
+---
+
+**Splitter**: collapsing or expanding the aside no longer fires `onSizeChangeEnd`.
+
+`onSizeChangeEnd` is the resize-settle channel (drag end, keyboard, double-click
+restore) — the seam consumers wire persistence to. Collapse/expand are signalled
+by `onCollapsedChange`, so they no longer also emit `onSizeChangeEnd`. This fixes
+controlled-size consumers that feed the settled value back into `size`: a
+programmatic collapse previously fed the collapsed size back, so the aside
+reopened at the collapsed width (e.g. 0) instead of its real width. `onSizeChange`
+(the live, per-tick channel) still reports collapse/expand.
--- a/.changeset/splitter-collapse-no-size-settle.md
+++ b/.changeset/splitter-collapse-no-size-settle.md
@@ -1,14 +1,16 @@
 ---
-"@commercetools/nimbus": patch
+"@commercetools/nimbus": minor
 ---
 
-**Splitter**: collapsing or expanding the aside no longer fires `onSizeChange` or
-`onSizeChangeEnd`.
+`Splitter`: collapsing or expanding the aside is no longer reported on the size
+callbacks (`onSizeChange` / `onSizeChangeEnd`) — only on `onCollapsedChange`.
 
-Collapse/expand change only the aside's layout (main takes the remainder) and are
-signalled by `onCollapsedChange` — they are not size *changes* a consumer should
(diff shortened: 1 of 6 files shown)


jmcreasman (approved):
Thank you for working on this! 🚀

valoriecarli (approved):
🙌🏻

tylermorrisford (approved):

PR #1662, opened 2026-06-23
chore(deps): update workspace dependencies to latest minor/patch versions

misama-ct:
> [!NOTE]
> Usually I just merge the housekeeping PRs if CI passes, this one needed a couple adjustments though due to some react-aria upgrade (see below), affecting keyboard-navigation the `TagGroup` component

## Summary

Updates `pnpm-workspace.yaml` catalog dependencies to their latest **minor/patch** versions (no major bumps), in risk-ordered groups with build + test checkpoints. Two React Aria behavior changes required small, contained source adaptations (documented below).

## 🔧 Tooling Group (6 updated)

| Package | Old → New |
| --- | --- |
| `globals` | ^17.6.0 → ^17.7.0 |
| `typescript-eslint` | ^8.61.0 → ^8.62.0 |
| `@arethetypeswrong/cli` | ^0.18.3 → ^0.18.4 |
| `@react-types/shared` | ^3.35.0 → ^3.36.0 |
| `rollup` | ^4.62.0 → ^4.62.2 |
| `playwright` | ^1.60.0 → ^1.61.0 |

These are devDependencies → no changeset needed for this group.

## ⚛️ React Aria Group (3 updated, runtime deps)

| Package | Old → New |
| --- | --- |
| `react-aria` | 3.49.0 → 3.50.0 |
| `react-aria-components` | 1.18.0 → 1.19.0 |
| `react-stately` | 3.47.0 → 3.48.0 |

These are **runtime dependencies** of the published `@commercetools/nimbus`, so a **changeset** is included.

`react-aria-components` 1.19.0 introduced two behavior changes that this PR adapts to:

1. **`Menu.onAction` now passes `(key, value)`** ([adobe/react-spectrum#10191](https://github.com/adobe/react-spectrum/pull/10191)). The menu story handler forwards the new `value`; affected story + docs-spec assertions now assert `(key, value)`. `SplitButton` keeps its **single-arg** `onAction` public API (typed explicitly rather than borrowing the Menu type).
[...]

PR #1663, opened 2026-06-23
fix(nimbus-icons): resolve @svgr/babel-plugin-add-jsx-attribute reliably (Vercel build fix)

misama-ct:
## Problem

Vercel deployments fail in the `nimbus-icons` build step:

```
packages/nimbus-icons build: Error: Cannot find module '@svgr/babel-plugin-add-jsx-attribute'
    at jsxPlugin (.../@svgr/plugin-jsx/dist/index.js:77:23) { code: 'MODULE_NOT_FOUND' }
```

Adding `--frozen-lockfile` to the install command (done previously) did **not** help, because the lockfile was never the problem.

## Root cause

`packages/nimbus-icons/.svgrrc.js` references the babel plugin **by bare name**:

```js
plugins: [["@svgr/babel-plugin-add-jsx-attribute", { ... }]]
```

`@svgr/plugin-jsx` passes that config to `@babel/core`, which resolves the bare module name relative to its resolution base. The plugin is only a **transitive** dependency (`@svgr/cli`/`@svgr/plugin-jsx` → `@svgr/babel-preset` → `@svgr/babel-plugin-add-jsx-attribute`) and is declared by **no package in the chain** that references it.

Under pnpm's strict `node_modules` layout this is non-deterministic: it happens to resolve locally (the plugin is hoisted into a reachable spot) but **not** on Vercel's clean `pnpm install --frozen-lockfile`, where it isn't hoisted there. Same lockfile, different hoisting → the bare-name resolution fails.

## Fix

1. **Declare** `@svgr/babel-plugin-add-jsx-attribute` as a direct `devDependency` of `nimbus-icons` (the package that references it — matching the repo's "declare what you reference" convention, alongside the existing `@svgr/plugin-jsx` devDep).
2. **Reference it via `require.resolve(...)`** in `.svgrrc.js`, so Babel receives an **absolute path** and performs no name resolution — independent of pnpm hoisting or install mode.

## Verification

- `@svgr/babel-plugin-add-jsx-attribute` now symlinks into `packages/nimbus-icons/node_modules/@svgr/` and resolves directly from the package.
- `pnpm --filter @commercetools/nimbus-icons run build:icons` succeeds (previously failed).
- **Generated icon output is byte-identical** — `git status` shows zero changes under `src/material-icons/`. This fix only changes resolution, not output.

No changeset: devDependency + build-config only; the published package's runtime deps and built artifacts are unchanged.

### DIFF
--- a/packages/nimbus-icons/.svgrrc.js
+++ b/packages/nimbus-icons/.svgrrc.js
@@ -3,9 +3,16 @@ module.exports = {
   jsx: {
     babelConfig: {
       plugins: [
-        // For an example, this plugin will remove "id" attribute from "svg" tag
+        // Resolve to an absolute path so Babel loads the plugin regardless of
+        // how pnpm hoists it. A bare module name is resolved by @babel/core
+        // relative to its resolution base, which is not reliable under pnpm's
+        // strict node_modules layout (e.g. a clean CI install) and caused
+        // "Cannot find module '@svgr/babel-plugin-add-jsx-attribute'" on Vercel.
+        // The plugin is declared as a direct devDependency of this package so
+        // require.resolve always finds it here.
         [
-          "@svgr/babel-plugin-add-jsx-attribute",
+          // eslint-disable-next-line no-undef
+          require.resolve("@svgr/babel-plugin-add-jsx-attribute"),
           {
             elements: ["svg"],
             attributes: [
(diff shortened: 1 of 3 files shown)


misama-ct:
Superseded — moved this fix onto the housekeeping branch (#1662) per request; no separate branch needed.

PR #987, opened 2026-01-29
Craft 2055  - Remove `react-intl` devDep & implement custom i18n extraction

valoriecarli:
- This PR completes the `react-intl` → `@internationalized/string` migration started in https://github.com/commercetools/nimbus/pull/841. 
- The final `react-intl` dep is removed, all `defineMessages` imports are eliminated & `@formatjs/cli` is replaced with a custom extraction script. 
- Generated i18n files (144 total) are now `gitignored` to follow best practices for generated code. 
- All documentation has been updated to fix incorrect library references & clarify the new compile-time i18n architecture.                                                                                                                                                                

                                                                                                                                                                                     
**TL;DR**: `react-intl` is completely gone, extraction is now custom & cleaner, generated files are `gitignored` & all docs have been updated.

********

**Major Change:**  

Created custom extraction script (`packages/i18n/scripts/extract-messages`) to remove the `react-intl` devDep.   
`.i18n` files are now plain TS objects with no external dependencies. They exist solely for extraction purposes & are never imported by components at runtime.   

We now:
  - Dynamically import .i18n files & read the messages export                                                                                                                 
  - Output Tx JSON format (packages/i18n/data/core.json)                                                                                                                     
  - Keep it simple, maintainable                                                                                                                          
  - No external CLI dependencies   

                            
  Before:                                                                                                                                                                            
 ```
 import { defineMessages } from "react-intl";                                                                                                                                       
                                                                                                                                                                                     
  export const messages = defineMessages({                                                                                                                                           
    dismiss: {                                                                                                                                                                       
      id: "Nimbus.Alert.dismiss",                                                                                                                                                    
      description: "aria-label for the dismiss button",                                                                                                                              
      defaultMessage: "Dismiss",                                                                                                                                                     
    },                                                                                                                                                                               
  });   
```                                                                                                                                                                             
                                                                                                                                                                                     
  After:                                                                                                                                                                             
```
  export const messages = {                                                                                                                                                          
    dismiss: {                                                                                                                                                                       
      id: "Nimbus.Alert.dismiss",                                                                                                                                                    
      description: "aria-label for the dismiss button",                                                                                                                              
      defaultMessage: "Dismiss",                                                                                                                                                     
    },                                                                                                                                                                               
  };                                                                         
```                                                                                                        
                                                                                            
Other goodies in the PR include (but not limited to):                                                                    

- Gitignore generated i18n Files (144 files)                                                                               
- Tracked down all references to i18n (documents, templates, etc) & made updates accordingly. 
[...]

PR #1669, opened 2026-06-24
chore(scripts): make nimbus:reset/init truly clean + add surgical nimbus:clean

misama-ct:
## Why

`pnpm nimbus:init` is meant to bring the repo to a known-clean state, but it didn't anymore. `nimbus:reset` (which `init` runs first) only deleted `node_modules` and `dist`:

```jsonc
"nimbus:reset": "find . -name 'node_modules' -type d -prune -exec rm -rf '{}' + && find . -name 'dist' -type d -prune -exec rm -rf '{}' +"
```

Everything else the repo generates is **gitignored and lives outside those two dirs** — component `intl/` + `*.messages.ts`, docs/mcp data, `material-icons`, and caches (`.turbo`, `.swc`, `*.tsbuildinfo`). Git leaves gitignored files untouched on branch switch, and reset never removed them, so they go stale and leak across branches. The classic symptom: build a component on branch A (generating its i18n), switch to branch B where it was never committed, and the orphaned generated files remain — TypeScript compiles them and you get **phantom errors on every branch** that `nimbus:init` couldn't fix.

## What

- **New `scripts/clean.mjs`** — a tiered, glob-based cleanup. Declarative target patterns instead of a hand-rolled tree walk, using the already-in-catalog [`glob`](https://www.npmjs.com/package/glob) (`dot: true` — Node's built-in `fs.globSync` silently skips dot-dirs) and Node core `util.styleText` for colored output. Supports `--dry-run`.
- **Git-tracked guard** — the script builds the `git ls-files` set and **never deletes a tracked file, or any directory containing one**, and never touches the committed `packages/tokens/src/generated` tree. It reports what it kept.
- **Three tiers** wired in `package.json`:

  | Command | Removes |
  | --- | --- |
  | `pnpm nimbus:clean` | caches + **orphaned** i18n only — keeps `node_modules` and live generated i18n, so the tree still typechecks immediately. Fast everyday fix for cross-branch phantom errors. |
  | `pnpm nimbus:reset` | everything `clean` does **+** all generated artifacts + `dist` + `node_modules` |
  | `pnpm nimbus:init` | unchanged chain (`reset && install && extract-intl && build && playwright:install`), now starting from a genuinely clean tree |

- **Removed 7 accidentally-committed i18n files** (`toast/toast.messages.ts`, `toast/intl/*.ts`, `combobox/intl/de.ts`). These predate the i18n gitignore policy (added in #987, which deleted ~150 such files) and were re-introduced by a Transifex bot sync (#1166) after the policy was in place. They're pure build output: verified they regenerate **byte-identical** from `packages/i18n/data/*.json` via `pnpm extract-intl` (the real translations live in `data/`, nothing is lost).
- **Docs** updated for the three-tier model: `README.md` (branch-switch tip, reset section, troubleshooting), root `CLAUDE.md`, and `packages/color-tokens/CLAUDE.md`. Added a MAINTAINERS note in `clean.mjs` on where to register new caches/artifacts.

## Testing

- `pnpm nimbus:clean` removes orphans + caches; `node_modules`, committed token tree, live component i18n, and local config (`.env.local`, `.vscode/`) all preserved; idempotent on re-run.
- `--all`/`--all --node-modules` dry-runs confirmed correct target set (matches `packages/tokens/generated/*/dist` but never `src/generated`).
- Empirically removed the 7 files, ran `pnpm extract-intl`, and diffed: all regenerate byte-identical with German intact (`Schließen`, `Auswahl löschen`); `data/*.json` untouched.
- `eslint scripts/clean.mjs` clean.

No changeset: internal tooling + docs only, no consumer-facing package change.


valoriecarli (approved):
Oh noes, thanks for catching that and tidying it up. 🙌🏻

tylermorrisford (approved):
i likes it

PR #1671, opened 2026-06-24
feat(button): allow Tooltips on disabled Buttons via allowFocusWhenDisabled

misama-ct:
## Why

Consumers want to attach a `Tooltip` to a **disabled** `Button` — almost always to explain *why* the action is unavailable ("Complete required fields", "You don't have permission"). Today that's impossible: `Button` delegates disabled state to React Aria's `useButton`, which applies the native `disabled` attribute to a `<button>`. A natively-disabled element is removed from the tab order and fires no hover/focus events, so `TooltipTrigger` (which opens on hover **and** focus and needs a focusable trigger) has nothing to bind to — for mouse *or* keyboard users. The control that most needs an explanation is the one that can't carry one.

## What changed

Adds an opt-in boolean prop **`allowFocusWhenDisabled`** to `Button` (default `false`). It only has an effect together with `isDisabled`. When both are set, the button is **soft-disabled**:

- stays **focusable** and in the natural tab order (tooltip opens on focus),
- stays **hoverable** (tooltip opens on hover),
- exposes **`aria-disabled="true"`** instead of the native `disabled` attribute (existing `_disabled` recipe styling still applies — no visual change),
- **cannot be activated** — press, click, `Enter`/`Space`, form submit/reset, and link navigation are all suppressed.

```tsx
<Tooltip.Root>
  <Button isDisabled allowFocusWhenDisabled>Publish</Button>
  <Tooltip.Content>Complete all required fields to publish.</Tooltip.Content>
</Tooltip.Root>
```

Default behavior (prop omitted/`false`) is **completely unchanged**: `isDisabled` continues to apply the native `disabled` attribute. Purely additive, non-breaking.

## How

Confined to one branch in `button.tsx`: when soft-disabled, `useButton` is told the button is *enabled* (so it omits the native `disabled` attribute and stays focusable) with activation handlers withheld; we then re-apply `aria-disabled` and a click guard (`preventDefault`/`stopPropagation`). A native button dispatches a click for mouse, touch, and keyboard activation — and that click is what triggers form submit/reset and link navigation — so a single click guard covers every activation path.

### Why suppress activation, not just visual-only `aria-disabled`?

Visual-only disable already works today via `<Button aria-disabled>`. The reason it isn't enough: `aria-disabled="true"` promises assistive tech the control is *non-operable*. Without suppression the button still fires `onPress`, and a `type="submit"` button still submits its form on Enter **with no handler wired** — a "looks/announces disabled but still acts" bug that's worse than the original limitation, and the exact case consumers ship. Getting it right means guarding press/click/keydown/submit/nav together, which is identical for every consumer, so it belongs in the component. Full rationale + rejected API shapes in the OpenSpec `design.md`.

## Spec

OpenSpec change: `openspec/changes/add-button-allow-focus-when-disabled/` (proposal, design, `nimbus-button` spec delta, tasks).

## Testing

- `pnpm --filter @commercetools/nimbus typecheck` — clean
- `pnpm lint` — clean (button)
- Storybook story tests (source): **18/18** — adds `DisabledButFocusable`, `DisabledButFocusableDoesNotSubmit`, `DisabledWithTooltip`
- Docs-spec consumer tests (source): **14/14** — adds a copyable "Tooltip on a disabled Button" example

## Docs

- `button.dev.mdx` — "Keeping a disabled button focusable (for tooltips)"
- `tooltip.dev.mdx` — "With disabled buttons"

Changeset: minor bump on `@commercetools/nimbus`.

### DIFF
--- a/.changeset/button-allow-focus-when-disabled.md
+++ b/.changeset/button-allow-focus-when-disabled.md
@@ -1,0 +1,26 @@
+---
+"@commercetools/nimbus": minor
+---
+
+`Button`: add `allowFocusWhenDisabled` to support tooltips on disabled buttons.
+
+A disabled button normally can't host a `Tooltip` because the native `disabled`
+attribute removes it from the tab order and stops hover/focus events — exactly
+when a tooltip explaining _why_ the action is unavailable is most useful. Pass
+`allowFocusWhenDisabled` alongside `isDisabled` to keep the button focusable and
+hoverable: it is announced as disabled via `aria-disabled` (not the native
+`disabled` attribute), stays in the tab order, and its action stays fully
+suppressed (press, click, `Enter`/`Space`, form submit/reset, and link
+navigation).
+
+```tsx
+<Tooltip.Root>
+  <Button isDisabled allowFocusWhenDisabled>
+    Publish
+  </Button>
+  <Tooltip.Content>Complete all required fields to publish.</Tooltip.Content>
+</Tooltip.Root>
+```
+
+Default disabled behavior is unchanged — omit the prop and `isDisabled` works
+exactly as before.
--- a/openspec/changes/add-button-allow-focus-when-disabled/design.md
+++ b/openspec/changes/add-button-allow-focus-when-disabled/design.md
@@ -1,0 +1,134 @@
+# Design: `allowFocusWhenDisabled` on Button
+
+## Problem framing
+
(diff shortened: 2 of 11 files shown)


tylermorrisford (approved):
Really nice improvement

valoriecarli (approved):
🙌🏻

PR #1672, opened 2026-06-24
feat(markdown): add Markdown component (FEC-984)

misama-ct:
image

## Summary

Adds a new **`Markdown`** component to `@commercetools/nimbus` that renders a Markdown string into Nimbus-styled, accessible React elements. It targets **agentic / LLM chat output** as well as authored content (release notes, descriptions).

Built on the headless [`react-markdown`](https://github.com/remarkjs/react-markdown) engine, with all styling owned by Nimbus via the default component map (design tokens, no slot recipe). GitHub Flavored Markdown — tables, task lists, strikethrough, autolinks, footnotes — is on by default. Backed by an OpenSpec proposal (`openspec/changes/add-markdown-component`).

## Motivation

Merchant Center surfaces increasingly need to render Markdown — chat/assistant responses and authored content alike — with Nimbus styling and WCAG compliance, without each consumer wiring up a renderer, a styling layer, and a sanitization story by hand.

## What changed

- **Default Nimbus renderers** for every standard element, reusing existing primitives (`Heading`, `Link`, `Code`, `Text`) and styled `chakra.*` elements on the Figma `Markdown/*` typography scale — no new tokens, no slot recipe.
- **Per-element overrides** — `components={{ a: MyLink }}` replaces one element; every other default stays intact.
- **Custom component tags** — register a non-standard key (`components={{ SearchQueryResultCard: Card }}`) and embed that tag in the source (`<SearchQueryResultCard id="foo" />` or a paired form with markdown children). String attributes arrive as props with original casing preserved. Only registered tags are materialized; unregistered tags stay inert.
- **GitHub alerts** — `> [!NOTE]` / `[!TIP]` / `[!IMPORTANT]` / `[!WARNING]` / `[!CAUTION]` blockquotes render as Nimbus callouts (color palette + icon), with the type exposed to assistive tech via a visually-hidden localized label.
- **Streaming** — `isStreaming` completes half-written tokens, memoizes settled blocks so only the final block re-parses per token, keeps paired custom-tag regions coherent across chunks, and manages `aria-busy` + a single completion announcement.
- **Safe by default** — raw HTML is never rendered (`skipHtml`, no `rehype-raw`), rendering is restricted to a safe element allowlist, and dangerous URLs (`javascript:` / `vbscript:` / `file:`) are neutralized. Image-host security is delegated to the app CSP.
- **Accessible (WCAG 2.1 AA)** — `headingOffset` preserves the page outline, `th[scope]` tables, read-only task-list checkboxes, external links with `rel="noopener noreferrer"` + an "opens in new tab" indicator, and dev-mode warnings for heading skips / missing alt text.
- **Curated public API** — no `remarkPlugins` / `rehypePlugins` / `sanitizeSchema` passthrough; the engine stays an internal, swappable implementation detail. Consumers shape output through `components` and `allowedElements` / `disallowedElements`.

```tsx
<Markdown
  components={{ a: MyLink, SearchQueryResultCard: Card }}
  isStreaming={isResponding}
  headingOffset={1}
  maxW="60ch"
>
  {markdownString}
</Markdown>
```

## How it works

The `components` prop is the single renderer surface: standard element keys override that element's renderer; non-standard keys register an embeddable tag. A Nimbus-owned, dependency-free remark plugin (added only when such keys are present) walks the mdast and replaces each registered tag with a synthetic node carrying the original-cased name and parsed attributes, which `react-markdown`'s `components[tagName]` lookup then resolves. This is parsing only — no MDX, no code evaluation, no `rehype-raw` — preserving the safe-by-default posture with zero added runtime dependency for custom tags.

## Dependencies

New: `react-markdown`, `remark-gfm`, `remend`. No `rehype-raw` / `rehype-sanitize`. The Markdown engine is a **code-split chunk loaded only when a consumer imports `Markdown`**, so the main entry is unaffected (`pnpm check:bundle-size` runs in CI against the PR baseline).

## Test plan

- [x] `typecheck` — 0 errors
- [x] `lint` — clean
- [x] Storybook play functions (incl. `AllElements` showcase, GitHub alerts, custom self-closing/paired tags, unregistered-inert, code-fence-literal, streaming-partial, heading-offset clamp, footnotes)
- [x] Unit tests (attribute parser, remark plugins, block splitter incl. boundary regressions, heading-skip detection, empty/malformed input, dev warnings)
- [x] Consumer doc examples
- [x] `openspec validate add-markdown-component --strict` — valid

## Out of scope (v1)

Code-block chrome + syntax highlighting (`CodeBlock` is a documented fast-follow), Math (KaTeX), emoji shortcodes, generic custom directives, media embeds, a markdown editor, and Mermaid diagrams. Plugin-dependent features are deferred deliberately rather than handed to consumers via engine passthrough.

## Feedback wanted

- Is the curated public API (no engine passthrough) the right long-term contract, or should advanced consumers have an escape hatch sooner?
- Does the custom-component-tag ergonomics fit your expected LLM output shapes?

Closes FEC-984


misama-ct:
@claude please review this PR

misama-ct:
@claude Thanks for the review — I had three independent perspectives (ARIA/screen-reader correctness, React + library runtime behavior, and an adversarial skeptic) evaluate all four findings, then verified against the actual test file and `mdast-util-to-hast` source. **All four held up**, so I've addressed them in c5fc5d663:

- **🟡 Re-streaming re-announcement** — confirmed a real bug for the reused-instance / "regenerate" flow. The effect now calls `setAnnouncement("")` when `isStreaming` (re)starts, so the live region is empty while busy and the next settle produces a genuine `"" → completeLabel` change. (Verified the fire-once-per-stream-start behavior — deps stay `[isStreaming, completeLabel]`, so no extra renders.)
- **🟡 `readOnly` on checkboxes** — confirmed dead intent: browsers ignore `readOnly` on `<input type="checkbox">`, and `mdast-util-to-hast@13` already emits `disabled: true` (`lib/handlers/list-item.js:46`). Dropped the `readOnly` clone prop and updated the GFM story to assert `disabled` (the real non-interactivity mechanism) instead.
- **🔵 Redundant `aria-live`** — dropped; `role="status"` already implies `aria-live="polite"` + `aria-atomic="true"`.
- **🔵 Task-list `aria-label` absorbing nested content** — `getNodeText` now skips nested `ul`/`ol` (inline formatting text is still captured), with a new `hast-node.spec.ts` case covering it.

**Verification:** `typecheck` clean; `hast-node.spec.ts` 8/8; `markdown.stories.tsx` 18/18.

One gap worth flagging: no story yet exercises a `false→true→false` cycle *twice* on one instance, so the re-announcement fix is covered by reasoning rather than a regression test. Happy to add that story if reviewers want it gated.

misama-ct:
@claude happy now?

packages/nimbus/src/components/markdown/markdown.tsx
```diff
@@ -0,0 +1,277 @@
...
+      alertLabels,
+    });
+    return { ...nimbusComponents, ...componentsOverride } as Components;
+  }, [headingOffset, opensInNewTabLabel, alertLabels, componentsOverride]);
+
+  // Custom component tags = consumer `components` keys that are not standard
+  // markdown/GFM element names. Their presence (a) registers them with the
+  // custom-tag remark plugin so the matching tags are materialized, and (b) is
+  // unioned into the element allowlist so react-markdown does not strip them.
+  const customTagNames = React.useMemo(() => {
```
tylermorrisford, line 138: Nit from Claude: 

> Any non-allowlisted components key becomes a custom tag (sharp edge, low). customTagNames (markdown.tsx:138) treats every key not in DEFAULT_ALLOWED_ELEMENTS as an embeddable tag.
Overriding a standard-but-not-defaulted element (kbd, mark, sub) — or a typo'd key like { imgg: ... } — silently registers it as a tag and unions it into the allowlist rather than erroring. Harmless in practice (markdown can't emit those without raw HTML, which is skipped), and the open index signature is an acknowledged tradeoff in markdown.types.ts. Consider a dev-only warning when a custom key isn't Capitalized, or just document "custom tags must be PascalCase."

tylermorrisford (approved):
"This is an unusually high-quality PR."

packages/nimbus/src/components/markdown/components/markdown.renderers.tsx
```diff
@@ -0,0 +1,413 @@
...
+  return {
+    h1: makeHeadingRenderer(1, headingOffset),
+    h2: makeHeadingRenderer(2, headingOffset),
+    h3: makeHeadingRenderer(3, headingOffset),
+    h4: makeHeadingRenderer(4, headingOffset),
+    h5: makeHeadingRenderer(5, headingOffset),
+    h6: makeHeadingRenderer(6, headingOffset),
+
+    p: (props) => <Text as="p" {...withoutNode(props)} />,
+
```
valoriecarli, line 110:
```suggestion

em: (props) => <chakra.em fontStyle="italic" {...withoutNode(props)} />,
```

valoriecarli (approved):
Left a comment about `em`, but I know, I am picky. 
Just a little something I noticed with my eyeballs in Storybook & the docs.

Thank you for making Claude happy. 
This is great work, thank you so much!

packages/nimbus/src/components/markdown/components/markdown.renderers.tsx
valoriecarli: Just a little something I noticed with my eyeballs in Storybook & the docs.
Screenshot 2026-07-01 at 10 31 59
becomes
Screenshot 2026-07-01 at 10 32 46

and..
Screenshot 2026-07-01 at 10 34 39
becomes
Screenshot 2026-07-01 at 10 34 52

misama-ct: [Fixed](https://github.com/commercetools/nimbus/pull/1672/changes/fd83e5592882a78719b643ef0f9c4f8aec9fb082), thank you

PR #1683, opened 2026-06-26
chore(typecheck): resolve editor/dev type-checking against source

misama-ct:
## Problem

Stories and specs import components via the package barrel `@commercetools/nimbus`. TypeScript (the editor's language server **and** `tsc`) resolved that bare specifier through `package.json` `exports` to the **last-built** `dist/index.d.ts` — not live `src/`. So after changing a component's public types, the editor flagged stale-`dist` "phantom" errors until a rebuild, even though the dev tests (`test:dev`) already ran against source. This is the same source-vs-built mismatch Byron's `test` / `test:dev` split solved for tests — just never applied to type checking.

## Change

Extends the existing dev/prod split to type checking, **editor-first and minimal**:

- **`tsconfig.json`** — alias the barrel to `./src/index.ts` so the editor (and the new `typecheck:dev`) check live source; no rebuild needed.
- **`tsconfig.prod.json`** *(new)* — extends the base but redefines `paths` to drop the alias, so the barrel resolves back to `dist/index.d.ts`.
- **`package.json`** — `typecheck` now runs `tsc --noEmit -p tsconfig.prod.json` (built surface, unchanged behavior); added **`typecheck:dev`** → `tsc --noEmit -p tsconfig.json` (source).
- **`.storybook/main.ts`** — pin the barrel back to `dist` for the production Storybook build/test. Required because Vite 8's `resolve.tsconfigPaths` reads `tsconfig.json` and would otherwise silently redirect the **production** Storybook test (`test:storybook`) to source.
- **Docs/skills** — steer the development loop to `typecheck:dev` (CLAUDE.md with a decision diagram, `git-conventions.md`, the `writing-*` and `nimbus-code-connect` skills). Pre-commit/CI gates stay on the built `typecheck`.

## Resolution after this change (verified)

| Context | Resolves barrel to |
| --- | --- |
| Editor / `typecheck:dev` | `src/index.ts` |
| `typecheck` / `typecheck:strict` (CLI + CI) | `dist/index.d.ts` |
| Production Storybook test (`test:storybook`) | `dist/index.es.js` |
| Dev Storybook test (`test:storybook:dev`) | `src/index.ts` |
| `build:lib` | builds clean (alias inert — no bundled file imports the barrel) |

Each row checked empirically — including `vite:resolve` traces of the prod Storybook test **before** (leaked to `src`) and **after** (pinned to `dist`) the `main.ts` change, and the dev Storybook test still resolving to `src`.

## Behavior

- **Development** (editor / Claude / TDD): `typecheck:dev` + `test:dev` → source, no build.
- **CI / pre-push parity**: `build` → `typecheck:strict` → `test` → built `dist` (unchanged).

`typecheck:strict` is unchanged and is *not* a third mode — it's the same built-surface check as `typecheck`, just set to fail the run.

## Also: fix `eslint .` under nested worktrees

Same class of config-resolution bug, surfacing through typescript-eslint instead of `tsc`. Nested git worktrees under `.claude/worktrees/` each carry a full copy of the repo, including their own `tsconfig`. Flat-config ESLint does **not** read `.gitignore`, so `eslint .` walked into those trees, typescript-eslint saw multiple candidate `TSConfigRootDir`s under the repo root, and parsing failed on **every file in the workspace** (~10.8k errors) — even though linting an individual file passed clean.

- **`eslint.config.mjs`** — add `**/.claude/worktrees/**` to the global `ignores`.

Durable fix (not just local cleanup): any future worktree created under that path would re-trigger the failure the same way. Verified: full `eslint .` goes from exit 1 (~10.8k parsing errors) to exit 0 (one pre-existing, unrelated `react-hooks/exhaustive-deps` warning).

## Notes

- Internal tooling only; no consumer-facing change, so **no changeset**.
- After pulling, restart the editor's TS server so it re-reads `tsconfig.json`.


valoriecarli (approved):
thanks for doing this 🙌🏻

PR #1684, opened 2026-06-27
fix(nimbus): inputs no longer sit darker than the page in dark mode

misama-ct:
## What

TextInput, SearchInput and Combobox rendered their default (`solid`) variant with a fill that was **darker than the page background in dark mode** (looking recessed) and that **blended into the page in light mode**.

Before (input bg blends in with page bg in light-mode, but not in dark-mode):

image
<br/>
image

After (input bg blends in with page bg, even in dark-mode):

image

## Why

These three were the only inputs filling with `primary.1` (the ctviolet **brand** palette, step 1). Every other input uses `neutral.1` (gray step 1). The two step-1 values diverge in dark mode:

| | Light | Dark |
|---|---|---|
| Page `bg` | `#fff` | `#111111` |
| `neutral.1` | `#fcfcfc` | `#111111` |
| `primary.1` | `#FDFDFF` | **`#000`** (pure black) |

`primary.1` dark (`#000`) is darker than the page (`#111111`) → recessed; light (`#FDFDFF`) ≈ `#fff` → blends.

## Change

Point the resting fill of all three at `neutral.1`, matching NumberInput / MultilineTextInput and the rest of the inputs. Hover stays `primary.2` (unchanged, consistent with siblings). No token definitions change — three one-line edits.

- `text-input.recipe.ts`
- `search-input.recipe.ts`
- `combobox.recipe.ts`

## Verification

Verified locally in the docs dev server in both light and dark mode: the fields now sit consistently against the page and match the other inputs.

## Tests

The `solid`-variant Combobox story asserted the resting background was `rgb(255, 255, 255)` (the old `primary.1`), so it failed once the fill moved to `neutral.1`. Rather than re-pin the assertion to a new color we don't actually mean to guarantee, removed the Combobox stories that assert specific **recipe styling output** rather than behavior:

- removed `SizeSmallVariant` / `SizeMediumVariant` (root height `32px` / `40px`), `VariantSolid` (width + background), `VariantGhost` (transparent background + width)
- trimmed `StateIsInvalid` to keep the `data-invalid` attribute check and drop the `--border-color` color assertion

Behavioral coverage is unchanged — clear-button visibility (`display`/`opacity`/`pointerEvents`) and multi-select tag-wrapping tests are kept.


valoriecarli (approved):
I don't know if it's a me thing, so will not hold this up. It looks as expected in Chrome, however FF is my default, and it always throws me off at first. 

FF:
Image

vs Chrome:
Image

misama-ct:
>  I don't know if it's a me thing, so will not hold this up. It looks as expected in Chrome, however FF is my default, and it always throws me off at first.

Yeah no, something is fishy, can see it as well. On it. And maybe there is already a fix available when you wake up and read this ✨ .

misama-ct:
@valoriecarli Fortunately it was a storybook and not a nimbus problem, also not related to Firefox, it happens in chrome as well under circumstances. I squeezed the [fix](https://github.com/commercetools/nimbus/pull/1684/changes/5cb556f8b57ceac3fe5b1c073f65fe3411a1ddf3) in.

valoriecarli:
@misama-ct because you're awesome like that - thank you!

PR #1685, opened 2026-06-27
fix(splitter): prevent first-paint flash and keep the resize handle above pane content

misama-ct:
## Problems

Two `Splitter` issues:

1. **First-paint 50/50 flash** — a `Splitter` driven by `useResponsiveSplitterSizes` with a pixel/token `size` flashed the uncontrolled **50/50** split for a frame before settling to the configured size on load.
2. **Resize handle hidden behind pane content** — on hover the handle's track was painted over by `position: sticky` content inside a pane (e.g. a sticky header or footer with its own `z-index`), so the handle was only partially visible instead of overlaying the content.

### 1. First-paint flash

A pixel `size` resolves to a percentage only after the container is measured. Under a **busy concurrent mount** (a content-heavy view + async data + routing all committing at once), React **defers that measurement update past the first paint** — so `size` is briefly absent and the splitter paints its 50/50 seed. The splitter then adopted the resolved size via **passive effects** (pane registration + the size/collapse reconcile), keeping the pane at 50/50 for a painted frame.

This is a **production bug, not a dev/StrictMode artifact.** Confirmed in a `vite build` + `preview` (StrictMode is a no-op there) by isolating each candidate fix:

| Variant (production build)       | Visible 50/50 flash?                                           |
| -------------------------------- | -------------------------------------------------------------- |
| Unfixed                          | **Yes** (3/3)                                                  |
| Visibility gate in the hook only | **Yes** (3/3) — the imperative `visibility` never even applied |
| **Layout-effect reconcile only** | **No** (5/5)                                                   |
| Both                             | **No** (4/4)                                                   |

A trivial page flushes the deferred update before paint, which is why the flash does **not** reproduce in a minimal repro, in warm dev, or in jsdom/Storybook — it only surfaces under real concurrent-mount load.

### 2. Handle stacking

The handle is `position: absolute; z-index: 1`, a sibling of the panes. The panes were `position: static`, so they did **not** establish their own stacking context — a pane's `position: sticky; z-index: 1` descendant therefore participated directly in the **root** stacking context and tied the handle's `z-index: 1`. Ties break by DOM order, and the right pane follows the handle in the DOM, so its sticky header/footer won and painted over the handle. Confirmed with `elementsFromPoint` down the handle's centre line: the handle was occluded exactly where a sticky element sat (top + bottom), and topmost everywhere else.

### 1. First-paint flash

Make **pane registration and the size/collapse reconcile run as layout effects** (`useEffect` → `useLayoutEffect`) in `use-splitter-state.ts` + `splitter.pane.tsx`. Layout-effect work is flushed before the first paint, so a size that resolves after mount is adopted before paint instead of a frame later. **Four `useEffect` → `useLayoutEffect` lines; nothing else.**

> An earlier iteration of this PR added an imperative "visibility gate" in the hook. The matrix above shows it was **unnecessary and ineffective** in production (the layout-effect reconcile alone fixes the flash), so it was removed.

### 2. Handle stacking

Give each `pane` its own stacking context via `isolation: isolate` (`splitter.recipe.ts`). The handle (a positive-`z-index` sibling) then always paints above the entire pane subtree, regardless of what `z-index` values consumer content uses — no magic numbers, no layout/containing-block side effects, and `position: sticky` inside a pane is unaffected. **One line.**

## Test

`splitter.reconcile-timing.spec.tsx` — a deterministic regression guard for the flash. The visible flash is a production paint-timing transient that **cannot** be reproduced in jsdom/Storybook (verified: even a StrictMode-wrapped story passes with the bug present), so rather than ship a test that passes either way, this asserts the **property that prevents the flash**: the controlled-size reconcile is applied synchronously with the commit. `flushSync` runs layout effects (and the re-render they schedule) synchronously but defers passive effects, so right after a `flushSync` size change the pane already reflects the new size.

**Negative control:** reverting the reconcile to a passive effect makes this test fail (`Received: "30%"` instead of `"60%"`).

The handle-stacking fix is CSS-only (a stacking-context boundary); it's verified in-browser rather than with a unit test.

## Verification

- `splitter.reconcile-timing.spec.tsx` (new) + `use-splitter-state.spec.tsx` + `use-responsive-splitter-sizes.spec.tsx` — green; negative control confirms the new test catches the regression.
- Full splitter story suite (collapse / controlled / keyboard / drag) — 25 green.
- Production docs build: 50/50 flash present before, gone after (multiple runs).
- Handle stacking: with `position: sticky` content in a pane, `elementsFromPoint` down the handle's centre line returns the handle as the **topmost** element at every point — including over the sticky content — after the fix; occluded where a sticky element sat before.

Adds a `patch` changeset covering both fixes.


misama-ct:
@claude please review

misama-ct:
@claude Thanks for the thorough review — much appreciated.

On the two recommendations, I'll pass on both:

1. **SSR-safe `useLayoutEffect`** — Nimbus targets React 19+ (we don't support older versions), and React 19 no longer emits the *"useLayoutEffect does nothing on the server"* warning. So the console noise this guards against doesn't occur in any React version we support, and importing from `react` directly is fine here.

2. **Test assertion comment** — I'll leave the test as-is; the DOM order is already clear enough from the surrounding context.

Everything else checks out — merging.

valoriecarli (approved):

PR #1686, opened 2026-06-27
feat(tabs): animated active-tab indicator + shared useSlidingIndicator hook

misama-ct:
Components - TabNav - Variants ⋅ Storybook; Jul 1, 2 10 PM

## What & why

`Tabs` and `TabNav` are visual twins. This PR gives them one shared variant set and a single active marker that **slides between items** as the selection changes, instead of a static per-item marker.

Pure CSS can't move a marker *between* sibling items, so the slide is a small reusable component change rather than a recipe-only tweak.

## Changes

- **One shared variant set for both components** — `line` (default), `rounded` (soft rounded-rect highlight), `pill` (capsule). The `rounded`/`pill` highlights are themeable via `colorPalette`. `Tabs` additionally layers `orientation`/`placement` onto `line`.
- **The active marker is a single sliding indicator.** It slides between items on selection change — a bar for `line`, a filled highlight for `rounded`/`pill`. There is no per-instance toggle; the only "off" switch is global `prefers-reduced-motion: reduce`, under which the highlight snaps (the only motion lever Nimbus has).
- **Shared hook `useSlidingIndicator`** (`packages/nimbus/src/hooks/use-sliding-indicator/`) drives both components — DOM-measured with `getBoundingClientRect` in `useLayoutEffect`, re-measuring on active-item changes (`MutationObserver`) and layout changes (`ResizeObserver`). Callers pass `activeSelector`, `watchAttributes`, and a `getGeometry({container, active})` callback, so the same hook drives an edge bar (`line`) or a filled highlight (`rounded`/`pill`). It sets `data-animated` on the container on mount, so the recipe's static marker is the SSR / pre-hydration / no-JS fallback and the measured slide takes over before paint.
- **Snaps into place on first paint.** The initial placement is applied with the transition momentarily suppressed (jump committed via a synchronous reflow, then restored), so the indicator appears over the active item on first render rather than animating in from the container's corner — for every variant and orientation. It slides only on subsequent selection changes.
- **Deprecated aliases (non-breaking)** resolved in-component and typed `@deprecated`: `Tabs` `pills` → `pill`; `TabNav` `tabs` → `line`. Existing consumers (incl. the two docs-app `variant="pills"` usages) keep working with no edits.

For both components the indicator is `aria-hidden` / non-focusable, so `aria-current` / `aria-selected`, focus rings, and keyboard navigation are unaffected.

## Tests & docs

- Play-function stories for both components are **interactive** — clicking any item moves the active marker (and the slide). `Tabs` does this natively via React Aria; `TabNav.Item` is controlled via `isCurrent`, so its showcase stories run through a small stateful helper (a router does this in a real app). The behaviour is exercised in the regular stories — including a slide assertion in each `Base` story — plus deprecated-alias smoke tests.
- `tabs.mdx`/`tabs.dev.mdx` and `tab-nav.mdx`/`tab-nav.dev.mdx` cover the variants, animation-by-default, reduced-motion, and aliases.
- OpenSpec change at `openspec/changes/add-animated-tab-indicator/` (`nimbus-use-sliding-indicator` + `nimbus-tabs`); changeset included.

## Verification

- `pnpm --filter @commercetools/nimbus build-theme-typings` → `typecheck` — clean
- `pnpm test` for both story files — **23 passed** (11 TabNav + 12 Tabs), real headless Chromium, against the built `dist` bundle
- ESLint on changed files — clean
- `openspec validate add-animated-tab-indicator --strict` — valid

## Notes

- Recipe quirk worth knowing: arbitrary CSS selectors are only accepted by the type system inside `variants.*` slot objects (not `base.*` / `compoundVariants[].css`), so the `[data-animated]` rules live in the variant slots.

### DIFF
--- a/packages/nimbus/src/hooks/use-sliding-indicator/use-sliding-indicator.ts
+++ b/packages/nimbus/src/hooks/use-sliding-indicator/use-sliding-indicator.ts
@@ -72,6 +72,12 @@ export interface SlidingIndicatorOptions {
  * momentarily suppressed so the indicator snaps to the active item instead of
  * sliding in from the container corner on initial render.
  *
+ * Note: scroll is **not** tracked. Geometry is computed in container
+ * coordinates, so if the container is itself scrollable the indicator only
+ * re-aligns on the next selection change or resize, not on scroll. Today's
+ * callers (`Tabs`/`TabNav`) don't scroll their item strip, so this is a
+ * non-issue; a future scrollable caller would need to observe scroll itself.
+ *
  * Used to drive the animated active-highlight on `Tabs` and `TabNav`, where
  * `getGeometry` encodes whether the indicator is an edge bar (bottom or inner
  * side) or a filled highlight.
--- a/packages/nimbus/src/hooks/use-sliding-indicator/use-sliding-indicator.ts
+++ b/packages/nimbus/src/hooks/use-sliding-indicator/use-sliding-indicator.ts
@@ -1,0 +1,141 @@
+import { useLayoutEffect } from "react";
+import type { RefObject } from "react";
+
+/**
+ * Bounding rectangles passed to {@link SlidingIndicatorOptions.getGeometry}.
+ */
+export interface SlidingIndicatorRects {
+  /** Bounding rect of the positioned container the indicator lives in. */
+  container: DOMRect;
+  /** Bounding rect of the currently active item. */
+  active: DOMRect;
+}
+
+/**
+ * Geometry the indicator should adopt, in pixels, relative to the container's
+ * top-left. `width` / `height` are optional — omit one to leave it controlled
+ * by CSS (e.g. a fixed-thickness bar).
+ */
+export interface SlidingIndicatorGeometry {
+  x: number;
+  y: number;
+  width?: number;
+  height?: number;
+}
+
+export interface SlidingIndicatorOptions {
+  /**
+   * Whether the indicator is active. When `false` the hook is a complete no-op:
+   * it does not measure, observe, or mutate the indicator.
+   */
+  enabled: boolean;
+  /**
+   * Ref to the indicator element. Its positioned parent
+   * (`indicator.parentElement`) is used as the measurement container, so the
+   * indicator must be `position: absolute` inside a positioned ancestor.
+   */
+  indicatorRef: RefObject<HTMLElement | null>;
+  /** CSS selector (queried within the container) for the active item. */
+  activeSelector: string;
+  /**
+   * Attribute names whose mutation anywhere in the container should trigger a
+   * re-measure (e.g. `["aria-selected"]` or `["aria-current"]`).
+   */
+  watchAttributes: string[];
+  /**
+   * CSS selector for the items to observe for size changes (in addition to the
+   * container). Defaults to {@link activeSelector}'s tag-agnostic siblings via
+   * the container only when omitted.
+   */
+  itemSelector?: string;
+  /** Maps the container + active rects to the indicator's geometry. */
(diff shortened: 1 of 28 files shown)


misama-ct:
@claude review this branch

misama-ct:
@claude thanks for the thorough review — triaged all 7. Pushed `f99be868b` with the polish items; details and the one disagreement below.

**#1 (useLayoutEffect SSR warning) — skipped, premise doesn't hold on React 19.** This is a React-18-era warning. I verified against the installed deps: `@commercetools/nimbus` pins `peerDependencies.react` to `19.x`, and **react-dom 19.2.0 maps `useLayoutEffect` to a silent `noop` in the server renderer** — the *"useLayoutEffect does nothing on the server"* `error()` only exists in the react-dom 18 builds, it was removed in React 19. So nothing logs here, on the server or otherwise. Adding an isomorphic shim would suppress a warning that never fires while introducing an indirection the rest of `nimbus/src` doesn't use (zero other direct `useLayoutEffect`, no existing shim — because none is needed). If we ever widen the peer range back to include React 18, this becomes worth revisiting (and at that point `@react-aria/utils` already exports the isomorphic version, so still no hand-rolled shim).

**#3 (why `enabled` is always true) — done.** Added a comment at both call sites: it's intentional generic-hook flexibility, the slide is on by design with no per-instance toggle.

**#4 (duplicated alias resolution) — done.** Extracted a shared `resolveVariantAlias(value, aliases)` util; both roots now call it with their own alias map. Behavior-preserving.

**#2 (re-measure on scroll) — added a JSDoc note instead of code.** Worth clarifying: the `overflowX: "auto"` you spotted is on the **`tab`** slot (each tab), not the `list` strip — the strip itself has no overflow container and `flexShrink: 0`, and `tab-nav.recipe.ts` has no overflow at all. So the scrollable-strip drift isn't reachable with today's recipes. I documented that the hook doesn't track scroll so a future scrollable caller knows it'd need to observe scroll itself.

**#5 (effect deps) — no change**, agree it's correct + documented as-is.

**#6 (changeset bump scope) — confirmed intentional.** `.changeset/config.json` has a `fixed` group of exactly those 5 packages (nimbus, tokens, icons, ts-plugin, mcp), so the linked minor bump is by design, not a stray.

**#7 (global no-op router) — confirmed safe.** It only intercepts React Aria client-side navigation; `href` attributes still render so href-based assertions are unaffected, and no story relies on real iframe navigation. Good as a global default.

Verified: typecheck + lint clean, Tabs/TabNav story tests 23/23 pass.

packages/nimbus/src/hooks/use-sliding-indicator/use-sliding-indicator.ts
```diff
@@ -0,0 +1,180 @@
...
+ * The hook is DOM-driven: it measures with `getBoundingClientRect` inside
+ * `useLayoutEffect` (so the indicator is placed before paint — no first-frame
+ * flash), re-measures when the active item changes (`MutationObserver` on
+ * {@link SlidingIndicatorOptions.watchAttributes}) and when the layout changes
+ * (`ResizeObserver`), and hides the indicator (opacity 0) when there is no
+ * active item. The caller owns the indicator's appearance and transition; this
+ * hook only writes `opacity`, `width`/`height` (per `getGeometry`), and
+ * `transform`. The very first placement is applied with the transition
+ * momentarily suppressed so the indicator snaps to the active item instead of
+ * sliding in from the container corner on initial render.
```
tylermorrisford, line 73: that's 🧑‍🍳 cool!

tylermorrisford (approved):
hopefully Claude will stop yelling about needing a shim for useLayoutEffect in SSR contexts... this is really cool. Thank you!

valoriecarli (approved):

PR #1695, opened 2026-06-29
FCT-1100 - Wire up Chromatic

valoriecarli:
**Stands up Chromatic VRT & includes:** 
- a CI workflow
- Storybook-side config to make snapshots deterministic & opt-in
- the first set of components baselined
- stabilization fixes across the rest of the story suite so a full Chromatic build runs clean.
 - Foundational plumbing (rolldown lazyBarrel/CodeRoot fix + play-function timing races) was pulled in from PR (#1704) & rebased in as the base layer. 

**CI setup** (.`github/workflows/chromatic.yml)`
  - Runs on PRs (skips drafts), pushes to main, & a manual `workflow_dispatch`.
  - A changed-files gate (`tj-actions/changed-files`) only spends snapshots when files that affect rendered output change (`packages/nimbus, tokens, nimbus-icons`); ignores `config/main.ts/package.json.` `workflow_dispatch` bypasses the gate for a full baseline.
  - TurboSnap (onlyChanged) snapshots only diff-affected stories on normal runs, full on dispatch.
  - `exitOnceUploaded` returns as soon as the build is uploaded (before diffing), so the Chromatic / chromatic job goes green and never times out; Chromatic then diffs on its servers and reports the real verdict via the async "UI Tests" check.
  - Full rationale documented in the new runbook `docs/chromatic-visual-testing.md`.

  **Snapshot model** (`.storybook/preview.tsx + preview-head.html`)

  - Snapshots are opt-in: global default chromatic: { disableSnapshot: true }; a story opts in with tags: ["vrt"] + disableSnapshot: false.
  - A Date shim (preview-head.html) offsets window.Date from a fixed anchor (not freezing it) so date-dependent components stop drifting daily while timers/debounces/Date.now() IDs still work.

**Components baselined so far**
  - The whole `Button` family: button, split-button, icon-button, icon-toggle-button, toggle-button, toggle-button-group
  - `Avatar` (static demo image + play function waits for both broken images to error/hide before snapshot)

  These are the ones with real VRT stories (tags: ["vrt"] + snapshots enabled), tightened to capture meaningful states (focus rings, disabled+tooltip, etc.).

Small stabilization fixes to everything else — no snapshots, just making the run deterministic/error-free: timing waitFor guards, snapshot-disable on manual-observation-only stories, & date/render stabilization across calendar, combobox, markdown, radio-input, region, rich-text-input, splitter, tabs, tree, collapsible-motion, floating-action-button, & form-dialog.

Docs/tooling kept in sync: new runbook, stories.md + writing-stories skill updated with the opt-in patterns, housekeeping command gained a Chromatic dispatch step, & both CLAUDE.md files updated.

***
⚠️ Follow-up (post-merge): make UI Tests a required check

This PR makes the UI Tests check report on every PR (real Chromatic verdict on UI PRs; an auto-passing status on non-UI PRs so it's never left unreported). It does not block merges yet

After merge (repo admin): add UI Tests to the required status checks on main via GH rules.

Why wait? The auto-pass step only exists once this lands on main. Requiring UI Tests before then would hang every non-UI PR on a check that never posts.

Note: coverage is currently avatar + the button family only, so only those can produce a blocking diff for now.

****

//future self TODOs:
Chromatic's interaction-test reporting necessary if we have vitest running these?
Storybook publishing twice?
real gaps (hover?)

consider breaking the more complex ones down (eg splitting a large RichTextInput down to snapshot italics, bold, etc)
comb thru tests that really dont need a snapshot
clean up CI
review GHA output
review docs
add documentation, etc in nimbus relating to chromatic

should FAB be under Button folder or Pattern?


packages/nimbus/src/components/rich-text-input/rich-text-input.stories.tsx
```diff
@@ -67,6 +67,41 @@ const meta = {
...
+    // interaction step; the real assertion is the single waitFor below.
+    for (let i = 0; i < 10 && !landed(); i++) {
+      await new Promise((resolve) => {
+        setTimeout(resolve, 50);
+      });
+    }
+  }
+  await waitFor(() => expect(editor).toHaveTextContent(text));
+};
+
```
valoriecarli, line 104: Not my fav solution, but the only one that landed for now. 

These stories type into a **Slate** editor & a naive `click` + `type` is flaky in Chromatic's capture browser (and the Storybook manager UI), while passing locally. 
Two Slate-specific problems compound:

  1. **Caret race.** After `userEvent.click`, Slate adopts the caret
     asynchronously 
(browser `selectionchange` → 
Slate's rAF/throttled handler → 
editor.selection` → React render). 

At full speed `userEvent.type` fires *before* `editor.selection` exists, so Slate silently drops the keystrokes. It passes when you step the debugger only b/c the human pause lets Slate sync.

  2. **Placeholder node.** On an empty editor the click can land the caret on the `[data-slate-placeholder]` node (contenteditable=false), which Slate can't map to a text position, so typing has nowhere to go.

  The helper fixes both by construction: it places the caret in the real editable leaf (`[data-slate-string]` / `[data-slate-zero-width]`, never the placeholder), types, then **polls until the text actually lands** & retries (bounded) if the keystrokes raced the selection sync. 
The single `waitFor(toHaveTextContent)` at the end is the real assertion.

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,87 @@
...
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+
+      # Gate: run only when files that affect rendered output change (a manual
+      # dispatch bypasses it via the step `if:` conditions below). See runbook
+      # "The changed-files gate".
+      - name: Get changed files
+        id: changed-files
+        uses: tj-actions/changed-files@v46
```
valoriecarli, line 43: Additional info:
  - changed-files gate decides whether the job runs at all. If a PR touches zero rendering-affecting files, it skips install + package build + Storybook build + upload entirely, saving GHA minutes & avoids consuming a Chromatic build.
  - TurboSnap runs after all that work is done. It always installs, builds packages, builds Storybook, and uploads to Chromatic, _then_ uses the git diff to decide which stories to snapshot. It saves
  snapshot count (the main billable unit), nothing upstream of the upload.

.claude/commands/housekeeping.md
```diff
@@ -349,6 +349,25 @@ For each dependency group (or the specified target group):
      deps, only list that one)
    - Include the changeset in the final commit before pushing
 
+7. **Trigger a Chromatic visual check (when rendering deps changed):**
+
+   The Chromatic gate ignores `packages/nimbus/package.json`, so a
```
misama-ct: Huh? I'd argue that a dependency update causing a (visual) regression is a fairly standard case and not always will a dependency update happen via the housekeeping command. Merging dependabot PRs right away is still something I do regularly. 

Is that something we can change?

packages/nimbus/.storybook/preview.tsx
```diff
@@ -63,12 +64,31 @@ const preview: Preview = {
...
   },
+  // Chromatic snapshots at the end of the play function, so a focus-visible ring
+  // left on the last-interacted element bleeds into the diff. Blur it unless the
+  // story opts in via the `preserveFocusRing` parameter (i.e. it is testing focus
+  // or its final state legitimately keeps focus).
+  async afterEach(context) {
+    if (context.parameters?.preserveFocusRing) return;
+    const active = context.canvasElement.ownerDocument.activeElement;
+    if (active instanceof HTMLElement) active.blur();
+  },
```
misama-ct: I found this option earlier in the skill file and was immediately wondering _'What is this option for? Why was the outline removed at all?`_  

I don't think this option is necessary. I think testing two things in the same story is the actual issue and we should rather separate the visual- from the behavior-tests in the future whenever we touch existing story files.

packages/nimbus/.storybook/preview-head.html
```diff
@@ -0,0 +1,35 @@
+<script>
+  // Chromatic-only: override window.Date so date-dependent snapshots don't drift.
+  // OFFSET from a fixed anchor, don't freeze: today() stays May 15 while Date.now()
+  // keeps advancing, so timers/debounces/async loads resolve and Date.now() IDs stay
+  // unique. Runs here (pre-bundle) because today() is read at module load.
+  // Full rationale: docs/chromatic-visual-testing.md "Deterministic dates in snapshots".
```
misama-ct, line 6: Reading this, it sounds logical to me that tests need the same date every time they run, but that applies to all tests, not just to chromatic ones.

I'd argue that the solution is to fix our tests: We make tests deterministic rather than patching the global Date function to make them appear deterministic.

Not to mention this override could cause other unwanted side-effects (we're not aware of yet).

I'd get rid of this entirely.

package.json
```diff
@@ -17,6 +17,7 @@
     "build:docs-data": "pnpm --filter docs build:docs",
     "build:docs": "pnpm --filter './apps/docs' build",
     "build:tokens": "pnpm --filter @commercetools/nimbus-tokens run build && preconstruct build",
+    "build-storybook": "pnpm --filter @commercetools/nimbus build && pnpm --filter @commercetools/nimbus run build-storybook",
```
misama-ct: nit: `build:storybook`

.github/workflows/chromatic.yml
```diff
@@ -0,0 +1,95 @@
...
+      'pull_request'
+    permissions:
+      contents: read
+      # Skip path posts a passing "UI Tests" status so a required check still
+      # reports when Chromatic doesn't run. See runbook "Merge gating".
+      statuses: write
+    steps:
+      # Full history: Chromatic needs it to find the baseline build to diff against.
+      - name: Checkout repository
+        uses: actions/checkout@v5
```
tylermorrisford, line 34: These three action refs are pinned by tag rather than by commit SHA — `actions/checkout@v5` here, `tj-actions/changed-files@v46` on line 43, and `chromaui/action@v18` on line 83. The rest of the repo's workflows were just migrated to SHA-pinning as part of supply chain hardening, so noting this here for the rebase.

`tj-actions/changed-files` is worth calling out specifically — it was the action at the centre of the March 2023 supply chain incident. The fix is mechanical: look up the commit SHA for each tag and apply the `@sha # vX.Y.Z` pattern used in every other workflow file.

packages/nimbus/src/components/button/button.stories.tsx
```diff
@@ -258,21 +270,15 @@ export const DisabledWithTooltip: Story = {
...
       );
     });
-
-    await step("Tooltip closes on blur", async () => {
-      button.blur();
-      await waitFor(() =>
-        expect(canvas.queryByRole("tooltip")).not.toBeInTheDocument()
-      );
-    });
   },
```
tylermorrisford, line 277: The previous version of this play function had a "Tooltip closes on blur" step here — `button.blur()` followed by `expect(canvas.queryByRole("tooltip")).not.toBeInTheDocument()`. The story now ends with the tooltip open (intentional, for the VRT snapshot), but the close-on-blur behaviour is no longer exercised in this story.

If the Tooltip component's own stories cover this path, that's probably fine — just worth confirming. A tooltip that never dismisses is an accessibility failure so it's a path worth keeping green somewhere.

.claude/commands/housekeeping.md
valoriecarli: @misama-ct This was done due to cut costs, but agree that it was too aggressive & finicky.
Changes have been made in https://github.com/commercetools/nimbus/pull/1695/commits/51255e0f139f903e7887c25c9a098fe07c0cab3f  (remove the `allowList`,  resume watching `pnpm-lock, update docs, etc)`

packages/nimbus/.storybook/preview.tsx
valoriecarli: @misama-ct Thank you for catching this, rn my eyes went right past it. This was part of a pattern I tried early on to do a mass clean up, which we ultimately abandoned in favor of a proper test clean up. 
As we will be combing through the components in sections & explicitly requesting snapshots, this was removed w/o issue in https://github.com/commercetools/nimbus/pull/1695/commits/8d597acf3aca90f0bc7af7f3f239f1af995ca40a

YA GONE!

packages/nimbus/.storybook/preview-head.html
valoriecarli: @misama-ct, I will pay you a MILLION Euros if I can tackle the `Date` component family next and remove this then. You know I am good for it.

package.json
valoriecarli: @misama-ct corrected in https://github.com/commercetools/nimbus/pull/1695/commits/9aadc4a1f9e3b7648eafff0d71ec8c391c24da4d
Note that I only renamed the root build:storybook script & its in-repo callers due to Vercel invocation. 

If we want this updated everywhere like I did originally with https://github.com/commercetools/nimbus/pull/1695/commits/58c19f28057032aa8d8342d3a6b0bf50026bbccd, I can do that in a next step. Otherwise [splat](https://vercel.com/commercetools/nimbus-storybook/GKSMgrvbpuR2EYyC84W28fzHgbgj)!

.github/workflows/chromatic.yml
valoriecarli: @tylermorrisford and I spoke about this gem irl, so resolving the comment for now. This will be handled once this PR is merged. Thanks!

packages/nimbus/src/components/button/button.stories.tsx
valoriecarli: @tylermorrisford   _Great_ q & exactly the kind of scrutiny I'm looking for as I start combing through these.
I opted not to add a snapshot here for two reasons:

1. It's redundant visually. The 'no tooltip present' appearance is the default render of almost every story in the library, so a dedicated 'dismissed' snapshot would duplicate what's already baselined.

2. It's already covered functionally. `tooltip.docs.spec.tsx` proves this w/ real tab-away navigation & asserts the tooltip leaves the DOM (the actual a11y guarantee). 
A snapshot could only infer it from absent pixels, adding resource cost on a frame w/ no unique visual signal & it couldn't distinguish 'dismissed correctly' from 'never opened'.

Could be splitting hairs, but that's the reasoning as I worked through it at the time. If you disagree, I'm more than happy to add a snapshot for this scenario 🙌
Either way, I will make sure to comment on these decisions in future PRs.

packages/nimbus/.storybook/preview-head.html
tylermorrisford: I agree with Michael here - but I'm cool with it if it's getting killed on the next pass

packages/nimbus/src/components/button/button.stories.tsx
tylermorrisford: love this - thank you!

tylermorrisford (approved):
Other than Michael's suggestion, this looks great and was a huge effort. Yes I know the effort continues, but still!

.claude/commands/housekeeping.md
valoriecarli: Though I have reverted this idea for now, I am still not convinced that Chromatic needs to be running on _all_ dep updates. I know we have groupings that will cut down on runs, but just unsure of how much will be running once Renovate is introduced, and what will be running even on the automated dependabot prs that are closed out. 

I will need to keep 👀  on this to see if we need to find a better balance.

tylermorrisford: I wonder if having specific groupings of smoke tests running only in cases where deps change or on renovate/housekeeping PRs would be wise, or if it will just hide significant issues

valoriecarli: Ohh... I like where you're going with this, some kind of middle (automated) ground. 🤔

misama-ct: The [chromatic config file](https://github.com/commercetools/nimbus/pull/1695/changes#diff-a21748e8f557f5ccebf3a1bfd7b063dcbed457203122344cf52d0f29d59f6a78R4) suggests that there is _some_ sort of mechanism, trying to determine what changed.

misama-ct (approved):
🙇‍♂️ Thank you. Looking forward to see it in action.

PR #1696, opened 2026-06-29
FEC - Imports

valoriecarli:
> **Background**

Chromatic fails to build our Storybook with `ReferenceError: CodeRoot is not defined`([here](https://github.com/commercetools/nimbus/actions/runs/28395539421/job/84133207699#step:4:66)) (once fixed, the error moves on to `DefaultPageRoot` ([here](https://github.com/commercetools/nimbus/actions/runs/28389164441/job/84111491872#step:6:66)), usw). 

During discovery, it was uncovered that this issue can also be seen in Storybook prod under various components:
[datatable](https://nimbus-storybook.vercel.app/?path=/docs/components-datatable--docs) & [avatar](https://nimbus-storybook.vercel.app/?path=/story/components-avatar--base), usw.

These are runtime errors thrown by our own bundled code: a chunk references a variable that was never declared in the output.

The chain:
1. Many component source files import the mega-barrel `@/components` (e.g. `icon-button.tsx` → `import { Button } from "@/components"`). That barrel re-exports everything, including code & default-page.
  
2. In a prod Storybook build, `@commercetools/nimbus` resolves to the compiled dist, but the `@/` alias always points at source. So those imports drag the raw source barrel into the bundle alongside dist.
  
3. Our compound components define their root in a separate module & re-export it (`export const Code = CodeRoot`). Rollup (under Vite 8) tree-shakes away the module that declares CodeRoot but keeps the consumer that references it, leaving a dangling free variable that throws when the chunk loads.

> **How we resolved it**

We fixed the root cause: **migrated all ~59 source files off the `@/components` mega-barrel** to deep imports (`@/components/<name>`), which is what our own CLAUDE.md "Cross-Chunk Imports" rule already requires. Each component now imports only the specific sibling it needs, so the mega-barrel is never pulled into the bundle.

Changes:
- **59 source files**: Replaced `import { X } from "@/components"` with `import { X } from "@/components/button"` (etc.)
- **8 stories/helpers**: Changed `@/components` to `@commercetools/nimbus` (resolves to dist in prod builds)
- **ESLint guard**: `no-restricted-imports` rule banning the bare `@/components` barrel across ALL of `packages/nimbus/src/` — not just stories/utils
- **Removed build workaround**: The prod-only `@/components→dist` alias in `.storybook/main.ts` is no longer needed and has been removed

> **Why we're only seeing it once we started on Chromatic?**

It's a runtime error, not a build error, so `build-storybook` succeeds & publishes. 
In the deployed Storybook, each affected story is caught by Storybook's per-story error boundary & shows an error panel - the site stays up. 
Chromatic extracts every story via `page.evaluate`, outside that error boundary, where the same thrown error is fatal & aborts the whole run.

> **Why this wasn't detected before?** 

- Known: Chromatic worked ~3 months ago, so this is relatively recent.
- Theory (not proven!) - the barrel mis-imports have existed since late 2025 & were harmless under Vite 7, its tree-shaking didn't drop the re-exported roots. [PR #1451](https://github.com/commercetools/nimbus/pull/1451) (Vite 7 → 8) is assumed to have changed that behavior.
- Nothing in CI builds the production Storybook & exercises every story. Dev Storybook serves modules individually (no production tree-shaking, so the bug can't appear) & the published Storybook's per-story error boundaries hide it during normal browsing. 
- Chromatic just happened to be the first thing implemented since that does a production build & evaluates all stories.

> **Prevention**

- **ESLint guard (in this PR)**: bans the bare `@/components` barrel across ALL source files in the nimbus package.
- A production-build smoke check in CI: run build-storybook & fail on any chunk that references an undeclared identifier. This catches the failure class w/o needing Chromatic.
- Longer term, Chromatic itself (once landed) becomes the standing guard, since it's the only check that does a full production build across every story.


tylermorrisford (approved):
Seems sensible - let's do it and roll it back if 🤯

misama-ct (changes requested):
A recent vite/rolldown version update caused the issues in the original branch. I made another PR with a [minimal fix](https://github.com/commercetools/nimbus/pull/1704) that just turns off a feature flag (which is turned on by default in latest vite/rolldown). 

There certainly is something wrong with dependency resolution, and this could potentially even be the fix for story-files, but I am quite sure it does not just affect story-files, they just made it obvious.

PR #1697, opened 2026-06-29
fix(scripts): use Node built-in globSync in clean.mjs

ByronDWall:
## Summary
- `pnpm nimbus:init` fails on a fresh clone because `scripts/clean.mjs` imports the `glob` npm package, but `node_modules` doesn't exist yet (chicken-and-egg: reset deletes node_modules, then the clean script can't run without its dependency).
- Replaces the external `glob` import with Node's built-in `fs.globSync` (available since Node 22) so the script has zero external dependencies.
- Adds a `matchesIgnore` helper to replicate the `ignore` option behavior via post-filtering, since `fs.globSync` uses a different `exclude` API.

note: this is especially helpful to support working with git worktrees

## Test plan
- [x] Verified `node scripts/clean.mjs --dry-run` works without `node_modules`
- [x] Verified `node scripts/clean.mjs --all --node-modules --dry-run` finds all expected targets (86 items)
- [ ] Run `pnpm nimbus:init` from scratch on a fresh clone

### DIFF
--- a/scripts/clean.mjs
+++ b/scripts/clean.mjs
@@ -48,12 +48,17 @@
  * pattern matches what you expect and nothing more.
  */
 
-import { readdirSync, rmSync, existsSync, statSync } from "node:fs";
+import {
+  readdirSync,
+  rmSync,
+  existsSync,
+  statSync,
+  globSync,
+} from "node:fs";
 import { execSync } from "node:child_process";
 import { join, relative } from "node:path";
 import { fileURLToPath } from "node:url";
 import { styleText } from "node:util";
-import { globSync } from "glob";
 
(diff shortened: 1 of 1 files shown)


misama-ct (changes requested):
Good catch. I remember vaguely, that claude tried to use nodes' `glob` but unfortunately it couldn't look into dot-folders, so claude happily suggested (and I agreed) to use the glob-package instead, since it was already a dependency. I failed to see the obvious.

So yeah, I agree, a fix is needed, but the fix also needs to be able to delete the the specified dot-folders and their contents (`.cache/*.*` etc.).

valoriecarli (approved):

PR #1704, opened 2026-06-30
fix(storybook): unblock Chromatic rollout — rolldown lazyBarrel (CodeRoot) + play-function timing races

misama-ct:
## What

Makes the production Storybook build (Chromatic, added in #1695) green. It bundles **two independent fixes** plus a dependency-hygiene change — all of them surface **only in the built production bundle**, never in `pnpm test:storybook` / local dev (vitest runs through Vite's dev-transform; Chromatic runs the rolldown-bundled `build-storybook` output).

> Includes the changes from #1705 (merged into this branch).

## 1. Rolldown `lazyBarrel` → `ReferenceError: CodeRoot is not defined`

The production Storybook crashed at module-eval time (`CodeRoot is not defined`), aborting story extraction (seen on Code, Avatar, DataTable, …).

vite 8.1.0 ships rolldown 1.1.3, where `lazyBarrel` (default-on since rolldown 1.1.0) drops a still-referenced binding re-exported through a `sideEffects:false` `export *` barrel — here the `CodeRoot` that `Code` re-exports (`export const Code = CodeRoot`).

**Fix** — disable the flag for the Storybook build, staying on vite 8.1.0 (`.storybook/main.ts` `viteFinal`):

```ts
build: { rolldownOptions: { experimental: { lazyBarrel: false } } }
```

Upstream: rolldown/rolldown#9806, #9964, #9961. The flag is experimental and slated for removal upstream, so this is a temporary workaround for the **trigger**, not the underlying dev/prod module-resolution split (`@/*`→`src` vs `@commercetools/nimbus`→`dist`) that lets a source barrel and the dist barrel coexist in one production bundle. (Pinning vite to 8.0.16 was evaluated first — see history — and reverted in favour of the flag.)

## 2. Play-function async-settle timing races (was #1705)

After the crash was fixed, Chromatic surfaced interaction-test failures on **Region**, **Splitter**, **Tree**, and **FormDialog**. The stories render correctly; their **play functions** read async-settled state **synchronously** and lose a race that only the slower production bundle exposes:

- **Tree** — `getAllByRole("row")` / `getByRole("row")` before React Aria finishes building the collection (treegrid mounts `data-empty="true"`, rows arrive ~0.5–1s later).
- **Region** — `getByTestId(...)` before the region portal commits the projected node.
- **Splitter** — `aria-valuemin/valuemax` / `aria-controls` before the initial pane measurement settles them (read the pre-measurement default, e.g. `25` instead of `15`).
- **FormDialog** — `toBeDisabled()` on Save/Cancel/Close before the `isSaveDisabled` / `isSaveLoading` lockout applies after the dialog mounts.

**Fix** — wrap the racing reads in `waitFor` / `findBy*` so each assertion waits for the collection build / portal commit / measurement / lockout to settle. No component source changes — the components already render correctly once settled. (Subtlety: `findByRole` waits for the element to *exist*, not for its settled attribute/children — so the *value* checks are wrapped in `waitFor`.)

## 3. vite catalog dedup

`vite: catalog:tooling` added to the root and `design-token-ts-plugin` manifests. They declared `vitest` but no `vite`, so pnpm resolved their vitest peer independently and drifted off the catalog; with an explicit catalog dep the lockfile holds a single vite version.

## Verification

- Production Storybook builds; running the previously-failing play functions against the built `storybook-static` (the Chromatic-equivalent surface): all pass — including under CPU throttling, which reproduces the FormDialog `toBeDisabled` race on the old code and confirms the fix.
- `pnpm --filter @commercetools/nimbus typecheck`: clean.
- The timing changes only *add* waiting, so the source/vitest path is unaffected.

## Follow-ups (intentionally NOT in this PR)

- **Close the CI coverage gap**: nothing runs play functions against the *built* Storybook in CI (only Chromatic does), which is why this whole class slips through. A `@storybook/test-runner` step against `storybook-static` would catch it pre-Chromatic.
- **Recurring overlay-close flake**: `dialog` / `info-dialog` (and likely ModalPage / Drawer / ConfirmationDialog) intermittently fail `expect(queryByRole("dialog")).not.toBeInTheDocument()` — the *mirror* of the races fixed here (an element lingering through its exit animation past the `waitFor` timeout, rather than appearing late). Needs its own fix (close-`waitFor` timeout and/or disabling exit animations under test).

Stacked on #1695; base is that branch.

### DIFF
--- a/packages/design-token-ts-plugin/package.json
+++ b/packages/design-token-ts-plugin/package.json
@@ -25,6 +25,7 @@
     "@commercetools/nimbus-tokens": "workspace:^",
     "@types/node": "catalog:utils",
     "typescript": "catalog:tooling",
+    "vite": "catalog:tooling",
     "vitest": "catalog:tooling"
   },
   "files": ["dist"]
--- a/package.json
+++ b/package.json
@@ -114,6 +114,7 @@
     "tsx": "catalog:tooling",
     "typescript": "catalog:tooling",
     "typescript-eslint": "catalog:tooling",
+    "vite": "catalog:tooling",
     "vite-bundle-analyzer": "catalog:tooling",
     "vitest": "catalog:tooling"
   },
--- a/packages/nimbus/.storybook/main.ts
+++ b/packages/nimbus/.storybook/main.ts
@@ -103,6 +103,22 @@ const config: StorybookConfig = {
             }
           : {},
       },
+      build: {
+        // Disable rolldown's lazy-barrel optimization (default-on since
+        // rolldown 1.1.0, shipped in vite 8.1.0). It drops a still-referenced
+        // binding re-exported through a `sideEffects:false` `export *` barrel —
+        // here the `CodeRoot` binding that `Code` re-exports — so the
+        // production Storybook bundle throws `ReferenceError: CodeRoot is not
+        // defined` at module-eval time, aborting story extraction.
+        // Upstream: rolldown/rolldown#9806, #9964, #9961.
+        // Remove once a fixed rolldown ships (the flag itself is slated for
+        // removal upstream).
+        rolldownOptions: {
+          experimental: {
+            lazyBarrel: false,
+          },
+        },
+      },
     });
   },
 };
--- a/packages/nimbus/src/components/region/region.stories.tsx
+++ b/packages/nimbus/src/components/region/region.stories.tsx
@@ -71,8 +71,11 @@ export const ProjectIntoNamedRegion: Story = {
       "Content authored in one place paints at the target",
       async () => {
         const host = canvasElement.querySelector<HTMLElement>("#sidebar-host")!;
-        const projected = canvas.getByTestId("projected");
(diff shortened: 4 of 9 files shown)


packages/design-token-ts-plugin/package.json
```diff
@@ -25,6 +25,7 @@
     "@commercetools/nimbus-tokens": "workspace:^",
     "@types/node": "catalog:utils",
     "typescript": "catalog:tooling",
+    "vite": "catalog:tooling",
```
misama-ct, line 28: I explicitly added `vite` here (and [here](https://github.com/commercetools/nimbus/pull/1704/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R117)) as dependency, to prevent pnpm from pulling in the version specified as `vitest`'s peer-dependency .

valoriecarli (approved):
thanks for doing this, and addressing some of these tests. i know exactly what happened here.

PR #1713, opened 2026-07-02
feat(drop-zone): add DropZone component

misama-ct:
Components - DropZone - Docs ⋅ Storybook; Jul 2, 4 13 PM

## Summary

Adds a new **`DropZone`** component to `@commercetools/nimbus` — a styled,
accessible drag-and-drop target for accepting files or other dragged data,
built as a thin wrapper over React Aria's headless `DropZone` primitive.

## Motivation

Nimbus had no drop target for upload experiences. Consumers building asset
managers, CSV importers, or avatar pickers had to either hand-roll drag-and-drop
with raw DOM events — easy to get wrong for keyboard and screen-reader users — or
fall back to a bare `FileTrigger` button with no drag affordance. React Aria
ships an accessible primitive; this wraps it with design-system styling and a
sensible default state so teams get an on-brand, WCAG 2.1 AA drop target for
free.

## What changed

- **New component** at `packages/nimbus/src/components/drop-zone/` wrapping
  `DropZone` from `react-aria-components`. Deliberately minimal (MVP): it adds
  only the drag-and-drop and accessibility mechanics — content and layout stay
  the consumer's to provide.
- **Recipe** (`nimbusDropZone`) styling only the states React Aria exposes on
  the container: idle (dashed border), hover, dragOver (`data-drop-target`
  highlight), focus (`data-focus-visible` ring), and disabled. Registered in
  `theme/recipes/index.ts`.
- **Default state**: a bare `<DropZone />` renders a centered upload icon + one
  localized instruction line ("Drag and drop files here") as a
  `<Text slot="label">`, so it's never an empty box and is labelled for free.
  Passing children replaces the default entirely (all-or-nothing).
- **Faithful RAC forwarding**: `onDrop`, `getDropOperation`,
  `onDropEnter/Move/Exit/Activate`, and `isDisabled`; re-exports the `DropEvent`,
  `DropOperation`, and `DragTypes` types so consumers can type handlers without
  depending on `react-aria-components` directly.
- **Click-to-upload path**: not re-implemented — consumers compose the existing
  Nimbus `FileTrigger` as children (keyboard/SR drag can't reach the OS file
  picker). Demonstrated in stories and docs.
- **i18n**: one new message `Nimbus.DropZone.defaultLabel` across all locales.
- **Docs & tooling**: Overview / Guidelines / Implementation / Accessibility doc
  tabs, Figma Code Connect, and the OpenSpec change under
  `openspec/changes/add-drop-zone-component/`.

## Accessibility

Keyboard- and screen-reader-accessible by default (Enter to start drag mode, Tab
between targets, Enter to drop, Escape to cancel). Idle, hover, drag-over, focus,
and disabled states meet WCAG 2.1 AA — non-color-only drag-over feedback,
forced-colors support, and `prefers-reduced-motion` handling.

## Test plan

- [ ] Storybook play functions in `drop-zone.stories.tsx` (idle / hover /
      drag-over / focus / disabled states, drop handling, `FileTrigger`
      composition) — run `pnpm test:storybook:dev packages/nimbus/src/components/drop-zone/drop-zone.stories.tsx`
- [ ] Consumer examples in `drop-zone.docs.spec.tsx`
- [ ] `pnpm typecheck` and `pnpm lint`

## Feedback wanted

- Is the all-or-nothing children/default-state behavior the right call, or
  should there be icon/title/description slots?
- Naming and shape of the re-exported RAC types (`DropEvent`, `DropOperation`,
  `DragTypes`).

Closes FEC-986


misama-ct:
@claude please review

misama-ct:
@claude Thanks — addressed the review.

**Main point (aria-label blanking content):** fixed by decoupling. The render now gates default content on `hasChildren` only; `hasExplicitLabel` still controls the `aria-label` / `slot={null}` wiring. A custom `aria-label` now overrides only the accessible name — the default icon + instruction stay visible. This works cleanly because the existing `slot={null}` on the default `<Text>` already keeps the visible text out of the accessible-name computation, so there's no name/visible conflict. Updated the `AccessibleName` story and the dev.mdx (Accessible-name section, Labeling bullet, Notes & limitations) to match.

**Nit 2:** added a comment clarifying the `getDropOperation` case is an illustrative consumer pattern, not a component assertion.

**Nit 1** (typed `slot` helper) and **Nit 3** (RAC type re-exports): skipped — Nit 3 you agreed to keep as-is, Nit 1 is a non-blocking style tweak.

Verified locally: `typecheck` clean, drop-zone `lint` clean, `test:storybook:dev` 8/8 passing.

packages/nimbus/src/components/drop-zone/drop-zone.stories.tsx
```diff
@@ -0,0 +1,316 @@
...
+// Composition (children replace the default) + disabled
+// ============================================================
+
+export const WithFileTriggerComposition: Story = {
+  args: {
+    "data-testid": "drop-zone-root",
+  } as never,
+  render: (args) => (
+    <DropZone {...args}>
+      <Text slot="label">Drop files, or browse to choose them</Text>
```
tylermorrisford, line 189: Extends the earlier `@claude` review thread on the `aria-label` / `slot={null}` wiring. That discussion covered the _default-content_ and _explicit-`aria-label`_ paths and concluded there's "no name/visible conflict" — but it didn't reach this **composed-children** path, which does have one.

When children are composed **without** an explicit `aria-label`/`aria-labelledby`, `drop-zone.tsx:72` still forwards the default `aria-label="Drag and drop files here"` (it gates on `hasExplicitLabel`, not `hasChildren`). And this `<Text slot="label">` is Nimbus's `Text`, which does **not** wire into React Aria's `LabelContext`, so it contributes nothing to the accessible name. I verified in jsdom against source:

| rendered                                                                             | accessible name                        |
| ------------------------------------------------------------------------------------ | -------------------------------------- |
| this story — Nimbus `<Text slot="label">Drop files, or browse to choose them</Text>` | **`Drag and drop files here`**         |
| same, but React Aria's `Text` (`react-aria-components`)                              | `Drop files, or browse to choose them` |
| `<DropZone aria-label="Upload area">…</DropZone>`                                    | `Upload area`                          |

So the composed zone announces the stale default label, not the visible instruction the consumer wrote — a WCAG 2.5.3 (Label in Name) mismatch, and `slot="label"` silently has no effect when it's a Nimbus `Text`.

The play functions here and in `CustomContent` (line 231) only assert `getByText(...)` / that the default text is _not visible_ — never `toHaveAccessibleName`, so this slips through. Note the assertion at line 204-205 checks the default text isn't _rendered_ (true — with children the `slot={null}` line isn't emitted), which is easy to read as "the default label is gone" — but it survives as the button's `aria-label`.

Options (?):

- **Minimum:** add `toHaveAccessibleName(...)` to these composition stories — that alone surfaces the mismatch (the `Default` and `AccessibleName` stories already assert it at lines 61 and 91, so there's precedent).
- **Real fix:** either gate the default `aria-label` on `!hasChildren` _and_ steer consumers to an explicit `aria-label` for composed content (gating alone just swaps the wrong name for RAC's English-only `"DropZone"` fallback, since Nimbus `Text` still won't drive the name), or make Nimbus `Text` participate in RAC's label slot so `<Text slot="label">` works as the spec assumes.

The spec assumes the latter — see the related drift at `spec.md:54`.

openspec/changes/add-drop-zone-component/specs/drop-zone/spec.md
```diff
@@ -0,0 +1,168 @@
...
+
+- **WHEN** DropZone is rendered without a composed `FileTrigger`
+- **THEN** it acts as a drag-only drop target and does not render any upload
+  button of its own
+
+### Requirement: Default state and content
+
+When given no children, DropZone SHALL render a default state — a centered upload
+icon and one localized instruction line (`Nimbus.DropZone.defaultLabel`), the
+label rendered as a React Aria `<Text slot="label">` so it also serves as the
```
tylermorrisford: Spec drift, related to the stories finding at `drop-zone.stories.tsx:189`.

This says the default label is "rendered as a React Aria `<Text slot="label">` so it also serves as the accessible name," but the shipped implementation uses `aria-label` + `<Text slot={null}>` for the default — the name comes from `aria-label`, not from a `slot="label"` element.

More consequentially, lines 145 and 152-153 tell consumers they "MAY instead label it with a `<Text slot="label">` child." With Nimbus's `Text` that doesn't set the accessible name at all — I verified `aria-labelledby` stays empty; only React Aria's own `Text` wires into the label slot. So this labelling path doesn't work as written.

Worth reconciling the spec with what the component actually does, or making `<Text slot="label">` genuinely drive the name so the spec holds.

tylermorrisford (approved):
One note about the construction of the label - but regardless, a really nice addition ❤️

ByronDWall (approved):

PR #1726, opened 2026-07-07
feat(chat): add ChatMessage and ChatMessageList components (FEC-987)

misama-ct:
## Summary

Adds two new compound components for building AI chat / assistant experiences:
**`ChatMessage`** (a single message) and **`ChatMessageList`** (a scrollable
transcript that arranges messages, owns autoscroll, and hosts the live region
that announces streamed replies). Also updates `Avatar` (a new `solid` variant,
a custom-`children` fix, and one default-palette change) that the chat
components build on.

image

```tsx
// ═══════════════ ChatMessageList — the scrollable transcript ═══════════════
// Arranges the items; owns autoscroll, the live region, and jump-to-latest.
<ChatMessageList.Root aria-label="Conversation" height="360px">
  {/* ─────────── ChatMessage — a single (user) message ─────────── */}
  <ChatMessageList.Item>
    <ChatMessage.Root sender="user">
      <ChatMessage.Avatar firstName="Ada" lastName="Lovelace" />
      <ChatMessage.Body>
        <Text>Summarise last week's orders?</Text>
      </ChatMessage.Body>
    </ChatMessage.Root>
  </ChatMessageList.Item>

  {/* ────── ChatMessage — a single (agent) message, streaming ────── */}
  <ChatMessageList.Item>
    <ChatMessage.Root sender="agent" isStreaming>
      <ChatMessage.Avatar>
        <AutoAwesome />
      </ChatMessage.Avatar>
      <ChatMessage.Body>
        <Markdown isStreaming>{reply}</Markdown>
      </ChatMessage.Body>
    </ChatMessage.Root>
  </ChatMessageList.Item>
</ChatMessageList.Root>
```

## What changed

**`ChatMessage`** — parts: `Root`, `.Avatar`, `.Body`, `.Actions`, `.Meta`, `.Typing`.

- **`sender`** (`"user" | "agent"`, default `"agent"`) — layout
  direction + per-sender surface; `agent` leads with the avatar, `user`
  trails it.
- **`tone`** (`"neutral" | "error"`, default `"neutral"`) — orthogonal status
  overlay; `error` tints the body to flag a failed generation (an _agent_
  message can still fail, so error is a tone, not a sender).
- **`isStreaming`** sets `aria-busy`; **`ChatMessage.Typing`** renders an
  `ActivityIndicator` "generating…" affordance while a reply streams.
- **`ChatMessage.Avatar`** wraps the Nimbus `Avatar` (`xs`), auto-colored per
  sender; content via `firstName`/`lastName`, `src`, or an icon as `children`.
  Decorative by default unless named.
- **`ChatMessage.Body`** holds the payload; `.Actions` pins to its bottom;
  `.Meta` is a below-body row (timestamp/status).
- Renders a semantic `<article>` by default; `as` overrides the element on
  `Root` and on the layout parts (`.Body`, `.Actions`, `.Meta`, `.Typing`) as
  well as `ChatMessageList.Item`.

**`ChatMessageList`** — parts: `Root`, `.Item`.

- **`Root`** is the scroll container and the transcript's single
  `role="log" aria-live="polite"` region; owns autoscroll / stick-to-bottom and
  a "jump to latest" control.
- Props: **`autoScroll`** (default `true`), **`emptyState`**, and a `ref` handle
  exposing **`scrollToBottom()`**.
- **`.Item`** is content-agnostic — it holds a `ChatMessage` **or any other
  content**, so consumers render their own centered system notices / dividers as
  peers of messages rather than as a `sender` variant.

**`Avatar` (mostly additive — one visual change):**

- Custom `children` now render (previously documented but ignored).
- New `variant` prop — `subtle` (default, unchanged look) and `solid` (matches
  the `Button` solid look); crisper fallback-icon sizing.
- **Visual change:** a bare `<Avatar>` with no `colorPalette` now inherits the
  ambient palette (global default `neutral`) instead of a pinned `primary`,
  aligning `Avatar` with how `Button` resolves its palette. Avatars that already
  set a `colorPalette` (including `ChatMessage.Avatar`) are unaffected; pass
  `colorPalette="primary"` to keep the previous tint. See the changeset.

## Accessibility

- Each message names itself via `aria-label`/`aria-labelledby`; sender identity
  is **never** derived from the avatar.
- `ChatMessage.Avatar` is decorative (`aria-hidden`) unless named.
- The list uses `role="log"` (not `feed`) so mixed children — message
  `article`s plus consumer-rendered notices — stay valid, and appended replies
  are announced without re-reading the whole history. Per-token announcement of
  a streaming reply is governed by `aria-busy` (and coalesced when the body is a
  streaming `Markdown`). See the **Accessibility** docs tab for the full
  `log`-vs-`feed` rationale and how streaming announcement ownership is split
  between the list and the message.

## Notes

- Reshaped during review: `ChatBubble` → `ChatMessage` (`.Bubble` → `.Body`), a
  `ChatMessageList` transcript was added, and the exploratory `ChatNotice` leaf
  was dropped (too small a pattern to standardize — the content-agnostic `Item`
  already covers it). All unreleased, so there is no consumer-facing breaking
  change.
- OpenSpec: `add-chat-message` + `add-chat-message-list` changes authored and
  archived.

## Test plan

- [ ] `pnpm --filter @commercetools/nimbus typecheck:dev` — clean
- [ ] Storybook play functions (incl. axe) for `chat-message`,
      `chat-message-list`, and `avatar`
- [ ] `chat-message.docs.spec.tsx` + `chat-message-list.docs.spec.tsx` consumer
      tests
- [ ] `pnpm lint` — clean

## Feedback wanted

- Modelling `error` as an orthogonal `tone` prop rather than a fifth `sender` —
  reasonable?
- Leaving system notices / dividers to consumer-rendered content inside an
  `Item` (no dedicated Nimbus component) — agree?

Closes FEC-987


misama-ct:
@claude please review

misama-ct:
Dear @claude,

Thorough and fair — thank you. Addressed in `cf9a55665`.

**🔴 Avatar `colorPalette` default — resolved as an intentional, now-documented change.**
You're right that it's not "additive," and good catch that it's library-wide. This one is deliberate: `Avatar` is meant to resolve its palette the way `Button` does (inherit the ambient palette, global default `neutral`) rather than pinning `primary`. So we've **kept the neutral default and stopped denying it**:
- Changeset now calls out the visual change explicitly ("a bare `<Avatar>` shifts from the primary tint to neutral; pass `colorPalette="primary"` to keep the old look"), replacing the false "unchanged" line.
- PR body no longer claims "additive, non-breaking" for that bullet.
- Chat is unaffected (`ChatMessage.Avatar` sets its own palette), as you noted.

**🟡 1 — `sender` collapsing responsive values → fixed.** Narrowed the public `sender` prop to a non-responsive `"user" | "agent"` union, so a `{ base, md }` value is now a type error rather than something that styles the body per breakpoint while the context (avatar variant/glyph) silently flattens to `"agent"`.

**🟡 2 — smooth scroll swallowing a user scroll-up → fixed.** The guard now tracks distance during the animation and releases the pin the moment the user scrolls up (distance grows while the animation only ever shrinks it), so the jump-to-latest control reappears immediately instead of waiting for the animation to land.

**🟡 3 — `role="log"` on `ScrollArea.Root` vs. the viewport — conscious choice, no change.** Agreed on all counts; the live region is a stable ancestor of the mutating content and the focusable viewport is the inner node (which the jump handler focuses). `ChakraScrollArea.Root` does forward `role`/`aria-*` to a real DOM node — that's what the `getByRole("log")` story queries assert.

**🟢 Nits.** `isNamed` left as-is (agreed). The `nimbus-mcp` / `nimbus-icons` / `design-token-ts-plugin` minor bumps are changesets' automatic dependent-version propagation, not an over-wide scope in this PR — happy to trim if that's not desired. Localization Jira ticket for the two new strings: noted; preliminary de/es/fr-FR/pt-BR translations are already in `packages/i18n/data`.

**On the design questions** — thanks for the sign-off on `error`-as-`tone` and on skipping `ChatNotice`.

Happy now?

packages/nimbus/src/components/chat-message-list/hooks/use-stick-to-bottom.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  // Track the user's scroll position → derive whether we are pinned.
+  useEffect(() => {
+    const el = viewportRef.current;
+    if (!el) return;
+
+    const update = () => {
+      const distanceFromBottom =
+        el.scrollHeight - el.scrollTop - el.clientHeight;
+      const atBottom = distanceFromBottom <= BOTTOM_THRESHOLD_PX;
```
ByronDWall, line 151: **Suggestion: consider `IntersectionObserver` as a future alternative to per-scroll layout reads**

The `update` handler reads `scrollHeight`, `scrollTop`, and `clientHeight` on every scroll event. With `{ passive: true }` this is fine for typical transcripts, but for very long histories with rapid scrolling it could add up. An `IntersectionObserver` on a sentinel `<div>` at the bottom of the content flow can derive the same "am I at the bottom?" signal with zero per-scroll layout reads — the browser batches the geometry check internally.

Not blocking — the current approach is correct and performant for realistic workloads. Just flagging as an option if scroll perf surfaces as an issue down the road.

packages/nimbus/src/components/chat-message-list/hooks/use-stick-to-bottom.ts
```diff
@@ -0,0 +1,217 @@
...
+
+    // Start at the newest message. This effect runs after the scroll-tracking
+    // effect (which may have derived `pinned=false` from the initial
+    // scrollTop=0), so re-assert the pin here before observing growth.
+    if (enabledRef.current) {
+      el.scrollTop = el.scrollHeight;
+      setPinned(true);
+    }
+
+    const resizeObserver = new ResizeObserver(scheduleStick);
```
ByronDWall, line 194: **Suggestion: consider a comment explaining why `disposedRef` exists**

The `ResizeObserver` callback routes through `scheduleStick`, which checks `disposedRef.current` and `viewportRef.current` before acting — so this is already well-defended against the spec-allowed case where `ResizeObserver` fires one last queued notification after `disconnect()`. The guard works, but the *reason* it exists is non-obvious to someone reading the code cold.

Mostly a readability note — a one-line comment on `disposedRef` (something like "guards against late-firing observer callbacks after cleanup") would save future maintainers from wondering why a plain `useRef(false)` boolean is threaded through the microtask.

packages/nimbus/src/components/chat-message/chat-message.context.ts
```diff
@@ -0,0 +1,21 @@
...
+  sender: ChatMessageSender;
+}
+
+/**
+ * Carries the message's `sender` from `ChatMessage.Root` to its parts, so a part
+ * can derive a `sender`-dependent default that CSS alone can't express (e.g. the
+ * avatar's `variant`). Defaults to `"agent"` to match the recipe when a part is
+ * rendered without a `Root` (it never should be in practice).
+ */
+export const ChatMessageContext = createContext<ChatMessageContextValue>({
```
ByronDWall, line 18: **Suggestion: document the `"agent"` default's role as a safety net**

The `createContext` default of `{ sender: "agent" }` means that if `ChatMessage.Avatar` (or another sender-aware sub-component) is accidentally rendered outside a `ChatMessage.Root`, it silently falls back to agent styling rather than crashing. That's good defensive design. The JSDoc here explains it, but the avatar's JSDoc (`chat-message.avatar.tsx:9–30`) doesn't mention that its sender-dependent defaults inherit from this fallback — a reader looking only at the avatar might not realize the behavior is intentional.

Minor — consider a one-liner in the avatar JSDoc like "Reads `sender` from `ChatMessage.Root` context (defaults to `"agent"` if no provider is present)."

ByronDWall (commented):
Five non-blocking suggestions from a review pass — nothing here needs to block merge.

packages/nimbus/src/components/chat-message/chat-message.docs.spec.tsx
```diff
@@ -0,0 +1,231 @@
...
+          </ChatMessage.Avatar>
+          <ChatMessage.Body>
+            <Text>Here is the summary you asked for.</Text>
+          </ChatMessage.Body>
+        </ChatMessage.Root>
+      </NimbusProvider>
+    );
+
+    const body = screen.getByTestId("body");
+    expect(body.tagName.toLowerCase()).toBe("article");
```
ByronDWall: **Suggestion: `.docs.spec.tsx` tests may be asserting internal behavior that belongs in stories**

Per the testing strategy guide, `.docs.spec.tsx` files are consumer-facing examples that get injected into `.dev.mdx` docs — they should demonstrate real-app integration patterns (state management, async data, form wiring) rather than internal structural assertions. Several tests in this file assert tag names (`tagName.toLowerCase() === "article"`), ARIA attributes (`aria-busy`), `role="log"`, and `aria-hidden` avatar suppression — all of which are already covered (and better suited) in the `.stories.tsx` play functions.

Consider replacing these with genuine consumer-integration examples — e.g. rendering a transcript fed by `useState`/an async fetch, or wiring `ChatMessage.Actions` buttons into a mutation call. The ARIA and structural coverage in stories is already thorough.

packages/nimbus/src/components/avatar/avatar.recipe.ts
```diff
@@ -12,30 +12,48 @@ export const avatarRecipe = defineRecipe({
...
+        color: "colorPalette.11",
+      },
+      solid: {
+        backgroundColor: "colorPalette.9",
+        color: "colorPalette.contrast",
+      },
     },
   },
   defaultVariants: {
     size: "md",
```
ByronDWall, line 56: **Suggestion: the `colorPalette` default removal may warrant a `major` bump**

Removing `colorPalette: "primary"` from `base` means every existing `<Avatar>` that doesn't pass `colorPalette` explicitly will silently shift from the primary tint to neutral on upgrade — with no compile-time signal. The changeset documents this well (including the `colorPalette="primary"` escape hatch), but classifying it as `minor` may understate the visual break for consumers who'll see their avatars change color without touching any code.

Worth considering whether this should be `major`, or at minimum flagged prominently in the upgrade guide since it's not TypeScript-detectable.

packages/nimbus/src/components/chat-message-list/hooks/use-stick-to-bottom.ts
misama-ct: Good call-out, and agreed it's non-blocking. Sticking with the scroll-listener approach for now: the numeric `distanceFromBottom` it produces is load-bearing beyond the at-bottom boolean — the 32px near-bottom threshold *and* the "release the pin if the user scrolls up during a smooth jump" guard both need the actual distance, which an `IntersectionObserver` boolean wouldn't give without extra sentinels + `rootMargin`. The `{ passive: true }` reads are cheap for realistic transcripts (and there's no virtualization here, so DOM node count would bite well before scroll-handler cost). Noted as a future option if scroll perf ever surfaces. 👍

packages/nimbus/src/components/chat-message-list/hooks/use-stick-to-bottom.ts
misama-ct: Done in 196260e5b — added a comment: `disposedRef` guards against the late `ResizeObserver` notification the spec allows to fire once after `disconnect()`, which would otherwise run the microtask stick against a torn-down list.

packages/nimbus/src/components/chat-message/chat-message.context.ts
misama-ct: There's a doc comment right above the `createContext` default already, but it didn't spell out the safety-net framing — sharpened it in 196260e5b to say a sender-aware part accidentally rendered outside a `Root` falls back to agent styling rather than throwing.

packages/nimbus/src/components/chat-message/chat-message.docs.spec.tsx
misama-ct: Agreed — per the testing-strategy guide those structural/ARIA assertions belong in the story play functions, which already cover them. In 196260e5b I trimmed the story-owned assertions (`tagName` / `as` / `aria-busy` / `role="log"` / decorative-avatar) so the injected examples read as consumer usage, kept the composition examples, and expanded the action example to wire two buttons (Approve/Dismiss) to their own handlers. A fuller state/async-driven integration example would be a nice follow-up but felt beyond this PR's scope.

packages/nimbus/src/components/avatar/avatar.recipe.ts
misama-ct: Deliberately keeping this at `minor`. The removal aligns `Avatar` with `Button` — both now start neutral and take an explicit `colorPalette` — so it's an intentional design convergence, not a defect. Nothing functionally breaks: existing avatars render a different (not broken) tint, and the changeset documents the one-line restore (`colorPalette="primary"`). A documented, trivially-reversible visual shift with no behavioral or API change doesn't rise to a major under our conventions. The changeset wording already calls the visual change out prominently.

ByronDWall (approved):
Other than the scroll perf suggestions, everything here is looking very good. 

I think `tone=error` is fine. Having system notices, etc be consumer-generated makes sense to me, and we can revisit this if consumers really all end up following the same pattern and need the component from the design system.

packages/nimbus/src/components/chat-message-list/hooks/use-stick-to-bottom.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  const scrollToBottom = useCallback(
+    (behavior: ScrollBehavior = "auto") => {
+      const el = viewportRef.current;
+      if (!el) return;
+      const effective = prefersReducedMotion() ? "auto" : behavior;
+      // A smooth scroll animates over several frames; suppress pin-release for
+      // its duration so the jump-to-latest control doesn't flicker back in.
+      programmaticScrollRef.current = true;
+      lastProgrammaticDistanceRef.current = Number.POSITIVE_INFINITY;
```
tylermorrisford: **`scrollToBottom()` can leave `isPinned` permanently stuck when the list is already at the bottom.**

Setting `lastProgrammaticDistanceRef.current = Number.POSITIVE_INFINITY` before calling `el.scrollTo(...)` is safe when the scroll actually moves — a `scroll` event fires and clears the guard. But when the list is already at the bottom, `scrollTo` is a no-op and no event fires, so `programmaticScrollRef.current` stays `true` indefinitely.

On the next user scroll-up, the guard check evaluates:
```ts
distanceFromBottom > Number.POSITIVE_INFINITY + 4  // always false in JS
```
`userScrolledUp` is never `true`, `isPinned` never flips to `false`, and the "jump to latest" button never appears. The state self-heals only when the next append fires a real scroll, but a consumer calling `scrollToBottom()` defensively after each append (a natural pattern) can keep the list stuck indefinitely.

Fix: snapshot the real distance at call time rather than using `Infinity`:
```ts
lastProgrammaticDistanceRef.current =
  el.scrollHeight - el.scrollTop - el.clientHeight;
```

packages/nimbus/src/components/chat-message/components/chat-message.typing.tsx
```diff
@@ -0,0 +1,33 @@
...
+ * @supportsStyleProps
+ */
+export const ChatMessageTyping = ({
+  ref,
+  children,
+  ...props
+}: ChatMessageTypingProps) => {
+  return (
+    <ChatMessageTypingSlot ref={ref} {...props}>
+      <ActivityIndicator />
```
tylermorrisford, line 27: **No accessible announcement when the typing indicator first appears.**

`ActivityIndicator` is always `aria-hidden`, so when `ChatMessage.Typing` is mounted (with no `children`) it appends an `<article>` to the `role="log" aria-live="polite"` region whose entire subtree is invisible to AT. Screen readers receive no signal that the agent has started generating a reply.

`aria-busy="true"` on `ChatMessage.Root` (a descendant of the live region, not on the region itself) has inconsistent AT support and is not a reliable substitute. The JSDoc already guides consumers to "pass a visible label as `children`", but there is no default fallback and no enforcement.

A minimal fix would be a visually-hidden default string (e.g. "Agent is typing…") rendered alongside `ActivityIndicator` that consumers can override via `children` or suppress with an explicit `null`. Could also be addressed in a follow-up if the current approach is intentional.

packages/nimbus/src/components/avatar/avatar.tsx
```diff
@@ -69,6 +76,11 @@ export const Avatar = (props: AvatarProps) => {
   // while the image is loading, or when the image failed to load.
   const shouldShowFallback = !src || !imageLoaded || imageError;
 
+  // Custom content (e.g. an icon) overrides the default rendering entirely.
+  if (children != null) {
```
tylermorrisford, line 81: **`children` silently discards `src` with no warning or JSDoc mention.**

When both `children` and `src` are provided the early-return here means `src` is never used — `onLoad`/`onError` are never attached, and the image never renders. TypeScript accepts the combination without complaint because both props are optional. A consumer who intends `children` as a fallback icon while a real photo loads would find the photo silently ignored.

Documentation-only fix: the `children` JSDoc should explicitly state that it takes unconditional precedence over `src` and that `src` is ignored when `children` is present.

tylermorrisford (approved):
One smallish find and two other notes - approving! So lovely

packages/nimbus/src/components/chat-message-list/hooks/use-stick-to-bottom.ts
misama-ct: Dear Tyler, fixed in 0727e39. Seeded `lastProgrammaticDistanceRef` with the real pre-scroll distance (`scrollHeight - scrollTop - clientHeight`) instead of `Infinity`, so an already-at-bottom no-op `scrollToBottom()` no longer wedges the pin — the next genuine scroll-up releases it. Added a regression story (`RedundantScrollToBottomStillReleases`) that fails on the old `Infinity` seed.

packages/nimbus/src/components/chat-message/components/chat-message.typing.tsx
misama-ct: Dear Tyler, keeping the current approach for now. The transcript's live-region announcement is the container's responsibility (`ChatMessageList.Root`'s `role="log"` / `aria-live="polite"`), and a bare `Typing` has no non-hidden text to announce, so the documented path is passing a `children` label. An accessible-by-default label is worth doing, but it pulls i18n scaffolding into `chat-message` and interacts with `aria-busy` (which can defer announcement of the hidden text while streaming), so it's better as a dedicated follow-up than as scope on this PR.

packages/nimbus/src/components/avatar/avatar.tsx
misama-ct: Dear Tyler, documented in 0727e39. Clarified the `children` JSDoc — it takes unconditional precedence over `src` (the image never renders and `onLoad`/`onError` never fire when `children` is present) — and added the same note to the inline comment.

valoriecarli (approved):
Looking slick! Note that the `/components` landing page is missing cards for `Chat` & `Context` due to their missing corresponding `.mdx` files. Almost pushed them, but you prob wouldn't like my icon & order choices. 

jsyk - I've also noted the new Avatar snapshot that will need to be included... and your favorite... LOC ;)

misama-ct:
> [...] missing cards for Chat & Context due to their missing corresponding .mdx files

Only read it after merge 😬 ,  I'll do that in a separate PR soon.

PR #1730, opened 2026-07-08
refactor(nimbus): two-lane imports + ban value `export *` in leaf barrels (fix prod-Storybook CodeRoot crash)

misama-ct:
> [!NOTE]
> **Two-lane import convention — the rule to follow going forward:**
>
> - **Implementation files** (the library's own component / hook / pattern source): import other Nimbus code via the **`@/` alias** — barrel (`@/components/button`) or deep (`@/components/button/button`), both safe.
> - **Stories, tests, docs** (`*.stories.tsx`, `*.spec.tsx`, `*.docs.spec.tsx`): import Nimbus components from the **package** `@commercetools/nimbus`.
>
> ESLint-enforced. Made safe by naming **leaf barrels** — which is also what fixes the prod-Storybook `ReferenceError: CodeRoot` crash; rollup barrels keep `export *`. Repo-internal only — no public-API, consumer, or bundle change.

## Summary

Decouples Nimbus's internal import notation from the build target and makes the `ReferenceError: CodeRoot is not defined` prod-Storybook / Chromatic crash class **structurally impossible**. Repo-internal only — no change to the published package, public API, or consumer behavior.

## Problem & root cause

The production Storybook build crashed at module-eval with `ReferenceError: CodeRoot is not defined` (then `DefaultPageRoot`, …), aborting Chromatic story extraction. Root cause (proven via controlled `build-storybook` runs):

The dev/prod resolution split let a _source_ barrel (`@/…`→src) coexist with the _dist_ barrel (`@commercetools/nimbus`→dist) in one production bundle. vite/rolldown's `lazyBarrel` then mis-shook the source barrel's `export *` **chains** for split-module compound components (`export const Code = CodeRoot`, where `CodeRoot` lives in a separate module) — dropping the declaring module while keeping the consumer, leaving a free variable.

Two independent conditions each remove the bug: **no source barrel in the bundle**, or **no value `export *` in the chain**. This PR takes the second route — the `@/`→src source barrel stays present, but the `export *` chain is broken.

## What changed

- **No value `export *` in leaf barrels.** Every leaf barrel — the per-component/hook/pattern `index.ts` that sits directly on implementation files (~158 of them) — names its value re-exports explicitly (`export { Button } from "./button"`); `export type *` is kept (erased at build, can't trigger the defect). Naming the leaf barrel breaks the `export *` chain at the point where a split-module compound root first enters the graph — which is all the fix requires. The **root + category rollup barrels** (`src/index.ts`, `components/index.ts`, `hooks/index.ts`, `patterns/index.ts`, `constants/index.ts`) keep `export *`, because a star over an already-named barrel is safe — so new components/hooks are picked up automatically. Enforced by ESLint (`no-restricted-syntax` on `ExportAllDeclaration`), with the rollup barrels exempt.
- **Two import lanes**, ESLint-enforced:
  - Consumer files (`*.stories`, `*.spec`, `*.docs.spec`) → import Nimbus components from the package `@commercetools/nimbus`.
  - Implementation files → import via the `@/` alias (barrel **or** deep — both safe now that leaf barrels are named).
- **Single resolution switch:** `@/`→src is invariant in every command; `@commercetools/nimbus` is the only specifier that flips src↔dist, decided per-command. Removed the `resolve.tsconfigPaths` leak and the `.storybook/main.ts` regex override that only existed to counter it.
- **Docs & tooling sync:** inverted the old "import implementation files, not barrels" guidance to the two-lane model; `docs/file-type-guidelines/barrel-exports.md` is the canonical source.

## `lazyBarrel`-agnostic — no override

The `.storybook/main.ts` `lazyBarrel: false` override (from #1704) is **removed**; the build inherits the flag's default. It stays clean with the flag at its **default (ON)** because named leaf barrels give the bundler a static name→module map. Verified locally at **both** `lazyBarrel` ON and OFF (see Testing). Going forward this class is guarded by the ESLint rule (which prevents the `export *` trigger) and Chromatic's built-Storybook capture — the check that surfaced the original crash.

## Testing

Verified locally on this branch:

- **Built Storybook loads all 1040 stories with 0 module-eval errors** — checked at `lazyBarrel` **default-ON**, explicit **OFF**, and the committed config (override removed).
- `pnpm lint` (incl. the scoped `no value export *` rule) — clean.
- `typecheck:dev` (source) + `typecheck` (built dist) — clean.
- `pnpm --filter @commercetools/nimbus build` — success.

Confirmed by CI (re-run before flipping out of draft): `test:storybook` + `test:storybook:dev` full suites, `check:package-shape`, `check:bundle-size`.

## Consumer impact

**None.** The published dist is a single pre-bundled chunk (never affected by the source-barrel mis-shake); the public type surface and bundle sizes are expected unchanged — confirm via `check:package-shape` / `check:bundle-size` in CI.

## Risk

Low — repo-internal, no published-artifact change. Rollback = revert the PR.

## Related

- Supersedes #1696 (same intent; regenerated under the new lint rules for a complete, drift-proof migration).
- Removes #1704's `lazyBarrel: false` override.
- Housekeeping follow-up to #1729 (pnpm 11 upgrade).


misama-ct:
## Superseded — the crash is an already-fixed upstream rolldown bug

**TL;DR:** the `ReferenceError: CodeRoot is not defined` prod-Storybook crash this PR restructures the barrels to avoid is a **rolldown tree-shaking bug that is already fixed in the rolldown version we ship**. The structural approach here (named leaf barrels + two-lane import rule) isn't needed.

### Root cause
Under `sideEffects: ["*.css"]`, rolldown's `lazyBarrel` optimization dropped the module that *defines* `CodeRoot` while keeping `export const Code = CodeRoot` — violating the invariant "keep the use → keep the definition." Tracked upstream as three faces of one bug: rolldown [#9806](https://github.com/rolldown/rolldown/issues/9806) / [#9964](https://github.com/rolldown/rolldown/issues/9964) / [#9961](https://github.com/rolldown/rolldown/issues/9961).

### The fix already shipped
rolldown [PR #10080](https://github.com/rolldown/rolldown/pull/10080) fixed all three in **rolldown 1.1.5** (2026-07-08). This repo already resolves `rolldown@1.1.5` (via `vite@8.1.5`).

### Verified
Rebuilt the prod Storybook from `main`'s `export *` barrels with `lazyBarrel` at its default (on) on rolldown 1.1.5, then loaded the former crash victims (`DefaultPage`, `Drawer`, `Dialog`) in a real browser → **0 `ReferenceError`s**, all rendered.

### Replacement
The only change actually required is removing the now-obsolete `lazyBarrel: false` override added in #1704 — a one-line diff on `main`, on branch `fix/remove-obsolete-lazybarrel-override`. Recommend closing this PR in favor of that.

misama-ct:
Replacement PR opened: #1795 — the one-line removal of the obsolete `lazyBarrel: false` override (the actual fix, now that rolldown 1.1.5 ships the upstream patch).

PR #1740, opened 2026-07-09
feat(breadcrumbs): add Breadcrumbs component

misama-ct:
## What

Adds a **Breadcrumbs** component — a navigation control that shows the hierarchical path to the current page as an ordered list of links.

image

- `Breadcrumbs.Root` — renders a `<nav>` landmark wrapping an `<ol>`; takes an `aria-label` and an optional `separator` (any node, default `›`).
- `Breadcrumbs.Item` — renders an `<li>` containing a React Aria `Link`. The **last** item is automatically treated as the current page (non-interactive text with `aria-current="page"`, removed from the tab order) — currentness is derived from position, so there is no `isCurrent` prop. Also supports `isDisabled` (renders as a non-interactive `<span>`), `target`, `rel`, and `routerOptions`.

Two authoring styles share the same separator context: compound `Breadcrumbs.Item` children (static trails) or a declarative `items` array on `Breadcrumbs.Root` with an `onAction` handler (data-driven trails).

```tsx
<Breadcrumbs.Root aria-label="Breadcrumb">
  <Breadcrumbs.Item href="/">Home</Breadcrumbs.Item>
  <Breadcrumbs.Item href="/orders">Orders</Breadcrumbs.Item>
  <Breadcrumbs.Item>Order #123</Breadcrumbs.Item>
</Breadcrumbs.Root>
```

## Why

Breadcrumb was a feature-parity gap vs. every major component library (shadcn, MUI, Ant, Carbon, Polaris) and has no prior equivalent in Nimbus. It's a Tier-1 item on our component parity backlog.

## Provenance

The implementation originated on the exploratory branch `exp/breadcrumbs-after-docs` (a docs/tooling A/B experiment). This PR cherry-picks the single, self-contained component commit onto current `main` — no other experiment changes are included — and adds a changeset + a clean commit message. It was subsequently rebuilt on React Aria Components primitives (see the `refactor: rebuild on React Aria Components primitives` commit): current-page detection moved from an `isCurrent` prop to position-derived, the default separator became `›`, and disabled items now render as a `<span>` rather than a neutralized `Link`.

## Included

Full component per Nimbus conventions: main (`breadcrumbs.tsx`), `Root`/`Item`/context, `types`, `recipe` (registered as the `nimbusBreadcrumbs` slot recipe), `slots`, stories with play functions, docs (`.dev.mdx`, `.a11y.mdx`, `.mdx`), and a consumer docs spec. Changeset: `minor`.

## Verification

- `typecheck:dev` (source surface): clean.
- `pnpm test:dev packages/nimbus/src/components/breadcrumbs/`: **18/18 pass** (7 story + 11 docs-spec).
- Regenerated Chakra theme typings (`build-theme-typings`) so the new `nimbusBreadcrumbs` recipe is in the generated type map.

> CI's `typecheck:strict` + `test:storybook` run against the built `dist` surface and remain the merge gate.

## Known / minor

- **Disabled last item:** disabling the _last_ item is not recommended. It is already the non-interactive current page, and because a disabled item renders as a plain `<span>` outside React Aria's collection context, a disabled final item will not carry `aria-current="page"`. Documented on the `isDisabled` prop and in the a11y docs.

## Follow-ups

- Reviewer should confirm the recipe visuals against design (separator spacing, current-page treatment, disabled dimming).

### DIFF
--- a/.changeset/add-breadcrumbs-component.md
+++ b/.changeset/add-breadcrumbs-component.md
@@ -3,20 +3,35 @@
 ---
 
 **Breadcrumbs**: new component for showing the hierarchical path to the current
-page as an ordered list of links.
+page as an ordered list of links. Built on React Aria Components'
+`Breadcrumbs`/`Breadcrumb`.
 
 - `Breadcrumbs.Root` renders a `<nav>` landmark wrapping an ordered list; give
   it an `aria-label` (e.g. `"Breadcrumb"`) and an optional `separator` (any
-  node, defaults to `/`).
-- `Breadcrumbs.Item` renders a link. Set `isCurrent` on the last item to mark
-  the current page — it renders as non-interactive text with
-  `aria-current="page"` and is removed from the tab order. Items also support
-  `isDisabled`, `target`, and `rel`.
+  node, defaults to `›`). It accepts either compound `Breadcrumbs.Item` children
+  or a declarative `items` array, and an `onAction(id)` handler for client-side
+  routing.
+- The **last** item is the current page automatically — it renders as
+  non-interactive text with `aria-current="page"` and is removed from the tab
+  order. There is no `isCurrent` prop.
+- `Breadcrumbs.Item` renders a link and supports `href`, `target`, `rel`,
+  `routerOptions`, and `isDisabled`. `size` (`sm`/`md`) is set on the root.
 
 ```tsx
+// Compound API
 <Breadcrumbs.Root aria-label="Breadcrumb">
   <Breadcrumbs.Item href="/">Home</Breadcrumbs.Item>
   <Breadcrumbs.Item href="/orders">Orders</Breadcrumbs.Item>
-  <Breadcrumbs.Item isCurrent>Order #123</Breadcrumbs.Item>
+  <Breadcrumbs.Item>Order #123</Breadcrumbs.Item>
 </Breadcrumbs.Root>
+
+// Data-driven API — the last entry is the current page
+<Breadcrumbs.Root
+  aria-label="Breadcrumb"
+  items={[
+    { id: "home", label: "Home", href: "/" },
+    { id: "orders", label: "Orders", href: "/orders" },
+    { id: "detail", label: "Order #123" },
+  ]}
+/>
 ```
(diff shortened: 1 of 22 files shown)


misama-ct:
@claude what do you think? is this ready for production?

misama-ct:
@claude thanks — worked through all three, verified each against the code (`6bcf664`).

**#1 — PR description stale.** ✅ Rewritten to match the current API (default `›`, position-derived current item, no `isCurrent`).

But the sub-claim that the empty-`href` warning "no longer fires" is **not correct** — and it was mis-attributed. It's not the disabled items (they render clean `<span>`s). It's the **current/last item with no `href`**: passing `href={undefined}` to `RaLink` still routes through React Aria's `useHref`/`RouterProvider` path, which materializes `href=""` on the element in the browser and trips React DOM's dev warning. It fired ~10× in the Storybook browser run (the CI gate); the JSDOM `.spec` tests never surfaced it, which is why it looked resolved. Fixed by omitting the prop when falsy (`{...(href ? { href } : {})}`): 10 → 0 warnings, all 21 tests still pass, current item still renders as `<span role="link" aria-current="page">`.

**#2 — disabled span drops `restProps`/`ref`.** ✅ Partly. Now spreading `{...restProps}` onto the `<span>` so `data-*`/`className`/`aria-*`/handlers reach disabled items. I did **not** attach `ref`: it's typed `React.Ref<HTMLAnchorElement>`, but the disabled element is a `<span>` — forwarding an anchor-typed ref to a span would be a type/element mismatch.

**#3 — disabled last item loses `aria-current`.** ✅ Confirmed empirically (a disabled final item renders without `aria-current="page"`, since the `<span>` bypasses RA's collection context). Restoring it isn't a trivial guard — it'd mean re-plumbing the context the span deliberately skips — and it only matters for near-nonsensical usage (disabling the current page, which RA already renders non-interactive). Went with a documented note on the `isDisabled` prop + in the a11y docs rather than code.

Still outstanding (unchanged from your list): a green built-surface CI run and design sign-off on the recipe visuals.

valoriecarli (approved):
so niiiiice! 🙌🏻

ByronDWall (approved):

PR #1745, opened 2026-07-13
feat(slider): add Slider & RangeSlider with filled, minimal & enclosed variants

misama-ct:
## Summary

Adds **Slider** and **RangeSlider** (React Aria + Chakra slot recipe) with three
visual variants and full Storybook / interaction-test coverage.

- **`filled`** (default) — thin track with a colored progress fill and a ring knob.
- **`minimal`** — understated at rest, alive when engaged. A thin neutral track
  carries a single, solid-color handle (`colorPalette.9` → `.10` on
  hover/focus/drag); the track — and, on a `RangeSlider`, the colored range
  segment — grow from a thin line to the full track thickness on engage. A single
  `Slider` shows no progress fill. For low-emphasis controls such as a
  thumbnail-size picker.
- **`enclosed`** — iOS-style thick contained bar; the knob sits inside it.

All variants come in two sizes, either orientation, and with optional tick marks,
and work inside `FormField` or standalone.

---

### Storybook

- [Slider](https://nimbus-storybook-git-feat-slider-component-commercetools.vercel.app/?path=/docs/components-slider--docs)
- [RangeSlider](https://nimbus-storybook-git-feat-slider-component-commercetools.vercel.app/?path=/docs/components-rangeslider--docs)

### Docs

- [Slider](https://nimbus-documentation-git-feat-slider-component-commercetools.vercel.app/components/inputs/slider)
- [RangeSlider](https://nimbus-documentation-git-feat-slider-component-commercetools.vercel.app/components/inputs/range-slider)

---

image

### One shared geometry

All three variants share **one geometry**: the interactive track is inset by the
thumb radius and the visible bar is painted full-width via `::before`, so React
Aria's own pointer/thumb math produces the contained layout. That single model
gives every variant, for free:

- thumbs stay contained at the extremes (no half-off overhang at 0/100),
- ticks align to the thumb centerline with clear rounded caps,
- clicking a tick lands the thumb exactly on it,
- clicking past the first/last tick clamps to min/max (no dead zone),
- the fill "cups" each thumb's outer edge.

`minimal` keeps this geometry untouched — only its painted layers (the track's
`::before` bar, the range's `::before` bar, and the knob) resize and recolor on
engage, so the DOM boxes, hit-targets and thumb-size grid stay identical at rest
and engaged. `enclosed` is just "base geometry with a thumb-tall track" plus a
transparent-halo thumb.

### Sizing

Thumb sizes follow the **Switch** control-size grid.


misama-ct:
@claude please review

misama-ct:
Dear @claude,

Thanks for the thorough review. Here's what we addressed, all pushed to `feat/slider-component`:

**🔴 #1 — Stale changeset.** Fixed. `.changeset/slider-visual-variants.md` now describes exactly the two shipped variants (`plain` default + `enclosed`) and drops the three removed ones (`solid`/`outline`/`minimal`); the thumb-centering note stays. (in `137e7b071`)

**🟡 #2 — Escape-dismissible tooltip (WCAG 1.4.13).** Fixed — and it turned out to matter more than the base finding. Because the slider controls the tooltip's open state (so it stays visible during a drag, which React Aria closes on press), React Aria's internal tooltip state — and the `stopPropagation` Escape shield that comes with it — desyncs from our controlled state on any value change. That produced a nasty bug: after an arrow-key value change, a single Escape closed **both** the tooltip and an enclosing Dialog at once. Root cause was confirmed in a real browser (React Aria fires `onOpenChange → false` on the arrow key, unregistering its Escape shield). Fix: the slider now owns the tooltip's Escape entirely — a capture-phase listener, active only while the tooltip is visible, dismisses it and stops propagation, so the first Escape closes just the tooltip and the next reaches the Dialog. (in `137e7b071`)

Heads-up: this desync only reproduces under real browser key events, not Storybook's synthetic `userEvent`, so CI can't regression-guard that exact path — it's verified manually and documented in the story. Added `EscapeDismissesTooltip` + `SliderInDialog` regression stories for the reproducible layering.

**🟡 #3 — No tick on `maxValue` for non-divisor `tickStep`.** Fixed by appending a `maxValue` tick (e.g. `0/30/60/90/100`), deduped when `maxValue` already lands on the last multiple. The final interval is then shorter than `tickStep`; documented on the `tickStep` prop and guarded with a `Ticks` assertion. (in `67c43590c`)

The built-surface gate (`pnpm build` + `typecheck:strict` + `test:storybook` against `dist`) is still pending as you noted — our verification was source-level tests + real-browser Playwright. And the Transifex/localization Jira ticket for the `.i18n.ts` change is on the list.

Happy now?

ByronDWall (approved):

PR #1746, opened 2026-07-13
feat(data-table): column align prop, default cell color, sticky backgrounds

ByronDWall:
## Summary

- Added `align` prop (`'start' | 'center' | 'end' | 'stretch'`) to `DataTableColumnItem<T>` for column-level cell alignment — applied to both body cells and column headers
- Changed default body-cell text color from `neutral.11` to `neutral.12` for improved contrast
- Applied opaque backgrounds to sticky header columns (`colorPalette.2`) and sticky body cells (`bg`) so scrolled content no longer shows through
- Fixed Firefox/Safari 1-2px gap at sticky cell edges caused by `border-collapse:collapse` rendering — uses a `::before` pseudo-element with `background: inherit` to extend coverage by 2px on each side

Closes FEC-883

## Test plan
- [ ] Verify `align="end"` right-aligns both column headers and body cells in Storybook `ColumnAlignment` story
- [ ] Verify `align="center"` centers both header and body cell text
- [ ] Verify `align="stretch"` makes ProgressBar fill the full cell width (header defaults to `end` alignment)
- [ ] Verify body cells render with `neutral.12` text color by default
- [ ] Verify sticky header and body cells have opaque backgrounds when scrolling horizontally
- [ ] Verify no 1-2px gap at sticky cell edges in Firefox and Safari
- [ ] Verify hover/selected row states still work correctly on sticky cells (pseudo-element inherits background)
- [ ] Verify stretch cells still support truncation (overflow is not `visible`)
- [ ] Check both light and dark mode for background consistency

### DIFF
--- a/packages/nimbus/src/components/data-table/data-table.stories.tsx
+++ b/packages/nimbus/src/components/data-table/data-table.stories.tsx
@@ -44,6 +44,8 @@ import {
   initialVisibleColumns,
   managerRows,
   initialHiddenColumns,
+  alignDemoColumns,
+  alignDemoRows,
 } from "./data-table.test-data";
 
 import { useDragAndDrop, createArrayHandlers } from "@commercetools/nimbus";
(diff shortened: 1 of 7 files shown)


misama-ct:
> Verify sticky header and body cells have opaque backgrounds when scrolling horizontally

Screenshot 2026-07-14 at 15 56 59

Not sure what causes it, but left and right there are 1px or 2px still uncovered. Maybe have a look if it's an easy fix.

misama-ct (approved):

ByronDWall:
@misama-ct good catch, should be fixed in [9ef1a04](https://github.com/commercetools/nimbus/pull/1746/commits/9ef1a0413995a98dc7f7f1708e822a65e856c633)

tylermorrisford (commented):
Quick secondary review - approving anyway 😎

packages/nimbus/src/components/data-table/data-table.stories.tsx
```diff
@@ -4975,3 +4977,179 @@ export const DragAndDropRows: Story = {
...
+      const stretchCells = canvasElement.querySelectorAll("[data-truncated]");
+      const stretchBlock = Array.from(stretchCells).find((el) => {
+        const s = window.getComputedStyle(el as HTMLElement);
+        return s.display === "block";
+      }) as HTMLElement;
+      expect(stretchBlock).toBeTruthy();
+      expect(
+        stretchBlock.style.overflow ||
+          window.getComputedStyle(stretchBlock).overflow
+      ).not.toBe("visible");
```
tylermorrisford, line 5060: The stretch `Box` has no explicit `overflow` set, so `getComputedStyle(stretchBlock).overflow` will return `"visible"` — the browser default for block elements. This assertion (`.not.toBe("visible")`) would then fail. If the intent is to confirm the stretch wrapper does not clip content via `overflow: hidden`, the assertion should be `.not.toBe("hidden")`.

tylermorrisford (approved):

PR #1747, opened 2026-07-13
fix(mcp): correct migrate_from_uikit mappings against live Nimbus types

ByronDWall:
## Summary

**Structured prop mappings** for the `migrate_from_uikit` MCP tool, validated at build time against both Nimbus and UIKit live types.

### What changed
- **`PropMapping` type**: Structured UIKit→Nimbus prop translations with `changeType` (`rename`, `value-mapping`, `structural`, `removed`), value mapping tables, and fixed values
- **59 entries populated** with `propMappings` (out of 80 total — remaining 21 have no prop-level changes)
- **Build-time Nimbus validation**: `validate-migration-data.ts` prebuild step checks every `nimbusProp` exists on the target component and every mapped value is valid against the live Nimbus type data
- **Build-time UIKit validation**: Uses TypeScript compiler API to resolve UIKit component types from `@commercetools-frontend/ui-kit` barrel and validates every `uiKitProp` exists on the source component — handles function components, class components, polymorphic/generic components (intersection types), and dotted sub-components (Text.Body, Spacings.Inline, etc.)
- **57 of 59 UIKit entries validated** — only `AdditionalInfoMessage` and `ThemeProvider` are skipped (not exported from UIKit barrel)
- **Shared validation helpers**: `validation-helpers.ts` eliminates duplication between build script and test spec
- **Compound component resolution**: `resolveTypeFile` prefers `Root.json` when it has props (fixes silent validation skip for Alert, Select, Card, TagGroup, Tooltip, Accordion) and strips dots so sub-component names like `Alert.Actions` resolve to `AlertActions.json`
- **Data fixes**: ContentNotification tone→colorPalette, Badge tone→colorPalette, Card.Content→Card.Body, SearchSelectInput.options→loadOptions, LocalizedMoneyInput.selectedLanguage→selectedCurrency, TagList.items→children, DataTable.onSelectionChange→onRowClick, DropdownMenu.options/onSelect→children, NumberField errors→errors (not errorMessage), FlatButton inverted tone note, IconButton theme changeType fix
- **Tool output**: `propMappings` flows through to MCP results, giving LLM consumers structured data
- **Invariant enforcement**: `value-mapping` changeType requires `valueMapping` or `fixedValue`

Closes FEC-884

## Test plan
- [x] 59 Nimbus entries validated against live type data (0 errors)
- [x] 57 UIKit components validated against live types (0 errors)
- [x] All 28 migration tests pass
- [x] Validation prebuild step passes standalone
- [x] Compound components (Alert, Select, Card, TagGroup, Tooltip, Accordion) resolve to Root.json and are validated
- [x] Dotted sub-components (Text.Body, Spacings.Inline, etc.) resolved and validated
- [x] CI green


packages/nimbus-mcp/scripts/validate-migration-data.ts
```diff
@@ -0,0 +1,300 @@
...
+
+interface ValidationError {
+  entry: string;
+  prop: string;
+  message: string;
+}
+
+function resolveTypeFile(componentName: string): string | null {
+  const direct = resolve(TYPES_DIR, `${componentName}.json`);
+  if (existsSync(direct)) return direct;
```
tylermorrisford: **Blocker: `resolveTypeFile` skips validation for all compound components**

This early return fires whenever `<Name>.json` exists in the types directory — but compound components (Alert, Select, Card, TagGroup, Tooltip, Accordion) emit an **empty** `<Name>.json` aggregator alongside a populated `<Name>Root.json`. Because the direct file exists first, the `Root.json` fallback on line 37 never fires, `loadTypeData` returns `{}`, and `validateMigrationData` hits the "no type data found, skipping" path for all six of these components.

Verified against `data/docs/types/`:
- `Alert.json` → 0 props, `AlertRoot.json` → 2 props
- `Select.json` → 0 props, `SelectRoot.json` → 45 props
- `Card.json` → 0 props, `CardRoot.json` → 4 props
- `TagGroup.json` → 0 props, `TagGroupRoot.json` → 14 props
- `Tooltip.json` → 0 props, `TooltipRoot.json` → 8 props
- `Accordion.json` → 0 props, `AccordionRoot.json` → 8 props

Suggested fix — fall through to `Root.json` when the direct file has no props:
```ts
function resolveTypeFile(componentName: string): string | null {
  const direct = resolve(TYPES_DIR, `${componentName}.json`);
  const root = resolve(TYPES_DIR, `${componentName}Root.json`);
  if (existsSync(root)) {
    const rootData = JSON.parse(readFileSync(root, 'utf-8'));
    if (Object.keys(rootData.props ?? {}).length > 0) return root;
  }
  if (existsSync(direct)) return direct;
  return null;
}
```
Fixing this would also have caught the `tone`/`colorPalette` mistake in the next comment at build time.

packages/nimbus-mcp/src/data/uikit-migration.ts
```diff
@@ -815,6 +1627,19 @@ const MIGRATION_DATA: UiKitMigrationEntry[] = [
       "Rename to Alert",
       "type prop replaced by tone ('info', 'success', 'warning', 'danger')",
     ],
+    propMappings: [
+      {
+        uiKitProp: "type",
+        nimbusProp: "tone",
```
tylermorrisford: **Blocker: `nimbusProp: "tone"` — Nimbus Alert has no `tone` prop**

The live Nimbus Alert exposes `colorPalette` for semantic intent, not `tone` — confirmed in `alert.types.ts:25`:
```ts
colorPalette?: Exclude<SemanticPalettesOnly, "neutral" | "primary">;
```
`AlertRoot.json` has only `variant` and `ref`.

The `valueMapping` targets are already correct (`info`→`info`, `success`→`positive`, `warning`→`warning`, `danger`→`critical`), so the fix is just:
```ts
nimbusProp: "colorPalette",
```
…plus updating the `notes` (line 1625) and `breakingChanges` (line 1628) prose to say `colorPalette` instead of `tone`.

This slipped through because the `resolveTypeFile` bug above caused Alert's validation to be silently skipped — fixing that first would have caught this at build time.

tylermorrisford (commented):
Two blockers found during review — inline comments below. Both are data-correctness issues; the second is a direct consequence of the first.

packages/nimbus-mcp/scripts/validate-migration-data.ts
ByronDWall: very good catch, fixed in [ef95a63](https://github.com/commercetools/nimbus/pull/1747/commits/ef95a63105aad99ceebde0be66bed5c09b766278)

packages/nimbus-mcp/src/data/uikit-migration.ts
ByronDWall: fixed in [ef95a63](https://github.com/commercetools/nimbus/pull/1747/commits/ef95a63105aad99ceebde0be66bed5c09b766278)

misama-ct (approved):

packages/nimbus-mcp/src/data/uikit-migration.ts
```diff
@@ -119,6 +271,33 @@ const MIGRATION_DATA: UiKitMigrationEntry[] = [
...
+      {
+        uiKitProp: "color",
+        nimbusProp: "variant",
+        changeType: "value-mapping",
+        valueMapping: [
+          { from: "solid", to: "solid" },
+          { from: "primary", to: "outline" },
+          { from: "info", to: "outline" },
+        ],
+        notes:
```
tylermorrisford, line 284: The `_component` sentinel (line 276) unconditionally injects `variant="outline"`, but the `color` value-mapping below maps `color="solid"` → `variant="solid"`. A consumer migrating a `<SecondaryIconButton color="solid">` sees two conflicting `propMappings` targeting `nimbusProp: "variant"`. The `_component` entry seems wrong for the solid case — either drop it (the outline default is implied by the component name and the notes text) or add a note that it represents the non-solid default.

packages/nimbus-mcp/scripts/validation-helpers.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export function extractNimbusComponentName(
+  nimbusEquivalent: string | null
+): string | null {
+  if (!nimbusEquivalent) return null;
+  const name = nimbusEquivalent.split(/[+,]/)[0].trim().replace(/^<|>$/g, "");
+  if (
+    name === "Design tokens" ||
+    name === "Material Icon Library" ||
+    name === "Text + FormField"
```
tylermorrisford: Dead guard: `name` is derived from `nimbusEquivalent.split(/[+,]/)[0].trim()`, so for an entry whose `nimbusEquivalent` is `"Text + FormField"`, `name` is already `"Text"` at this point — the string `"Text + FormField"` can never match. If any such entry gains `propMappings` in future, validation will silently run against the `Text` component instead of being skipped. Consider testing `nimbusEquivalent` before splitting, or removing the dead branch.

tylermorrisford (commented):
Secondary review — two minor findings below. Happy to defer these to follow-ups.

tylermorrisford (approved):
this is 😎 rull rull 😎  nice

PR #1755, opened 2026-07-14
feat(mcp): add get_docs_page tool for full page content retrieval

ByronDWall:
## Why this matters

Before this PR, `search_docs` returned snippets — short excerpts around the matched term. For component pages, agents could follow the `toolHint` to `get_component` and retrieve structured, section-based content. But for the 57 non-component pages (installation guides, style props, ADRs, design token docs, hooks, contribution guides), there was no follow-up tool. The snippet was all an agent got, and it often lacked the context needed to answer the user's actual question.

`get_docs_page` closes that gap. It's the escape hatch for when a `search_docs` snippet isn't enough: the agent calls `get_docs_page` with the route path and gets the full page content. For tabbed pages, the `section` parameter lets the agent retrieve just the view it needs.

The non-component pages this tool serves are small (median 1.5KB, max 12.5KB) — well within context budget. Component pages (50-85KB) are intentionally routed to `get_component` via `toolHint` instead, where they benefit from structured section-based retrieval.

By making `toolHint` always present (no more `undefined` for non-component pages), every `search_docs` result now tells the agent exactly which tool to call next — completing the discovery-to-detail pipeline across all doc types.

## Summary
- Adds new `get_docs_page` MCP tool that retrieves full documentation page content by route path (from `search_docs` results)
- Supports optional `section` parameter for tabbed pages (e.g. `overview`, `dev`, `guidelines`, `a11y`)
- Returns page metadata (title, description, available sections) alongside content
- Updates `deriveToolHint` in `search_docs` to always return a hint — no more `undefined` toolHint for 55+ non-component pages
- Adds `DocsPageResult` type to `types.ts`
- Extracts shared `routePathToSlug` utility, replacing 3 divergent slug implementations
- Adds WeakMap-based `stripMarkdown` caching for repeated page access
- Case-insensitive section matching and leading/trailing slash normalization

Closes FEC-1128

## Test plan
- [x] `get_docs_page` returns full content for single-page routes (e.g. installation)
- [x] `get_docs_page` returns all views concatenated for tabbed pages when no section specified
- [x] `get_docs_page` returns specific section content when section param provided
- [x] `get_docs_page` returns error with `search_docs` suggestion for unknown routes
- [x] `get_docs_page` returns error for unknown section on tabbed pages
- [x] `search_docs` results always include a `toolHint` (no more `undefined`)
- [x] Non-component/token/icon pages get `get_docs_page` as their toolHint
- [x] Path normalization handles leading/trailing slashes
- [x] Section matching is case-insensitive
- [x] All existing nimbus-mcp tests pass

### DIFF
--- a/packages/nimbus-mcp/src/tools/get-docs-page.ts
+++ b/packages/nimbus-mcp/src/tools/get-docs-page.ts
@@ -3,6 +3,18 @@ import { z } from "zod";
 import { getRouteData } from "../data-loader.js";
 import type { DocsPageResult } from "../types.js";
 import { stripMarkdown } from "../utils/markdown.js";
+import { routePathToSlug } from "../utils/route.js";
+
+const viewContentCache = new WeakMap<object, string>();
+
+function getCachedStripped(viewObj: { mdx: string }): string {
+  let cached = viewContentCache.get(viewObj);
+  if (!cached) {
+    cached = stripMarkdown(viewObj.mdx);
+    viewContentCache.set(viewObj, cached);
+  }
+  return cached;
+}
 
 export function registerGetDocsPage(server: McpServer): void {
   server.registerTool(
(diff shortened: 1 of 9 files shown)


packages/nimbus-mcp/src/tools/get-docs-page.ts
```diff
@@ -0,0 +1,125 @@
...
+            isError: true,
+          };
+        }
+
+        const { meta } = routeData;
+        const availableSections = Object.keys(routeData.views ?? {});
+
+        const normalizedSection = section?.toLowerCase();
+        if (normalizedSection) {
+          const view = routeData.views?.[normalizedSection];
```
tylermorrisford, line 66: nit: this normalises the input but not the data key, so it works only because all current view keys happen to be lowercase. If a future route ships a mixed-case key the lookup silently fails. More defensive:
```ts
const matchedKey = Object.keys(routeData.views ?? {}).find(
  (k) => k.toLowerCase() === normalizedSection
);
const view = matchedKey ? routeData.views![matchedKey] : undefined;
```

tylermorrisford (approved):
alles gut

misama-ct (approved):

valoriecarli (approved):

PR #1767, opened 2026-07-15
feat(alert): merge FeedbackCard capability into Alert (FEC-988)

misama-ct:
## Summary

Folds the (unreleased) **FeedbackCard** pattern's capability into the **Alert** component and removes FeedbackCard, backwards-compatibly. FeedbackCard was essentially "an Alert with a different look and a different announcement behaviour" — so rather than ship a parallel component, Alert gains the axes and behaviours needed to cover the inline agent-confirmation use case, and its accumulated debt is cleaned up along the way.

Alert is used ~48 times across 4 repos, so **backwards compatibility was the hard constraint.** Existing markup renders identically; every existing prop keeps working; `variant="outlined"` and `variant="flat"` are unchanged and `outlined` renders identically to `subtle`.

## What's new

- **Emphasis axis** — `variant="flat" | "subtle" | "outline" | "solid"` (`outlined` kept as a deprecated alias of `subtle`).
- **Layout axis** — `layout="stack" | "inline" | "banner"` (`stack` default; `inline` = content leading / actions trailing, wraps on narrow widths; `banner` = full-width edge-to-edge).
- **`colorPalette="neutral"`** now supported alongside `info` / `positive` / `warning` / `critical`.
- **Public `Alert.Icon` slot** + **`hideIcon`** — provide a custom icon or suppress the icon entirely; auto/explicit/hidden precedence.
- **`dismissible` + `onDismiss`** — a built-in localized dismiss button; a manually composed `Alert.DismissButton` still wins (no double button).
- **`Alert.Title` renders the `Heading` primitive** (default `as="p"`, promotable via `as`) and **`Alert.Description` renders `Text`**.
- Fixed **`Alert.DismissButton` prop forwarding** (consumer `variant`/`size`/`aria-label` now reach the interior `IconButton`).
- Removed the FeedbackCard pattern; its pedagogical examples were migrated into Alert stories as `layout="inline"` confirmations first.

## Behaviour changes (please review)

1. **Announcement politeness: assertive → polite.** `Alert.Root` now defaults to `role="status"` (polite live region) instead of `role="alert"`. `role` is now overridable via the native prop — **add `role="alert"` to any alert that must interrupt the user immediately**, or `role="group"` + `aria-label` for a silent inline confirmation.
2. **Unstyled default → `subtle` surface.** An Alert with no explicit `variant` now renders the `subtle` tinted surface (previously unstyled). `outlined`/`flat` are unchanged.

`Alert.Title` was deliberately pinned to `size="md"` (~16px) so its rendered appearance is **unchanged** — the primitive swap is semantic, not visual (verified via a computed-`fontSize` regression test; `heading`/`body` font tokens are identical).

## Testing

- `alert.stories.tsx` — 23/23 (source **and** built-bundle / CI parity)
- `alert.docs.spec.tsx` — 6/6
- `typecheck:dev` clean · clean production `build`
- New tests assert real behaviour (computed styles, DOM roles, `role` override precedence, dismiss-button count, `fn()` spies).

## Accessibility

Docs rewritten to describe the real role model (status default, `alert`/`group` overridable, no self-set `aria-live`), with a live-region priming note. Corrected two doc inaccuracies caught in review (a non-existent React Aria announcer API; an overclaimed auto-`aria-hidden` for custom icons).

## Follow-ups (non-blocking, tracked)

- `colorPalette="neutral"` without `hideIcon` renders an empty icon grid cell (~8px indent); docs recommend `hideIcon`. Cosmetic, new feature (not a regression).
- `flattenFragments` could produce colliding React keys with `hideIcon` + sibling/nested Fragment children (no current usage hits it).
- Test-symmetry / showcase-consolidation polish; a designer glance at Title `textStyle` line-height.
- `alert.figma.tsx` leaves the `layout` axis unmapped (connected Figma component exposes no such property) — for a Figma-authority maintainer.

## How it was built

Executed via subagent-driven development: one implementer per task with a two-stage (spec + quality) review after each, and a final whole-branch review — verdict **ready to merge**. Design doc and implementation plan are under `docs/superpowers/`.


misama-ct:
Closing this. `FeedbackCard` didn't _vibe_, making it a _variant_ of `Alert` creates it's own issues. This is a wasps-nest/rabbit-hole. I'll explore a more generic `<Item>` component approach [as can be seen here](https://ui.shadcn.com/docs/components/radix/item).

PR #1789, opened 2026-07-20
vc/ relocate `FlotingActionButton` component to patterns

valoriecarli:
FAB is structurally a pattern (lives in `src/patterns/buttons/`, thin wrapper around **IconButton,** no recipe/slots/React Aria, props type is just IconButtonProps), but its frontmatter declared it a Component, so it showed under Components → Buttons in the docs & Storybook.

This PR just updates it to be grouped with Patterns.

No changeset needed, docs-only recategorization.

### DIFF
--- a/packages/nimbus/src/docs/patterns-buttons.mdx
+++ b/packages/nimbus/src/docs/patterns-buttons.mdx
@@ -1,0 +1,24 @@
+---
+id: Patterns-Buttons
+title: Buttons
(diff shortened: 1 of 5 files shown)


tylermorrisford (approved):
I love it as long as them Michaels and Byrons are into it. thank you from the bottom of this bridal shop dressing room

misama-ct (approved):

PR #1796, opened 2026-07-21
feat(item): add Item and ItemGroup row components

misama-ct:
## Summary

Adds two new Nimbus components — **`Item`** (a horizontal
content-row compound) and **`ItemGroup`** (a standalone container that stacks
`Item` rows) — adapted from shadcn's React Aria `Item` and retrofitted to Nimbus
conventions (Chakra slot recipes, `useLink`, compound namespaces). They fill the
one container shape Nimbus lacked: a dense row of leading media + title/
description + trailing actions, complementing `Card` (vertical) and `List`/
`Menu` (interactive/selectable collections).

image

## Motivation

Product surfaces (settings rows, notification entries, file/resource rows)
repeatedly rebuild the same media·content·actions row by hand. `Item` gives them
one accessible, tokenized primitive. `FEC-988` — `Item` is the successor of the
originally planned component.

## What changed

- **`Item`** (`packages/nimbus/src/components/item/`) — compound namespace:
  `Item.Root`, `.Header`, `.Media`, `.Content`, `.Title`, `.Description`,
  `.Actions`, `.Footer`. Recipe `nimbusItem`: `variant` (`plain`/`outline`/
  `subtle`) × `size` (`xs`/`sm`/`md`); `Item.Media` carries its own `variant`
  (`default`/`icon`/`image`) via a `data-variant` attribute.
- **`ItemGroup`** (`packages/nimbus/src/components/item-group/`) — standalone
  peer: `ItemGroup.Root` (vertical stack) + `ItemGroup.Separator`. Recipe
  `nimbusItemGroup`.
- Recipes registered in the theme; both components exported from the package
  barrel.
- Full docs per component (designer `.mdx`, engineering `.dev.mdx`, `.a11y.mdx`,
  `.guidelines.mdx`) + consumer `.docs.spec.tsx`, a changeset, and the OpenSpec
  change (`nimbus-item` + `nimbus-item-group`).

## How it works

- **Link-upgrade, not a button.** `Item.Root` renders a `<div>` by default and
  upgrades to an accessible `<a>` via `react-aria`'s `useLink` when given an
  `href` (the same primitive Nimbus `Link` uses; Item's own recipe stays
  authoritative on appearance). There is deliberately **no** button/pressable
  mode — a row-as-button is semantically wrong and would invalidate nested
  action buttons. Real actions live in `Item.Actions`.
- **Actions don't hijack navigation.** In a link row, a capture-phase guard on
  the anchor cancels navigation when a click originates inside `Item.Actions`,
  so nested `Button`/`IconButton` controls stay independently operable. (React
  Aria's button stops propagation _below_ a container handler, so the guard runs
  in the capture phase.)
- **`ItemGroup` omits `role="list"`** on purpose: a list role requires
  `listitem` children and forbids the separator children this API composes, so
  it would fail axe's `aria-required-children`. The group is a plain stack;
  `ItemGroup.Separator` is `role="separator"`.

Two real defects were caught by the test discipline during development (an
action click navigating a link row; the invalid `role="list"` + separator) and
fixed as described above.

## Test plan

- [x] `typecheck:dev` — 0 errors
- [x] Storybook play functions (source): `item` 7/7, `item-group` 2/2 — incl.
      link-mode `<a>`, keyboard focus, action-without-navigation, media
      variants, grouped rows/separators; automatic axe checks pass
- [x] Consumer `docs.spec` (JSDOM): 2/2 (consumer-integration scenarios only;
      internal-behavior sections removed per testing-strategy guidelines)
- [x] `eslint` clean on both folders (incl. docs)
- [x] **Built-surface (CI gate):** `build` succeeded; `test:storybook` against
      `dist` — `item` 7/7, `item-group` 2/2
- [x] ItemGroup Base story tagged `vrt` for Chromatic visual regression

## Feedback wanted

- **Variant naming:** `Item` uses `outline`/`subtle` (matching the wider Nimbus
  vocabulary — `button`, `avatar`, `code`, `toast`), which differs from `Card`'s
  `outlined`/`muted`. Kept intentionally; is that the right call, or should Card
  be aligned separately?
- **Actions in link rows:** the capture-phase navigation guard supports
  "clickable row + trailing action." Is that the behavior we want, or should
  link rows disallow interactive actions entirely?
- **`ItemGroup` a11y:** dropping `role="list"` (for axe compliance with
  separators) vs. investing in `listitem` semantics — acceptable for v1?

## AI assistance disclosure

Generated with assistance from Claude Code. Human review checklist:

- [ ] Logic verified against intent
- [ ] No fabricated APIs / functions / imports
- [ ] Tests reflect actual behavior, not assumed behavior

Closes FEC-988

### DIFF
--- a/packages/nimbus/src/components/item-group/index.ts
+++ b/packages/nimbus/src/components/item-group/index.ts
@@ -1,0 +1,2 @@
+export * from "./item-group";
+export * from "./item-group.types";
--- a/packages/nimbus/src/components/item-group/index.ts
+++ b/packages/nimbus/src/components/item-group/index.ts
@@ -1,2 +1,2 @@
 export * from "./item-group";
-export * from "./item-group.types";
+export type * from "./item-group.types";
--- a/packages/nimbus/src/components/item/components/item.root.tsx
+++ b/packages/nimbus/src/components/item/components/item.root.tsx
@@ -66,8 +66,11 @@ const ItemRootLink = ({
     <ItemRootSlot
       as="a"
       ref={ref as React.Ref<HTMLDivElement>}
-      onClickCapture={handleClickCapture}
-      {...mergeProps(rest, linkProps)}
+      // Fold the guard into `mergeProps` (rather than passing it as a bare prop
+      // before the spread) so it *chains* with any consumer-supplied
+      // `onClickCapture` — react-aria composes `on*` handlers — instead of being
+      // silently overwritten by the spread.
+      {...mergeProps(rest, linkProps, { onClickCapture: handleClickCapture })}
     >
       {children}
     </ItemRootSlot>
--- a/packages/nimbus/src/components/item/components/item.root.tsx
+++ b/packages/nimbus/src/components/item/components/item.root.tsx
@@ -1,0 +1,112 @@
+import { useRef } from "react";
+import type { MouseEvent } from "react";
+import { useLink, useObjectRef, mergeProps } from "react-aria";
+import { mergeRefs } from "@/utils";
+import { ItemRootSlot } from "../item.slots";
+import type { ItemRootProps } from "../item.types";
+
+/**
+ * Internal: link-mode root.
+ *
+ * Renders the root slot as an `<a>` and applies the `useLink` hook (the same
+ * primitive Nimbus's `Link` uses) so the whole row is an accessible link —
+ * keyboard focusable, activatable with Enter, router-aware. The Item recipe
+ * remains authoritative for appearance; `useLink` only supplies interaction.
+ *
+ * Only the React Aria link options are handed to `useLink`; the remaining
+ * (styling / DOM) props are spread straight onto the slot.
+ */
+const ItemRootLink = ({
+  ref: forwardedRef,
+  children,
+  href,
+  target,
+  rel,
+  download,
+  ping,
+  referrerPolicy,
+  routerOptions,
+  onPress,
+  ...rest
+}: ItemRootProps) => {
+  const localRef = useRef<HTMLAnchorElement>(null);
+  const ref = useObjectRef<HTMLAnchorElement>(
+    mergeRefs<HTMLAnchorElement>(
+      localRef,
+      forwardedRef as React.Ref<HTMLAnchorElement>
+    )
+  );
+  const { linkProps } = useLink(
+    {
+      href,
+      target,
+      rel,
+      download,
+      ping,
+      referrerPolicy,
+      routerOptions,
+      onPress,
+      elementType: "a",
+    },
+    ref
+  );
+
+  // Guard against nested-interactive navigation: a click originating inside
+  // `Item.Actions` must not navigate the row. Runs in the capture phase (before
+  // the action's own handlers) so it only cancels the anchor's default/managed
+  // navigation, while the action control still receives and handles the event.
+  const handleClickCapture = (e: MouseEvent<HTMLDivElement>) => {
+    const target = e.target as Element | null;
+    if (target?.closest?.("[data-item-actions]")) {
+      e.preventDefault();
+    }
+  };
+
+  return (
+    <ItemRootSlot
+      as="a"
+      ref={ref as React.Ref<HTMLDivElement>}
+      onClickCapture={handleClickCapture}
+      {...mergeProps(rest, linkProps)}
+    >
+      {children}
+    </ItemRootSlot>
+  );
+};
+
+/**
+ * Internal: presentational (default) root — a plain `<div>` with no link or
+ * button semantics and no interactive tab stop of its own.
+ */
+const ItemRootStatic = ({
+  ref: forwardedRef,
+  children,
+  ...rest
+}: ItemRootProps) => (
+  <ItemRootSlot ref={forwardedRef} {...rest}>
+    {children}
+  </ItemRootSlot>
+);
+
+/**
+ * Item.Root - The row container and styling-context provider for all Item
+ * parts.
+ *
+ * Presentational by default (`<div>`). When `href` is provided it upgrades to
+ * an accessible link (`<a>`) via `useLink`. There is intentionally no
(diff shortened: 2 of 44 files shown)


misama-ct:
@claude please review

misama-ct:
## Triage + resolutions for the code review

@claude Thanks for the thorough pass. Verified each finding against the code (and against axe-core's actual rule internals for #1). Pushed `ab685fd`, `defabbd`, `7b1550d`.

### 🔴 #1 — `<button>` inside `<a>` (invalid content model) → **documented as a deliberate tradeoff**
The nesting is real and it is invalid per the HTML content model. On the "verify axe runs `nested-interactive`" ask: axe **does** run on every story (`a11y.test: "error"` globally), but the `nested-interactive` rule **cannot fire here by design**. Its `matches` only selects roles with `childrenPresentational: true` — verified `button → true`, `link → falsy`. So a control inside a link row structurally never trips the rule, which is why CI is green and always will be for this shape. There is no automated safety net for it.

Combining `href` + interactive `Item.Actions` is our own encouraged pattern (guard + `data-item-actions` + the `LinkWithActions` story), so rather than treat it as a bug we've made it an explicit, documented decision: a new "content-model tradeoff" section in `item.a11y.mdx` states the invalidity, why axe won't catch it, and how to opt for strict validity. Not re-architecting to a stretched-link for v1.

### 🔴 #2 — link inside `Item.Actions` won't navigate → **won't-fix, documented boundary**
Correct that the guard cancels a real `<a>`'s navigation. But `Item.Actions` is for JS-activated controls (`Button`/`IconButton`), not navigating links — that's out of the supported contract. Documented the boundary in `item.a11y.mdx` instead of engineering support for it.

### 🟡 #3 — capture handler → **split**
- **Clobber: fixed.** `onClickCapture` is now folded into `mergeProps(rest, linkProps, { onClickCapture: handleClickCapture })` so a consumer handler chains instead of overwriting the guard. Added a regression assertion to `LinkWithActions`.
- **Type `HTMLDivElement → HTMLAnchorElement`: keeping as-is.** `ItemRootSlot` is typed `SlotComponent<HTMLDivElement, …>`, so `onClickCapture` expects a div handler; the handler only reads `e.target` (an `Element`), never `currentTarget`. Switching to `HTMLAnchorElement` would fight the slot's declared type for no runtime benefit.

### 🟡 #4 — `ItemGroup` grouping → **added `role="group"`**
`ItemGroup.Root` now applies an overridable `role="group"` (with `aria-label`/`aria-labelledby` passthrough) — the grouping signal you suggested, without `role="list"`'s `aria-required-children` constraints. Docs + recipe comment updated; story asserts the named group.

### 🟢 Notes
- **Lifecycle:** `item.mdx` and the latest commit are `Beta` (intentional). The **PR description** still says "Experimental" — will update the PR body; no code change.
- Variant naming / recipe scoping: no action (agreed).

Verification: `typecheck:dev` clean, lint/prettier clean, item + item-group story tests 12/12 (a11y checks included).

ByronDWall:
> Variant naming: Item uses outline/subtle (matching the wider Nimbus
> vocabulary — button, avatar, code, toast), which differs from Card's
> outlined/muted. Kept intentionally; is that the right call, or should Card
> be aligned separately?

I think this is the right call, and if anything we should update `Card` to match the rest of the system


> Actions in link rows: the capture-phase navigation guard supports
> "clickable row + trailing action." Is that the behavior we want, or should
> link rows disallow interactive actions entirely?

I think this is the behavior we want, we can change it later if we find it to be problematic

> 
> ItemGroup a11y: dropping role="list" (for axe compliance with
> separators) vs. investing in listitem semantics — acceptable for v1?

I think investing in listitem semantics could be worth it?

packages/nimbus/src/components/item/components/item.root.tsx
```diff
@@ -0,0 +1,115 @@
...
+    },
+    ref
+  );
+
+  // Guard against nested-interactive navigation: a click originating inside
+  // `Item.Actions` must not navigate the row. Runs in the capture phase (before
+  // the action's own handlers) so it only cancels the anchor's default/managed
+  // navigation, while the action control still receives and handles the event.
+  const handleClickCapture = (e: MouseEvent<HTMLDivElement>) => {
+    const target = e.target as Element | null;
```
ByronDWall, line 59: nit: `target` here shadows the destructured `target` prop from line 23 (the link's `target` attribute). Not a bug since this closure doesn't need the outer one, but it reads a bit confusingly in a function that already deals with link targets.

Consider renaming to `eventTarget` or `clickOrigin`?

packages/nimbus/src/components/item/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./item";
+export * from "./item.types";
```
ByronDWall: Optional: since the `.types.ts` files export only TypeScript types, you could use `export type *` here for a slightly cleaner build signal. No functional difference, just a consistency option the barrel-exports guideline mentions.

```suggestion
export type * from "./item.types";
```

packages/nimbus/src/components/item-group/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./item-group";
+export * from "./item-group.types";
```
ByronDWall: Same suggestion here:

```suggestion
export type * from "./item-group.types";
```

ByronDWall (approved):
Really nice work. This is a clean, well-architected pair of components. The presentational-by-default / link-upgrade split is a smart design, the capture-phase navigation guard for nested actions is well thought out, and the accessibility rationale (especially the `role="group"` choice over `role="list"`) is thorough.

The documentation is outstanding: four tabs of docs per component, consumer tests that double as copy-paste examples, a VRT `SmokeTest` with offline imagery, and a `Showcase` story that exercises the full surface. Approving as-is; left a couple of minor suggestions below, nothing blocking.

PR #1805, opened 2026-07-22
vc/ resolve canary publishing error

valoriecarli:
Reference:[ Slack thread](https://commercetools.slack.com/archives/C0AF9C6TVKM/p1784292162265059)
[Error examples](https://github.com/commercetools/nimbus/actions/workflows/release.yml?query=is%3Afailure+branch%3Amain)

Currently canary fails on any feature merge & is skipped only when the merge is the VP PR itself.

The canary snapshot step in `release.yml `runs on every non-release merge to main:
`pnpm changeset version --snapshot canary`
`pnpm changeset publish --tag canary`

  - snapshot only re-versions packages that have a pending changeset & `@commercetools/nimbus-docs-build` was publishable (not private, not in ignore, not in the fixed group) but doesn't receive changesets, so it kept its static 1.1.0 while everything else got a fresh timestamped version. 
  
 - changeset publish then tried to republish 1.1.0 → X & `set -euo pipefail` threw an error.
 
It is believed to have been accidentally published in the first place.

The fix
 - Matched the repo's existing convention for internal packages (nimbus-i18n, color-tokens are both private & in ignore):
   - packages/nimbus-docs-build/package.json → added "private"
   - .changeset/config.json → added @commercetools/nimbus-docs-build to ignore 
   
Deemed safe b/c:
 - its only consumer is apps/docs (itself private, via workspace:^)
 - nimbus-mcp only references it in comments/docs, not as a dependency
 - the existing 1.1.0 on npm stays live for anyone who pinned it

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -16,7 +16,7 @@
       "@commercetools/nimbus-mcp"
     ]
   ],
-  "ignore": ["@commercetools/nimbus-i18n", "color-tokens", "blank-app", "docs"],
+  "ignore": ["@commercetools/nimbus-i18n", "@commercetools/nimbus-docs-build", "color-tokens", "blank-app", "docs"],
   "access": "restricted",
   "baseBranch": "main",
   "bumpVersionsWithWorkspaceProtocolOnly": false,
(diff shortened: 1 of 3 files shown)


ByronDWall (approved):
rebase and we good

PR #1814, opened 2026-07-22
fix(ci): send diagnostic messages to stderr in bundle size check

ByronDWall:
## Summary
- Fixes bundle size CI step crashing with `SyntaxError: Unexpected token 'B', "BUNDLE_SIZ"... is not valid JSON`
- `check-bundle-size.mjs` wrote informational messages (e.g. "BUNDLE_SIZE_BASELINE not set") to stdout via `console.log`, but in CI stdout is redirected to a JSON file — corrupting it with non-JSON text
- Changed two `console.log` calls to `console.error` so only JSON goes to stdout when `--json` is used

## Test plan
- [ ] Existing `check-bundle-size.spec.mjs` tests continue to pass
- [ ] CI bundle size check step no longer crashes on PRs where the baseline comment chain is missing

### DIFF
--- a/.github/actions/bundle-size/check-bundle-size.mjs
+++ b/.github/actions/bundle-size/check-bundle-size.mjs
@@ -150,7 +150,7 @@ function loadBootstrapBaseline() {
   if (!existsSync(BOOTSTRAP_PATH)) {
     return null;
   }
-  console.log(
+  console.error(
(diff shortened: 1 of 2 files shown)


misama-ct:
@ByronDWall I reviewed and took the liberty to merge, as @tylermorrisford  seems to be OOO till monday.

PR #1816, opened 2026-07-23
fix(tabs): stop forwarding an empty href to non-link tabs

misama-ct:
## Summary
Plain `Tabs` tabs logged a spurious React console error in development — _"An empty string (`""`) was passed to the `href` attribute"_ — one per tab. This removes it without changing any behavior.

## Root cause
React Aria keys its link handling off the **presence** of the `href` prop, not its value. `Tabs.Tab` always spread `href={href}` onto React Aria's `Tab`, so an ordinary tab (with `href` undefined) still carried the `href` key. React Aria's `useLinkProps` coerces a present-but-nullish href to `""` (`props.href ?? ''`), and since the tab is a non-link it renders as a `<div role="tab">`. React then rejects the resulting `href=""` on that non-anchor element with a dev-only `console.error`. Surfaced when opening the icon detail dialog on the docs Icons page (4 usage tabs → 4 errors), but it affected every non-link tab in the app.

## Fix
Only forward `href`/`target`/`rel`/`routerOptions` to React Aria's `Tab` when an `href` is actually provided, so plain tabs never carry the key. The fix lives in the single `Tabs.Tab` wrapper, so it covers both the compound (`Tabs.Tab`) and the `tabs`-prop APIs. Link tabs (with an `href`) are unaffected and still render as anchors.

## Regression test
New `tabs.spec.tsx` (JSDOM, mirrors the existing `markdown.spec.tsx` console-assertion pattern — the empty attribute is stripped from the final DOM, so the console error is the only observable signal):
- no empty-`href` warning for children-based plain tabs
- no empty-`href` warning via the `tabs` prop
- a tab **with** an `href` still renders as `<a href="…">`

## Test plan
- [x] New spec fails on the old code, passes on the fix
- [x] Full Tabs unit specs — 10/10 (incl. existing `.docs.spec`)
- [x] Tabs Storybook tests in a real browser (dev/source) — 12/12
- [x] `typecheck:dev` 0 errors · eslint 0 · prettier clean

## Notes
- Consumer-facing patch changeset included (`@commercetools/nimbus`).
- No linked ticket — this was an ad-hoc bug report.

## AI assistance disclosure
Generated with assistance from Claude Code. Human review checklist:
- [ ] Logic verified against intent
- [ ] No fabricated APIs / functions / imports
- [ ] Tests reflect actual behavior, not assumed behavior

### DIFF
--- a/.changeset/fix-tabs-empty-href.md
+++ b/.changeset/fix-tabs-empty-href.md
@@ -1,0 +1,8 @@
+---
+"@commercetools/nimbus": patch
+---
+
+`Tabs`: fixed a spurious React console error — _"An empty string (`""`) was
+passed to the `href` attribute"_ — that appeared in development for every
+ordinary (non-link) tab. Plain tabs no longer emit the warning. Tabs that opt
+into link behavior with an `href` are unaffected and still render as anchors.
--- a/packages/nimbus/src/components/tabs/components/tabs.tab.tsx
+++ b/packages/nimbus/src/components/tabs/components/tabs.tab.tsx
@@ -18,15 +18,16 @@ export const TabsTab = ({
   routerOptions,
   ...props
 }: TabProps) => {
+  // React Aria keys its link handling off the *presence* of the `href` prop,
+  // not its value: forwarding `href={undefined}` still makes it coerce the href
+  // to an empty string, and React then rejects the resulting `href=""` on the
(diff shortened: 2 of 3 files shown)


ByronDWall (approved):
Thank you so much for fixing this, it was getting to me too

ByronDWall (approved):

PR #1829, opened 2026-07-27
chore: bundle size workflow killswitch

tylermorrisford:
## feat(ci): add repository variable killswitch for bundle-size monitoring

Adds an escape hatch for maintainers to pause the bundle-size workflow immediately — no PR, no code change, one click in GitHub Settings.

### Motivation

The bundle-size comment chain is critical infrastructure but has proven fragile. When it breaks, every open PR is blocked until the chain is manually restored. There was no way to quickly un-block the team while a fix was in flight — you either left CI red for everyone or started reverting workflow changes. This closes that gap.

### What changed

A single repository variable (`BUNDLE_SIZE_ENABLED`) now gates both places where the feature touches CI:

- **`build-and-test.yml`** — the bundle-size check step is skipped when `BUNDLE_SIZE_ENABLED = false`, so no failing CI and no bot comments during the pause.
- **`bundle-size-comment.yml`** — the post-merge labeling step is also skipped, preventing broken-comment PRs from accumulating the `bundle-sizes` label while the feature is off (which would otherwise break the chain again on re-enable).

The default behavior is unchanged. The variable is absent in normal operation, and `'' != 'false'` evaluates to true, so nothing changes for everyday CI runs.

### How to use it

**To pause:** GitHub → Settings → Secrets and variables → Actions → Variables → create `BUNDLE_SIZE_ENABLED` with value `false`. Takes effect on the next CI run — no merge required.

**To resume:** delete the variable or set its value to anything other than `false`.

### Docs

`docs/bundle-size-monitoring.md` gains an "Emergency Pause (Killswitch)" section with step-by-step toggle instructions and a note about chain state on re-enable. "First-Time Setup" is renamed to "Restoring a Broken Chain" (more accurate for its primary use case) and the internal link is updated to match.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -25,7 +25,7 @@ jobs:
         uses: ./.github/actions/ci
 
       - name: Bundle size check
-        if: github.event_name == 'pull_request'
+        if: github.event_name == 'pull_request' && vars.BUNDLE_SIZE_ENABLED != 'false'
         uses: ./.github/actions/bundle-size
         with:
           gh-token: ${{ github.token }}
--- a/.github/workflows/bundle-size-comment.yml
+++ b/.github/workflows/bundle-size-comment.yml
@@ -14,7 +14,7 @@ jobs:
   post-merge-label:
     # For direct trigger: only run on merged PRs.
     # For workflow_call: caller is responsible for gating on merged.
-    if: inputs.pr-number || github.event.pull_request.merged == true
+    if: (inputs.pr-number || github.event.pull_request.merged == true) && vars.BUNDLE_SIZE_ENABLED != 'false'
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
--- a/docs/bundle-size-monitoring.md
+++ b/docs/bundle-size-monitoring.md
@@ -176,6 +176,12 @@ To re-enable, delete the variable or set its value to anything other than
 labeled PR before the pause. If that PR's comment has no valid data block,
 follow the [chain restoration steps](#restoring-a-broken-chain) below.
 
+**PRs opened during the pause:** these never received a bundle-size comment, so
+if one is merged after re-enabling, the post-merge workflow will add the
+`bundle-sizes` label but the comment will have no data block. The next PR will
+skip past it and fall back to the last labeled PR with a valid comment — so the
+chain recovers automatically as long as that earlier PR is still intact.
+
 ## Running Locally
 
 You need built packages and the [GitHub CLI](https://cli.github.com) (`gh`)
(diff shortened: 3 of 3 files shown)


.github/workflows/bundle-size-comment.yml
```diff
@@ -14,7 +14,7 @@ jobs:
   post-merge-label:
     # For direct trigger: only run on merged PRs.
     # For workflow_call: caller is responsible for gating on merged.
-    if: inputs.pr-number || github.event.pull_request.merged == true
+    if: (inputs.pr-number || github.event.pull_request.merged == true) && vars.BUNDLE_SIZE_ENABLED != 'false'
```
ByronDWall, line 17: **Suggestion:** The explicit parens here are a nice touch for readability. Worth noting that even without them the result would be the same (`==` binds tighter than `||` and `&&` in Actions expressions), but I think this is strictly better — keeps the intent obvious at a glance.

docs/bundle-size-monitoring.md
```diff
@@ -157,6 +157,25 @@ The action:
...
+   value to `false`.
+3. All subsequent CI runs will skip both the bundle-size check and the
+   post-merge labeling step.
+
+To re-enable, delete the variable or set its value to anything other than
+`false`.
+
+**When re-enabling after a pause:** the comment chain resumes from the last
+labeled PR before the pause. If that PR's comment has no valid data block,
+follow the [chain restoration steps](#restoring-a-broken-chain) below.
```
ByronDWall, line 177: **Suggestion:** This is a helpful callout. You might consider adding a brief note about what happens to PRs that were _opened_ during the pause — i.e., they never got a bundle-size comment, so if someone merges one of those after re-enabling, the post-merge label will be applied but there won't be a data block on that PR's comment. Probably fine to leave for a follow-up if you'd rather keep this focused.

.github/workflows/build-and-test.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         uses: ./.github/actions/ci
 
       - name: Bundle size check
-        if: github.event_name == 'pull_request'
+        if: github.event_name == 'pull_request' && vars.BUNDLE_SIZE_ENABLED != 'false'
```
ByronDWall, line 28: **Suggestion (future-proofing):** Right now the killswitch check lives in two separate workflow files. If a third bundle-size-related workflow shows up later, someone would need to remember to add the guard there too. Not a problem today — just flagging it so it's on the radar. One option down the road would be a small reusable workflow or composite action that centralizes the gate, but that's overkill for two call sites.

ByronDWall (approved):
Nice, clean change — killswitch is well-scoped and the docs section reads great. A couple of small suggestions inline, nothing blocking. Ship it!

PR #1807, opened 2026-07-22
FEC-1151 & 1152 - Chromatic batches 4 & 5

valoriecarli:
> You know the drill. Got all my notes, so ask away if you'd like. 

P1 Nimbus - Batch 4 & 5 Audit - Selection controls & time / Field composition & badges
  Relates to: #1797 (Batch 3), #1790 (Batch 2) & #1782 (Batch 1)

  VRT components audited & tagged:

  Batch 4 (FEC-1156) - selection controls & time
  - Checkbox
  - RadioInput
  - Switch
  - TagGroup
  - TimeInput

  Batch 5 (FEC-1157) - field composition & badges
  - Badge
  - Alert
  - FormField
  - FieldErrors
  - LocalizedField

Test-integrity fixes
Extended the VRT guidance in chromatic-related docs (chromatic-visual-testing.md, stories.md & writing-stories skill)

Possible bugs flagged noted, but not fixed here

PR #1797, opened 2026-07-21
FEC-1155 - Batch3 - Selection & File Inputs

valoriecarli:
**P1 Nimbus - Batch 3 Audit - Selection & file inputs**
  Relates to: #1790 (Batch 2) & #1782 (Batch 1)

VRT components audited & tagged (selection & file inputs)
  - `SearchInput`
  - `ScopedSearchInput`
  - `Select`
  - `DropZone`
  - `FileTrigger`

Each opts its resting-state matrix/showcase & a focus story into snapshots, plus the component-specific chrome (clear button, open popover, drag-over target, fused seam). Everything else stays behavioral (verified by play functions, not pixels).

Test-integrity fixes so play steps assert what their names claim.
Extended the VRT guidance in chromatic-related docs (chromatic-visual-testing.md, stories.md, writing-stories skill)

Bug fix (a11y): DropZone keyboard focus ring now renders in normal mode. focusRing was on the root, but keyboard focus lands on React Aria's inner button, so no ring showed.

Possible bugs flagged, not fixed here (RTL): ScopedSearchInput's fused seam & Select's clear/chevron cluster use physical CSS (borderRightRadius, right) rather than logical properties, so they mislay in RTL.

PR #1782, opened 2026-07-17
FEC- 1149 - Chromatic Batch1 - Date Components

valoriecarli:
PR contains:

1.) **CI** -  Omit changeset/release branches from Chromatic builds. Not needed. 

2.) **VRT Components audited & tagged**
- Avatar** - VRT tags on the new avatar stories introduced [here](https://github.com/commercetools/nimbus/pull/1726).
- Calendar
- DateInput 
- DatePicker 
- RangeCalendar
- DateRangePicker

Note that fixed-date stories snapshot. Live 'today' stories drift daily, so they're verified by play functions instead of pixels. This is the outcome of a prior [comment](https://github.com/commercetools/nimbus/pull/1695/changes#r3594792840).

PR #1830, opened 2026-07-27
FEC- 1182 & 1183 - Batch 11 (Layout primitives) & 12 (Structure, dividers & icons)

valoriecarli:
P1 Nimbus - Batch 11 & 12 Audit - Layout primitives / Structure, dividers & icons
    Relates to: #1818 (Batch 6 & 7), #1807 (Batch 4 & 5), #1797 (Batch 3), #1790 (Batch 2), #1782 (Batch 1)

  VRT components audited & tagged:

  Batch 11 (FEC-1182) - Layout primitives (None)
  - Box
  - Stack
  - Grid
  - SimpleGrid
  - Spacer

  Batch 12 (FEC-1183) - Structure, dividers & icons
  - Separator 
  - Icon 
  - InlineSvg 
  - Group 
  - Region
  
  tighten up docs

### DIFF
--- a/packages/nimbus/src/components/inline-svg/inline-svg.stories.tsx
+++ b/packages/nimbus/src/components/inline-svg/inline-svg.stories.tsx
@@ -72,6 +72,9 @@ const maliciousSvg = `<svg fill="none" height="24" stroke="currentColor" stroke-
  * Basic usage of InlineSvg component with a simple SVG path
  */
 export const Basic: Story = {
+  // VRT: baseline of the parse-and-render path (fill geometry).
+  tags: ["vrt"],
+  parameters: { chromatic: { disableSnapshot: false } },
   args: {
     data: simpleSvg,
     size: "md",
@@ -92,6 +95,9 @@ export const Basic: Story = {
  * Complex SVG with groups and multiple paths
  */
 export const ComplexSvg: Story = {
+  // VRT: stroke-based geometry, root stroke attrs preserved - Basic is fill-based.
+  tags: ["vrt"],
+  parameters: { chromatic: { disableSnapshot: false } },
   args: {
     data: complexSvg,
     size: "xl",
@@ -177,6 +183,9 @@ export const Colors: Story = {
  * Multi-color SVG (preserves inline colors)
  */
 export const MultiColorSvg: Story = {
+  // VRT: inline fills survive sanitization - the one surface `color` can't reach.
+  tags: ["vrt"],
+  parameters: { chromatic: { disableSnapshot: false } },
   args: {
     data: multiColorSvg,
     size: "xl",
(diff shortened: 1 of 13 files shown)


packages/nimbus/src/components/inline-svg/inline-svg.stories.tsx
```diff
@@ -72,6 +72,9 @@ const maliciousSvg = `<svg fill="none" height="24" stroke="currentColor" stroke-
  * Basic usage of InlineSvg component with a simple SVG path
  */
 export const Basic: Story = {
+  // VRT: baseline of the parse-and-render path (fill geometry).
+  tags: ["vrt"],
+  parameters: { chromatic: { disableSnapshot: false } },
   args: {
     data: simpleSvg,
     size: "md",
```
ByronDWall, line 80: Good picks on `Basic`, `ComplexSvg`, and `MultiColorSvg`. One thing I noticed: `Sizes` and `Colors` are left without VRT, but they render real visible surfaces (all six size values, six color variations). Icon's `Sizes` and `CustomColor` are opted in for equivalent coverage.

If the reasoning is that InlineSvg delegates to Icon internally so the recipe is already covered there, that makes sense to me. Could you add a short comment on `Sizes` and `Colors` noting that? Something like `// size axis covered by Icon VRT` so the omission reads as deliberate rather than overlooked.

packages/nimbus/src/components/separator/separator.stories.tsx
```diff
@@ -73,6 +73,9 @@ export const Vertical: Story = {
 };
 
 export const Orientations: Story = {
```
ByronDWall, line 75: Minor: `Orientations` and `MultipleBackgrounds` both render text labels inside the frame ("Content above horizontal separator", "White background", etc.). These will bake into the Chromatic baseline and diff if anyone tweaks the wording. Not a blocker since they provide useful spatial context for what the separator is separating, but worth a quick thought on whether you want that in the baseline. If you're happy with it, no change needed.

ByronDWall (approved):
Nice batch. The doc compression reads well, no semantic drift that I can see. The new "no-paint primitives" guideline is consistent across the skill, stories guide, and chromatic doc. Two small suggestions below, nothing blocking.

packages/nimbus/src/components/inline-svg/inline-svg.stories.tsx
valoriecarli: Good call, and yes. Comments have been added in https://github.com/commercetools/nimbus/pull/1830/commits/644328ded4c4d6af120b593d06eaf0ff723278aa

My notes for InlineSvg:
Screenshot 2026-07-28 at 13 20 40

packages/nimbus/src/components/separator/separator.stories.tsx
valoriecarli: You're right, thanks. I will leave this for now as the benefit of the added context outweighs the cost of a diff should the verbiage be updated.

PR #1847, opened 2026-07-29
chore(i18n): remove refs to translation requests

valoriecarli:
Source: https://commercetools.slack.com/archives/CCJ671V16/p1784896344483789

Updating documented Tx process

PR #1848, opened 2026-07-29
fix(ci): unbreak release publish under npm 12

valoriecarli:
The release workflow [failed](https://github.com/commercetools/nimbus/actions/runs/30469456601/job/90636403648#step:10:39) on the first merge to main after the 3.4.0 release (run 30469456601), with E403 "You cannot publish over the previously published versions: 3.4.0" for all five packages.

  Nothing on npm is damaged. 3.4.0 is fully published and correct. The workflow attempted to republish it.

  

> Root cause

  Two things combined:

  1. npm 12 changed the shape of npm info --json. It now wraps the manifest in an array ([ {...} ]); npm 11 returned a bare object. @changesets/cli 2.30.0 reads .versions off that value, which is undefined on an array, so its "is this version already on npm?" check reported every package as never published and it tried to publish over 3.4.0.
 
  2. The npm 11 pin was not in effect. release.yml pins npm@11 (added in #1744 for the npm 12 sigstore bug), but the shared .github/actions/ci action ran npm install -g npm@latest a few steps later, silently upgrading the runner to npm 12.0.1 before the publish step. npm 12.0.0 became latest on 2026-07-08, arming the bug three weeks ago.

> Why it surfaced now: 

changesets/action only enters publish mode when .changeset/ is empty. 
Every merge since npm 12 shipped had pending changesets, so it took the version-PR path instead and never exercised the
broken check.
#1839 (version packages) consumed all changesets, so the next merge, #1847, was the first to reach publish mode. Any PR in that slot would have failed identically.

>   The fix

  - @changesets/cli 2.30.0 → 2.31.1, which adds normalizeInfoJson() to unwrap npm 12's array response.
  - .github/actions/ci/action.yml: npm@latest → npm@11, so the deliberate pin in release.yml actually holds at runtime rather than being overridden mid-job.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@arethetypeswrong/cli": "catalog:tooling",
     "@babel/core": "catalog:tooling",
     "@changesets/changelog-github": "0.7.0",
-    "@changesets/cli": "2.30.0",
+    "@changesets/cli": "2.31.1",
     "@commercetools/nimbus": "workspace:^",
     "@eslint/js": "catalog:tooling",
     "@figma/code-connect": "catalog:tooling",
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "@arethetypeswrong/cli": "catalog:tooling",
     "@babel/core": "catalog:tooling",
     "@changesets/changelog-github": "0.7.0",
-    "@changesets/cli": "2.31.1",
+    "@changesets/cli": "catalog:tooling",
     "@commercetools/nimbus": "workspace:^",
     "@eslint/js": "catalog:tooling",
     "@figma/code-connect": "catalog:tooling",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -45,7 +45,7 @@
     "@arethetypeswrong/cli": "catalog:tooling",
     "@babel/core": "catalog:tooling",
     "@changesets/changelog-github": "0.7.0",
-    "@changesets/cli": "2.30.0",
+    "@changesets/cli": "2.31.1",
```
tylermorrisford: I think this needs to be in the `pnpm-workspace.yaml` file now

valoriecarli: @tylermorrisford dänks. updated in https://github.com/commercetools/nimbus/pull/1848/commits/a55a80ef7ec4cee6773813c614841510a7c73064

tylermorrisford (approved):
that's a yes from me

PR #1850, opened 2026-07-30
fix(badge,checkbox): add Badge sm variant and fix label-less Checkbox spacing

ByronDWall:
## Summary

- **Badge**: Added `sm` size variant between `xs` and `md` — uses `md` text sizing with `xs`-level density (gap, padding, icon size) at `h: 900`.
- **Checkbox**: Fixed excess horizontal spacing on label-less checkboxes by scoping `minWidth: 600` behind `&:has([data-slot='label'])` so it only applies when a visible label is present.

## Test plan

- [x] Verified Badge `sm` renders correctly in Storybook Sizes story
- [ ] Run `pnpm test:dev` to confirm all existing tests pass
- [ ] Chromatic visual regression captures the new Badge size
- [ ] Verify label-less Checkbox no longer has extra horizontal spacing in Storybook

### DIFF
--- a/.changeset/badge-sm-checkbox-spacing.md
+++ b/.changeset/badge-sm-checkbox-spacing.md
@@ -1,0 +1,9 @@
+---
+"@commercetools/nimbus": patch
+---
+
+`Badge`: Added `sm` size variant, providing an intermediate option between `xs`
+and `md`.
+
+`Checkbox`: Fixed excess horizontal spacing when rendered without a visible
+label.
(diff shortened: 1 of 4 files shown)


tylermorrisford (approved):
Reading the existing size tokens first was the right move — `sm`'s values hold up precisely: `fontSize: "400"` / `lineHeight: "500"` from `md`, `gap: "100"` / `px: "300"` / `_icon: 500×500` from `xs`, and `h: "900"` as the clean midpoint between xs's `800` and md's `1000`. The description matches the code exactly.

The checkbox fix is the right tool for the job. Unconditional `minWidth` on the root was always a blunt instrument — `:has([data-slot='label'])` scopes it to exactly the cases that need it, and leaves label-less checkboxes free to be as narrow as their control.

One thing mild: adding `sm` introduces a new valid value for the `size` prop — technically a `minor` semver bump — though the design system has precedent for treating new variants as `patch`. Either way, not a blocker.

-claudito

PR #1851, opened 2026-07-30
feat(data-table): add allowsPinning and allowsExpandColumn props

ByronDWall:
## Summary

- Add `allowsPinning` and `allowsExpandColumn` props to `DataTable`, both defaulting to `true` (no breaking change)
- `allowsPinning={false}` hides the pin column header and all pin toggle buttons
- `allowsExpandColumn={false}` hides the expand chevron column; rows with nested content can still be expanded via row click when no `onRowClick` handler is provided (if `onRowClick` is set, it takes precedence)
- Fixes `colSpan` calculation for nested row cells to account for hidden columns
- Documents all three usage patterns in `data-table.dev.mdx`: hidden pin column, hidden expand column with auto row-click, and hidden expand column with custom `onRowClick` + controlled expansion

## Test plan

- [x] `HiddenPinColumn` story — verifies no pin column header or pin buttons render
- [x] `HiddenExpandColumn` story — verifies no expand column renders, row click expands/collapses nested content
- [x] `HiddenPinAndExpandColumns` story — verifies both columns hidden, only data column headers present, row click still expands
- [x] `HiddenExpandColumnWithRowClick` story — verifies `onRowClick` takes precedence, controlled expansion works for expand/collapse
- [x] All 3164 tests pass (`pnpm test:dev`)
- [x] Typecheck clean (`pnpm --filter @commercetools/nimbus typecheck:dev`)


tylermorrisford (approved):
One minor observation: the four new stories reach rows via `querySelectorAll("tbody tr")` rather than ARIA roles. That's probably an established DataTable convention where rows don't carry accessible names, so it's fine — just worth being aware of if the internal DOM structure changes.

---

The `hasExpandableContent` / `showExpandColumn` split is the key structural decision and it's right — separating "do rows have content to expand" from "is the expand column visible" is the only way to make the nested row render (`{hasExpandableContent && (`) and the auto-row-click behavior independent of the column's visibility, without reintroducing the old boolean's dual meaning.

The `colSpan` fix is a good catch. The hardcoded `+ 1` was silently wrong whenever pinning was disabled — nested row cells would have been one column too wide. Replacing it with `(showPinColumn ? 1 : 0)` is exactly right.

`expandViaRowClick` is computed correctly: `hasExpandableContent && !showExpandColumn && !onRowClick` — it only activates when there's data to expand, the column is hidden, and no custom handler claims the click. The per-row `hasNestedContent` guard inside the timeout means leaf rows stay inert even when `expandViaRowClick` is true.

All three doc examples cover the right cases: simple hidden column, auto-expand via row click, and the controlled expansion pattern for when `onRowClick` is also present. Test plan fully checked and 3164 tests passing is a clean stamp. -claudito

PR #1852, opened 2026-07-30
feat(data-table): add renderNestedContent prop for expandable row content

ByronDWall:
## Summary

- Add `renderNestedContent` prop to DataTable for rendering a uniform component template below each row when expanded
- Uses the same expansion system as `nestedKey` — shows an expand chevron column by default, with row-click-to-expand when `allowsExpandColumn={false}`
- The render callback receives the row and a `DataTableNestedContentOptions` object with a `close` function for collapsing the content from within
- State is managed via `expandedRows` / `onExpandRowsChange`; the callback now receives `toggledRowId` and `columnId` arguments
- When both `renderNestedContent` and `nestedKey` are provided, rows with `nestedKey` data render that content; others fall through to `renderNestedContent`
- New `DataTableNestedContentOptions` type exported for consumer use
- `aria-controls` set on expanded rows linking to the nested content panel
- i18n label added for nested content rows (`Nimbus.DataTable.nestedContentRow`)

## Test plan

- [x] `RowNestedContent` story: expand/collapse via chevron, colspan, multiple rows open
- [x] `RowNestedContentWithSelection` story: checkbox doesn't toggle content, expand button does
- [x] `RowNestedContentWithClose` story: close callback dismisses panel, aria-controls present when expanded / absent when collapsed
- [x] `RowNestedContentViaRowClick` story: no chevron rendered with `allowsExpandColumn={false}`, row click expands/collapses
- [x] `NestedContentOverride` story: nestedKey data overrides renderNestedContent, fallback works
- [x] Docs spec tests: expand via button, toggle, close callback, aria-controls, onRowClick interaction
- [x] Existing nestedKey stories still pass (no regression)


valoriecarli (approved):
you already fixed the one thing i was eyeing, so 💯

packages/nimbus/src/components/data-table/components/data-table.row.tsx
```diff
@@ -395,10 +429,11 @@ const DataTableRowInner = <T extends DataTableRowItem = DataTableRowItem>({
           columns={activeColumns}
           ref={rowRef}
           id={row.id}
-          data-clickable={!!onRowClick && !isDisabled}
+          data-clickable={isClickable && !isDisabled}
+          hasChildItems={!!renderDetails || undefined}
```
tylermorrisford: Suggestion, and partly a genuine question — if I'm reading RAC right, `hasChildItems` here plus `expandedKeys={renderDetails ? detailExpandedRows : undefined}` in `data-table.table.tsx` promotes these rows to treegrid semantics, so `aria-expanded` ends up tracking the detail panel. What got me curious is a table that also passes `nestedKey`: the chevron toggles `expanded` while `aria-expanded` reflects `detailExpandedRows`, so a row whose chevron is visibly open could announce as collapsed. You'd know far better than me whether `renderDetails` + `nestedKey` is a combination worth supporting — if it isn't, a line in the prop JSDoc would save the next person the trip; if it is, a story pairing them would lock it in. Either answer works for me.

packages/nimbus/src/components/data-table/components/data-table.row.tsx
```diff
@@ -171,6 +177,9 @@ const DataTableRowInner = <T extends DataTableRowItem = DataTableRowItem>({
             if (expandViaRowClick && hasNestedContent) {
               toggleExpand(row.id);
             }
+            if (renderDetails) {
+              toggleDetails(row.id);
+            }
```
tylermorrisford: Small follow-on to the same thought: with nested content, no expand column and no `onRowClick`, `expandViaRowClick` is true and `renderDetails` is set, so a single click runs both `toggleExpand` and `toggleDetails`. Might well be fine — or even desirable — just flagging that it's reachable.

packages/nimbus/src/components/data-table/data-table.stories.tsx
```diff
@@ -2599,6 +2599,301 @@ export const FlexibleNestedChildren: Story = {
...
+        () => {
+          expect(checkbox).toBeChecked();
+        },
+        { timeout: 3000 }
+      );
+
+      const openDetails = canvasElement.querySelectorAll(
+        "[data-detail-row-expanded='true']"
+      );
+      expect(openDetails.length).toBe(0);
```
tylermorrisford: Dropping the 600ms sleep for a `waitFor` is the right instinct, and I only noticed the wrinkle because `handleRowClick` guards on a 300ms `setTimeout`: selection resolves within a few ms, so `expect(openDetails.length).toBe(0)` lands before the panel could have opened either way — the step would pass even if checkbox clicks did leak through to the row handler. If you want to keep the assertion sharp, the `waitFor` plus a wait past the 300ms window gives you both; entirely your call whether that's worth the wall-clock.

packages/nimbus/src/components/data-table/data-table.stories.tsx
```diff
@@ -2599,6 +2599,301 @@ export const FlexibleNestedChildren: Story = {
...
+        },
+        { timeout: 3000 }
+      );
+
+      const detailCell = canvasElement.querySelector("[data-detail-cell]");
+      expect(detailCell).toBeInTheDocument();
+
+      const headerCells = canvasElement.querySelectorAll("thead th");
+      const colSpan = detailCell?.getAttribute("colspan");
+      expect(Number(colSpan)).toBe(headerCells.length);
```
tylermorrisford, line 2714: Tiny coverage note, no action needed — this comparison only diverges from the `activeColumns.length + …` sum when `allowsPinning={false}`, which is the config the hardcoded colSpan slipped through last round. If you ever want a regression guard on that arithmetic, that's the knob to turn.

tylermorrisford (approved):
Re-reviewed after the fix commit — everything from last round landed cleanly: the detail cell's colSpan derives from the same internal-column arithmetic as the nested row, the `if (nonInteractiveCell)` guards became real `expect(...).toBeDefined()` assertions, `ariaRef` cleans up after itself, and the plan file is out of the diff. The `close` callback threaded through `DataTableRowDetailsOptions` is a nice touch — dismissible panels without the consumer tracking state.

Left three suggestions inline plus a small coverage note. None are blocking and I'd be glad to see this land as-is.

-claudito

PR #1858, opened 2026-07-31
chore(skills): use native Atlassian MCP linking instead of shell script

ByronDWall:
## Summary
- Replace the shell script generation step in `jira-generate-tickets-from-plan` skill with direct `mcp__atlassian__createIssueLink` MCP calls
- The Atlassian MCP now supports issue link creation natively, eliminating the need for a separate bash script with API tokens
- Replace hardcoded `CRAFT-` references with generic `PROJECT-` placeholders and "Jira keys" terminology
- Add explicit warning about unintuitive `inwardIssue`/`outwardIssue` direction semantics, with advice to verify the first link in the UI
- Improve error recovery: try `mcp__atlassian__fetch` DELETE before falling back to manual cleanup

## Test plan
- [ ] Invoke `/jira-generate-tickets-from-plan` with a plan file and verify dependency links are created via MCP
- [ ] Verify link direction is correct (predecessor shows "is predecessor of")
- [ ] Verify `dependency` link type is used (not "Blocks")

### DIFF
--- a/.claude/skills/jira-generate-tickets-from-plan/SKILL.md
+++ b/.claude/skills/jira-generate-tickets-from-plan/SKILL.md
@@ -234,11 +234,6 @@ additional_fields:
 After all tickets are created with their real Jira keys, create dependency links
 using `mcp__atlassian__createIssueLink`.
 
-### Link Type Discovery
-
-First, call `mcp__atlassian__getIssueLinkTypes` to discover available link types
-on the instance.
-
 ### Preferred Link Type
 
 Use the `dependency` link type, NOT "Blocks". Predecessor/successor conveys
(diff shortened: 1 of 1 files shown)


tylermorrisford (approved):
One thing in the new **Error Recovery → Wrong link direction** flow: the steps recreate the links with `inwardIssue`/`outwardIssue` swapped first, then note the user may need to delete the bad ones manually. Since there's no MCP delete for issue links, that ordering leaves the ticket showing both "is predecessor of" and "is successor of" until someone cleans up by hand. Flipping it — prompt for the manual delete, then recreate — avoids the duplicate window. Being upfront that the delete is manual is the right call either way.

Also minor: "First, call `mcp__atlassian__getIssueLinkTypes`" reads as unconditional, but the next section hardcodes `dependency` and Error Recovery already covers the not-found path. Might land better as a fallback than a required first step.

---

Good to see this limitation actually go away — I checked, and `mcp__atlassian__createIssueLink` is live with `inwardIssue` documented as the blocker/predecessor, so the direction guidance carried over correctly. Trading 86 lines of curl, base64 auth, and that `((var++))`-under-`set -e` footgun for four MCP parameters is a great deal.

-claudito

valoriecarli (approved):

PR #1859, opened 2026-07-31
fix(supply-chain): flip workspace settings

tylermorrisford:
This pull request updates the repository's supply-chain protection settings for dependency updates, making the enforcement of the `minimumReleaseAge` gate stricter and updating documentation and procedures to match. The main changes are: enabling strict enforcement of the age gate, ensuring that too-new package versions are never allowed (even temporarily), and clarifying that bypasses must never be automated or committed. The documentation is updated throughout to reflect these new requirements.

**Dependency resolution policy tightening:**

* In `pnpm-workspace.yaml`, `minimumReleaseAgeStrict` is now set to `true`, so dependency resolution will fail with an error if a requested version is too new, rather than silently bypassing the age gate. `minimumReleaseAgeIgnoreMissingTime` is set to `false` to ensure all packages are subject to the age gate, even those missing publish times.

**Documentation updates:**

* The `.claude/commands/housekeeping.md` instructions are updated to explain that with strict mode enabled, any attempt to install a package version younger than the age gate will result in a hard failure (`ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION`), not a silent bypass. [[1]](diffhunk://#diff-2783e359c3c3749cae902f40212c4d3c148434540addf41619838db099e30acdL86-R96) [[2]](diffhunk://#diff-2783e359c3c3749cae902f40212c4d3c148434540addf41619838db099e30acdL117-R121)
* The post-install supply-chain guard now instructs users to revert to an age-appropriate version and explicitly forbids adding `minimumReleaseAgeExclude` entries to bypass the gate. It also documents how to force a real resolution pass if needed.
* The section on handling failures is retitled and rewritten to clarify that bypassing the age gate is never acceptable in automated processes, and any necessary exceptions must be made manually and reviewed.

### DIFF
--- a/.claude/commands/housekeeping.md
+++ b/.claude/commands/housekeeping.md
@@ -152,7 +152,12 @@ For each dependency group (or the specified target group):
    younger than the age gate. Revert that package's catalog entry to its
    previous (age-appropriate) version, `pnpm install` again, and report the
    package as "held back by minimumReleaseAge". You MUST NOT add a
-   `minimumReleaseAgeExclude` entry to get past it:
+   `minimumReleaseAgeExclude` entry to get past it.
+
+   Under `minimumReleaseAgeStrict: true` pnpm cannot append that block itself,
+   so the check below is not the failure signal — the error above is. It is a
+   self-audit that you did not hand-add an exclude while working around a
+   held-back package:
 
    ```bash
    grep -n 'minimumReleaseAgeExclude' pnpm-workspace.yaml   # expect: no output
--- a/.claude/commands/housekeeping.md
+++ b/.claude/commands/housekeeping.md
@@ -83,14 +83,17 @@ For each dependency group (or the specified target group):
 1. **Fetch Latest Versions (age-gated):**
 
    This repo enables pnpm's `minimumReleaseAge` supply-chain gate in
-   `pnpm-workspace.yaml` (default `1440` minutes = 24h): freshly published
-   versions are deliberately held back until they have "aged", giving the
-   ecosystem time to catch a compromised release. Your target version MUST
-   respect that gate. If you pick a version younger than the gate, pnpm cannot
-   resolve it and will **silently rewrite `pnpm-workspace.yaml` with a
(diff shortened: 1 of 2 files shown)


tylermorrisford:
@claude please review

tylermorrisford:
Triaged all three — one taken, one I can rule out with evidence, one acknowledged.

**1. The vestigial `grep` — agreed, fixed.** Good catch on the framing. Under `minimumReleaseAgeStrict: true` pnpm can't append that block itself, so `# expect: no output` was carrying over the old detection semantics. It still earns its place as a self-audit (the surrounding text forbids the agent from hand-adding an exclude, and this is how the run proves it didn't), so I relabeled rather than removed it:

> Under `minimumReleaseAgeStrict: true` pnpm cannot append that block itself, so the check below is not the failure signal — the error above is. It is a self-audit that you did not hand-add an exclude while working around a held-back package.

**2. `--resolution-only` is real on 11.10.0 — verified, no change needed.** It's in `pnpm install --help` on the pinned version:

```
--resolution-only    Re-runs resolution: useful for printing out peer dependency issues
```

And I ran the full command in this repo before writing the doc line — it did a genuine 6.3s pass over 1813 packages, versus ~190ms for a bare `pnpm install`. Worth noting `--lockfile-only` is **not** a substitute here: it short-circuits under `optimisticRepeatInstall: true` exactly like a plain install does (187ms, "Already up to date"), which is the failure mode that doc note exists to warn about.

**3. The missing-`time` hard-fail surface — acknowledged, and partly measured.** Correct read of the consequence. One data point in its favor: that 1813-package resolution pass ran *with* `minimumReleaseAgeIgnoreMissingTime: false` already flipped and resolved cleanly, so nothing currently in the tree is missing `time`. Agreed it's the first suspect if CI starts hard-failing on an otherwise-fine package later, and the comment in `pnpm-workspace.yaml` should point whoever hits it in the right direction.

And agreed on the changeset — workspace config plus an agent command doc, nothing consumer-facing.

-claudito

ByronDWall (approved):

PR #1861, opened 2026-07-31
feat(data-table): expose Row, Cell, Column as compound sub-components + Sweet, Sweet DataTable Render Performance Improvements

ByronDWall:
## Summary

- Expose `DataTable.Row`, `DataTable.Cell`, and `DataTable.Column` as compound sub-components for fine-grained rendering control
- `Body`, `Header`, and `Row` accept optional render-function children that replace only data columns/cells — internal columns (selection, expand, pin) are always rendered automatically
- New exported types: `DataTableColumnRenderProps`, `DataTableRowRenderProps`, `DataTableCellRenderProps`
- Improved hover and re-render performance: removed compositing-layer churn from row hover styles and ref-stabilized consumer callback props so inline callbacks no longer cascade re-renders to every row
- Visual fixes: moved table border to root container, fixed sticky cell backgrounds with custom row bg, aligned header selection checkbox

## Test plan

- [ ] Compound composition stories render correctly (`CompoundCustomRowStyling`, `CompoundCustomRowWithSelection`, `CompoundCustomRowWithExpansion`, `CompoundCustomCells`, `CompoundCustomColumns`)
- [ ] All 50 existing DataTable story tests pass
- [ ] Default (no custom children) rendering is unchanged
- [ ] Selection, expansion, sorting, and pinning work with custom children
- [ ] Custom row `bg` prop propagates to sticky cells via `data-custom-bg`
- [ ] Hover performance is smooth — no compositing layer churn or context cascade
- [ ] `typecheck:dev` passes with no data-table errors


ByronDWall:
@claude lay a review on me claudicito

ByronDWall:
@claude uno mas mi claudicita

ByronDWall:
@claude una breve resena mi claudito

packages/nimbus/src/components/data-table/components/data-table.row.tsx
```diff
@@ -617,8 +629,8 @@ const DataTableRowInner = <T extends DataTableRowItem = DataTableRowItem>({
                   ? nestedKey && Array.isArray(row[nestedKey])
                     ? `${(row[nestedKey] as unknown[]).length} nested items`
                     : nestedKey && (row[nestedKey] as React.ReactNode)
-                  : renderNestedContent
-                    ? renderNestedContent(row, {
+                  : renderNestedContentRef.current
+                    ? renderNestedContentRef.current(row, {
```
tylermorrisford: This is the one I'd want a decision on. The other four callbacks you ref-stabilized are only invoked from event handlers, so the latest-ref pattern is exactly right for them. `renderNestedContent` is different — it's read during render, and the only thing about it left in context is the `hasRenderNestedContent` boolean. So when a consumer passes an inline function that closes over changing state, `contextValue` is unchanged, no row re-renders, and the open panel keeps rendering the first closure's output until expansion, pin state, or row data changes. Async-loaded detail content is the obvious case: panel opens, data arrives, panel doesn't update.

Two ways out and I'd be happy with either: document it alongside the `Body`/`Row` constraint (it's the same class of thing, but it's a *shipped* prop changing behavior rather than a new opt-in one), or put `renderNestedContent` itself back in `contextValue` — its identity churning is arguably the exact signal you want to react to here, unlike `onRowClick`.

packages/nimbus/src/components/data-table/components/data-table.body.tsx
```diff
@@ -46,21 +86,7 @@ export const DataTableBody = <T extends DataTableRowItem = DataTableRowItem>({
         renderEmptyState={renderEmptyState ?? DefaultEmptyStateMessage}
         {...restProps}
```
tylermorrisford, line 87: Suggestion: `restProps` spreads *after* `dependencies={[activeColumns, expanded, pinnedRows, pinnedRowIds]}` on line 85, and `DataTableBodyProps` only omits `children` — so `dependencies` is still part of the public surface. A consumer who reads the new reactivity constraint and reaches for the obvious lever, `<DataTable.Body dependencies={[activeId]}>`, silently *replaces* the internal array and loses expansion and pin invalidation. Worth noting `data-table.row.tsx` spreads `restProps` before its own `dependencies`, so `Row` wins that fight and `Body` loses it.

Matching Row's ordering fixes it. Merging instead — `[...internal, ...(restProps.dependencies ?? [])]` — would fix it *and* hand consumers the escape hatch the constraint note currently leaves them without.

packages/nimbus/src/components/data-table/data-table.recipe.ts
```diff
@@ -126,57 +128,27 @@ export const dataTableSlotRecipe = defineSlotRecipe({
             left: "2400",
           },
         _hover: {
```
tylermorrisford, line 130: I checked the removed declarations against the base block above: the `drag` / `selection` / `expand` z-indexes and `left` offsets are identical, so that dedupe is safe and the block really was dead weight. The one that isn't a pure duplicate is `[data-slot='pin-row-cell']` — it had `zIndex: 10` under `_hover` and the base sets `3`, so the pin cell now stays at 3 while hovered. Left- and right-sticky cells don't overlap, so I doubt it's visible, but it's the one value that genuinely changes.

Separately: hover only reaches a Chromatic snapshot if a play function leaves it applied, so if none of these stories does, this whole class of change merges unverified. Worth a look given the border also moved to the scroll container.

.changeset/datatable-compound-subcomponents.md
```diff
@@ -0,0 +1,21 @@
...
+- `DataTable.Row` children receive `({ columns, row, isDisabled })` for custom
+  cell rendering
+- `DataTable.Header` children receive `({ columns, allowsSorting })` for custom
+  column headers
+- New types `DataTableColumnRenderProps`, `DataTableRowRenderProps`, and
+  `DataTableCellRenderProps` are exported for typing render functions
+
+`DataTable`: Improved hover and re-render performance. Removed compositing-layer
+churn from row hover styles and ref-stabilized consumer callback props so inline
+callbacks no longer cascade re-renders to every row.
```
tylermorrisford, line 21: Two additions worth considering, since this is the text consumers actually read: if `renderNestedContent` keeps the ref treatment, its reactivity change belongs here — it's existing behavior changing, unlike the new opt-in render props. And `depth?: number` comes off `DataTableRowProps`; nothing in the repo reads it so removing it looks right, but it's a public type narrowing in a `minor` and anyone passing it gets a fresh TS error.

tylermorrisford (approved):
Read this as a perf change with a compound API bolted on, and both halves hold up better than I expected. A few things I specifically liked: applying the latest-ref pattern to the *event* callbacks while keeping `isRowClickable` / `hasRenderNestedContent` as booleans in context — that's the right split, you keep the signal rendering decisions need without the identity churn. Documenting the reactivity constraint in both the type JSDoc and `dev.mdx` in the same words is the part that usually gets skipped. And the comment owning why `DataTableColumnRenderProps` isn't generic over `T` is more honest than quietly casting.

Four inline notes: three suggestions, and one on `renderNestedContent` I'd call more than a suggestion — though the fix might just be a doc line.

-claudito

PR #1875, opened 2026-08-06
FEC-1180 & 1188 - Batch 9 & 17, Navigation, Lists & Combobox

valoriecarli:
FEC-1180 & 1188 - Batch 9 & 17, Navigation & lists, Combobox

The PR contains VRTs for:
- TabNav 
- Tabs
- Combobox 
- Tree 
- DraggableList 
- Link 
- Pagination 
    
Known bug captured in a baseline: Pagination's page input clips page numbers above 999 (FEC-1272). `Locales` keeps `currentPage={2500}` so the clipping stays visible until it's fixed.

Associated docs continue to be revised & a full sweep will be done at the end.

### DIFF
--- a/packages/nimbus/src/components/combobox/combobox.stories.tsx
+++ b/packages/nimbus/src/components/combobox/combobox.stories.tsx
@@ -336,6 +336,9 @@ export const MultiSelectCustomOptions: Story = {
  * - Error handling
  */
 export const AsyncLoading: Story = {
+  // VRT: the open listbox with async results loaded.
+  tags: ["vrt"],
+  parameters: { chromatic: { disableSnapshot: false } },
   render: () => {
     const [error, setError] = useState<string | null>(null);
     const getPokemonValue = useCallback((pokemon: Pokemon) => pokemon.name, []);
@@ -371,6 +374,7 @@ export const AsyncLoading: Story = {
   },
 
   play: async ({ canvasElement, step }) => {
+    canvasElement.style.caretColor = "transparent";
     const canvas = within(canvasElement);
 
     await step("Type search query", async () => {
@@ -799,6 +803,9 @@ export const AsyncMultiSelectCustomOptions: Story = {
  * Tests that leading element (icon) displays correctly when provided
  */
 export const LayoutLeadingElement: Story = {
+  // VRT: the `leadingElement` slot; SmokeTest renders no leading icon.
+  tags: ["vrt"],
+  parameters: { chromatic: { disableSnapshot: false } },
   render: () => {
     return (
       <ComposedComboBox
@@ -1113,6 +1120,9 @@ export const MultiSelectTagsInline: Story = {
  * Tests that tags wrap to new lines when space is limited by measuring container height
  */
 export const MultiSelectTagsWrapping: Story = {
+  // VRT: tags wrapping the field to a second row.
+  tags: ["vrt"],
+  parameters: { chromatic: { disableSnapshot: false } },
   render: () => {
     const [selectedKeys, setSelectedKeys] = useState<(string | number)[]>([1]);
 
@@ -1424,6 +1434,9 @@ export const InputWrapsToNewLine: Story = {
  * Tests that placeholder text displays when input is empty
  */
 export const InputPlaceholder: Story = {
+  // VRT: the `[data-placeholder]` treatment; the play clears back to it.
+  tags: ["vrt"],
+  parameters: { chromatic: { disableSnapshot: false } },
   render: () => {
     return (
       <ComposedComboBox
@@ -1435,6 +1448,7 @@ export const InputPlaceholder: Story = {
   },
 
   play: async ({ canvasElement, step }) => {
+    canvasElement.style.caretColor = "transparent";
     const canvas = within(canvasElement);
     const input = canvas.getByRole("combobox") as HTMLInputElement;
 
@@ -1685,6 +1699,9 @@ export const FocusLosesOnOutsideClick: Story = {
  * Tests that focus indicators are visible when navigating with keyboard
  */
 export const FocusIndicatorsVisible: Story = {
+  // VRT: `[data-focused]` outline on the active option.
+  tags: ["vrt"],
+  parameters: { chromatic: { disableSnapshot: false } },
   render: () => {
     return (
       <ComposedComboBox aria-label="Test combobox" items={simpleOptions} />
@@ -1692,6 +1709,7 @@ export const FocusIndicatorsVisible: Story = {
   },
 
   play: async ({ canvasElement, step }) => {
+    canvasElement.style.caretColor = "transparent";
     const canvas = within(canvasElement);
     const input = canvas.getByRole("combobox");
 
@@ -2777,6 +2795,9 @@ export const OptionHoverFeedback: Story = {
  * Tests that selected options have visual distinction (aria-selected, styling)
  */
 export const OptionSelectedVisuallyDistinguished: Story = {
+  // VRT: all three option states at once - selected, unselected and disabled.
+  tags: ["vrt"],
+  parameters: { chromatic: { disableSnapshot: false } },
   render: () => {
     const [selectedKeys, setSelectedKeys] = useState<(string | number)[]>([
       1, 3,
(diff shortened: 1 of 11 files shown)


packages/nimbus/src/components/tab-nav/tab-nav.stories.tsx
```diff
@@ -582,55 +669,106 @@ export const DeprecatedVariantAlias: Story = {
...
+ * is a product of both (a 2px bar for `line`, a full-height rounded-rect or
+ * capsule for `rounded`/`pill`, each at three widths).
  */
-export const SmokeTest: Story = {
+/**
+ * An item that is both current and disabled — reachable here because `isCurrent`
+ * and `isDisabled` are independent props. `_disabled` dims the item, but the
+ * indicator is a root-level sibling the dimming cannot reach, so the highlight
+ * stays at full strength over a dimmed label.
+ */
```
tylermorrisford, line 685: There are two doc comments stacked here. The first one (671-679) describes the `variant` x `size` grid, which is `SmokeTest` — but `SmokeTest` is down at line 725 now and has no description. Only the closest comment gets used as a story's description, so this write-up isn't showing up anywhere. Since the guide says the axis list should live in the matrix's doc comment, it's probably worth moving down with it.

packages/nimbus/src/components/tabs/tabs.stories.tsx
```diff
@@ -1319,3 +1325,156 @@ export const DeprecatedVariantAliases: Story = {
...
 };
+
+const VARIANTS = ["line", "rounded", "pill"] as const;
+
+/**
+ * Showcase the tab focus ring on each variant. The ring follows the tab's own
+ * `borderRadius` — square, `200`, `full` — so each variant is a separate surface,
+ * and only one element can hold focus at a time.
+ */
+const focusSelectedTab = async (canvasElement: HTMLElement) => {
```
tylermorrisford, line 1336: This looks like it's meant to describe the three `FocusedTab*` stories, but it's sitting on the helper function, so those three end up with no description in the docs. (The `// VRT:` line doesn't count — only `/** */` comments get picked up.) Moving this above `FocusedTabLine` and making the helper's version a regular `//` comment would keep it. `tab-nav.stories.tsx:451` does the same thing above `focusActiveItem`.

packages/nimbus/src/components/tabs/tabs.stories.tsx
```diff
@@ -1319,3 +1325,156 @@ export const DeprecatedVariantAliases: Story = {
...
+      // If the hook stopped, the cells would render with no highlight.
+      const indicators = Array.from(
+        canvasElement.querySelectorAll<HTMLElement>(".nimbus-tabs__indicator")
+      );
+      await expect(indicators.map((i) => i.style.opacity)).toEqual(
+        Array(9).fill("1")
+      );
+      await expect(indicators.filter((i) => !i.style.transform)).toHaveLength(
+        0
+      );
```
tylermorrisford, line 1477: The same check in TabNav puts these two assertions inside a `waitFor`, and I think Tabs wants that too. The step above only waits for the 9 tabs to render — it doesn't wait for the indicator to be measured and positioned, which happens a beat later. Both components do that the same way, so if TabNav needed to wait, this probably does as well. Cheap to add, and it saves a test that fails once in a while when CI is busy.

packages/nimbus/src/components/pagination/pagination.stories.tsx
```diff
@@ -717,10 +724,129 @@ export const WithoutPageSizeSelector: Story = {
...
+  render: () => (
+    <Stack direction="column" gap="600">
+      {["en-US", "de-DE", "es-ES", "fr-FR", "pt-BR"].map((locale) => (
+        <Stack key={locale} direction="column" gap="200">
+          <Text fontSize="300" color="neutral.11">
+            {locale}
+          </Text>
+          <NimbusProvider locale={locale}>
+            <Pagination
+              totalItems={250000}
```
tylermorrisford, line 826: One thing about the separator half of this story: Spanish doesn't put a separator in 4-digit numbers, so es-ES renders `2500` with nothing in it. Here are all five, at 2,500 pages and at 25,000:

```
en-US  "2,500"   "25,000"
de-DE  "2.500"   "25.000"
es-ES  "2500"    "25.000"
fr-FR  "2 500"   "25 000"
pt-BR  "2.500"   "25.000"
```

So the es-ES row isn't testing the separator at all right now. Multiplying `totalItems` by ten (`2500000`, which is 25,000 pages) gets all five to show one. `currentPage={2500}` can stay as it is, so the FEC-1272 clipping is still in the picture.

packages/nimbus/src/components/tree/tree.stories.tsx
```diff
@@ -696,7 +696,156 @@ const SmokeTestView = () => {
...
+      handle.focus();
+      await userEvent.keyboard("{Enter}");
+      await wait();
+    });
+
+    await step("Move the drop target down one position", async () => {
+      await userEvent.keyboard("{ArrowDown}");
+      await wait();
+
+      // Leave the drag open: this frame IS the snapshot.
```
tylermorrisford, line 837: This is on purpose and the comment says so — I'm just wondering about the story after it. `SmokeTest` runs right after this one in the same file, and this PR adds an Escape to DraggableList's `DisabledItems` for exactly this reason, so a drag doesn't carry over into the next story. The new `afterEach` closes dialogs and listboxes but not an open drag. If React Aria cancels the drag when the component unmounts then there's nothing here; if the DraggableList one was surviving unmount, this might too. Might be worth running tree's `SmokeTest` on its own and then right after this one, to see if it behaves the same both ways.

packages/nimbus/src/components/combobox/combobox.stories.tsx
```diff
@@ -6198,3 +6270,89 @@ export const ContentAreaScrollBehavior: Story = {
...
+      </Stack>
+    );
+  },
+};
+
+/**
+ * Option: Label And Description
+ * The two-line option layout, composed from the `label` and `description` slots.
+ */
+export const OptionWithDescription: Story = {
```
tylermorrisford, line 6322: `SmokeTest` is supposed to be the last story — the rule this PR restates in both docs — and this one comes after it, under the SMOKE TEST heading. Moving `OptionWithDescription` above that heading fixes both. Same thing over in `draggable-list.stories.tsx`, where `ColorPalettes` ends up after the new `SmokeTest`.

packages/nimbus/.storybook/vitest.setup.ts
```diff
@@ -1,7 +1,43 @@
...
+  try {
+    await waitFor(() =>
+      expect(document.querySelector(OPEN_OVERLAY)).toBeNull()
+    );
+  } catch {
+    // Overlays that intercept Escape (isKeyboardDismissDisabled) stay put; drop
+    // the inert so the next story is at least clickable.
+    document.querySelectorAll("[inert]").forEach((el) => {
+      if (!el.closest(".sb-wrapper")) el.removeAttribute("inert");
+    });
```
tylermorrisford: Two small things about this catch block.

It doesn't say anything when it runs, and if a dialog ignores Escape it's still on the page afterward. Some stories look for portal content with `within(canvasElement.parentNode)` — drawer does — so one of those could pick up the leftover dialog and fail with "found multiple elements" in a completely different file. A `console.warn` naming what was left open would save someone a long hunt.

Also, `.sb-wrapper` is a class from Storybook's own preview page (`sb-preparing-story`, `sb-nopreview`, `sb-errordisplay` in `storybook/assets/server/base-preview-body.html`). Vitest doesn't render that page, so `closest(".sb-wrapper")` is always null and the check never skips anything. Not harmful, it just looks like it's guarding something.

tylermorrisford (approved):
The doc changes are my favorite part of this. Deciding whether to build a matrix used to be a judgment call, and now there's something you can actually check: if the recipe has `compoundVariants` over two axes, build it. The note about `disabled` is good too — pulling it out into its own story only works when nothing else changes the same background, and that's an easy thing to get wrong twice, so writing it down with the FEC-1180 reference attached will save someone. I also like that every snapshotted story now has a one-line comment saying why it's there.

Everything below is a suggestion — nothing that should hold this up. If you only look at two: the doc comment above `DisabledCurrentItem` in tab-nav, where `SmokeTest` lost its description, and the missing `waitFor` in Tabs' `SmokeTest`, since the same check in TabNav has one.

-claudito

packages/nimbus/.storybook/vitest.setup.ts
valoriecarli: thanks, claudito, but this has already been banished.

valoriecarli:
Closed in favor of https://github.com/commercetools/nimbus/pull/1888
THINGS HAPPENED

PR #1888, opened 2026-08-07
FEC-1180 & 1188 - Batch 9 & 17, Navigation, Lists & Combobox

valoriecarli:
Things happened. This is take two for https://github.com/commercetools/nimbus/pull/1875

***

FEC-1180 & 1188 - Batch 9 & 17, Navigation & lists, Combobox

The PR contains VRTs for:
- TabNav 
- Tabs
- Combobox 
- Tree 
- DraggableList 
- Link 
- Pagination 

Known bug captured in a baseline: Pagination's page input clips page numbers above 999 (FEC-1272). Locales keeps currentPage={2500} so the clipping stays visible until it's fixed.

Associated docs continue to be revised & a full sweep will be done at the end.

### DIFF
--- a/packages/nimbus/src/components/combobox/combobox.stories.tsx
+++ b/packages/nimbus/src/components/combobox/combobox.stories.tsx
@@ -6273,6 +6273,48 @@ export const ContentAreaScrollBehavior: Story = {
   },
 };
 
+/**
+ * Option: Label And Description
+ * The two-line option layout, composed from the `label` and `description` slots.
+ */
+export const OptionWithDescription: Story = {
+  // VRT: the only frame with `[slot="description"]`; the rest are plain-text options.
+  tags: ["vrt"],
+  parameters: { chromatic: { disableSnapshot: false } },
+  decorators: [roomForPopover],
+  render: () => (
+    <ComboBox.Root aria-label="Option descriptions" menuTrigger="focus">
+      <ComboBox.Trigger />
+      <ComboBox.Popover>
+        <ComboBox.ListBox>
+          <ComboBox.Option textValue="Koala">
+            <Text slot="label">Koala</Text>
+            <Text slot="description">Sleeps up to 20 hours a day</Text>
+          </ComboBox.Option>
+          <ComboBox.Option textValue="Platypus">
+            <Text slot="label">Platypus</Text>
+            <Text slot="description">Lays eggs and is venomous</Text>
+          </ComboBox.Option>
+        </ComboBox.ListBox>
+      </ComboBox.Popover>
+    </ComboBox.Root>
+  ),
+  play: async ({ canvasElement, step }) => {
+    canvasElement.style.caretColor = "transparent";
+    const canvas = within(canvasElement);
+
+    await step("Opens with two-line options", async () => {
+      await userEvent.click(canvas.getByRole("combobox"));
+      await waitFor(() => {
+        expect(getListBox(document)).toBeInTheDocument();
+      });
+      expect(
+        document.querySelector('[role="option"] [slot="description"]')
+      ).toBeInTheDocument();
+    });
+  },
+};
+
 // ============================================================
 // SMOKE TEST
 // ============================================================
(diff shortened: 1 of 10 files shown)


packages/nimbus/src/components/tab-nav/tab-nav.stories.tsx
```diff
@@ -582,55 +669,106 @@ export const DeprecatedVariantAlias: Story = {
...
+ * is a product of both (a 2px bar for `line`, a full-height rounded-rect or
+ * capsule for `rounded`/`pill`, each at three widths).
  */
-export const SmokeTest: Story = {
+/**
+ * An item that is both current and disabled — reachable here because `isCurrent`
+ * and `isDisabled` are independent props. `_disabled` dims the item, but the
+ * indicator is a root-level sibling the dimming cannot reach, so the highlight
+ * stays at full strength over a dimmed label.
+ */
```
tylermorrisford, line 678: There are two doc comments stacked here. The first one (671-679) describes the `variant` x `size` grid, which is `SmokeTest` — but `SmokeTest` is down at line 725 now and has no description. Only the closest comment gets used as a story's description, so this write-up isn't showing up anywhere. Since the guide says the axis list should live in the matrix's doc comment, it's probably worth moving down with it.

packages/nimbus/src/components/tabs/tabs.stories.tsx
```diff
@@ -1319,3 +1325,156 @@ export const DeprecatedVariantAliases: Story = {
...
 };
+
+const VARIANTS = ["line", "rounded", "pill"] as const;
+
+/**
+ * Showcase the tab focus ring on each variant. The ring follows the tab's own
+ * `borderRadius` — square, `200`, `full` — so each variant is a separate surface,
+ * and only one element can hold focus at a time.
+ */
+const focusSelectedTab = async (canvasElement: HTMLElement) => {
```
tylermorrisford: This looks like it's meant to describe the three `FocusedTab*` stories, but it's sitting on the helper function, so those three end up with no description in the docs. (The `// VRT:` line doesn't count — only `/** */` comments get picked up.) Moving this above `FocusedTabLine` and making the helper's version a regular `//` comment would keep it. `tab-nav.stories.tsx:451` does the same thing above `focusActiveItem`.

packages/nimbus/src/components/tabs/tabs.stories.tsx
```diff
@@ -1319,3 +1325,156 @@ export const DeprecatedVariantAliases: Story = {
...
+      // If the hook stopped, the cells would render with no highlight.
+      const indicators = Array.from(
+        canvasElement.querySelectorAll<HTMLElement>(".nimbus-tabs__indicator")
+      );
+      await expect(indicators.map((i) => i.style.opacity)).toEqual(
+        Array(9).fill("1")
+      );
+      await expect(indicators.filter((i) => !i.style.transform)).toHaveLength(
+        0
+      );
```
tylermorrisford: The same check in TabNav puts these two assertions inside a `waitFor`, and I think Tabs wants that too. The step above only waits for the 9 tabs to render — it doesn't wait for the indicator to be measured and positioned, which happens a beat later. Both components do that the same way, so if TabNav needed to wait, this probably does as well. It's cheap to add, and it saves a test that fails once in a while when CI is busy. (No sign of that here — this run passed.)

packages/nimbus/src/components/pagination/pagination.stories.tsx
```diff
@@ -717,10 +724,129 @@ export const WithoutPageSizeSelector: Story = {
...
+  render: () => (
+    <Stack direction="column" gap="600">
+      {["en-US", "de-DE", "es-ES", "fr-FR", "pt-BR"].map((locale) => (
+        <Stack key={locale} direction="column" gap="200">
+          <Text fontSize="300" color="neutral.11">
+            {locale}
+          </Text>
+          <NimbusProvider locale={locale}>
+            <Pagination
+              totalItems={250000}
```
tylermorrisford: One thing about the separator half of this story: Spanish doesn't put a separator in 4-digit numbers, so es-ES renders `2500` with nothing in it. Here are all five, at 2,500 pages and at 25,000:

```
en-US  "2,500"   "25,000"
de-DE  "2.500"   "25.000"
es-ES  "2500"    "25.000"
fr-FR  "2 500"   "25 000"
pt-BR  "2.500"   "25.000"
```

So the es-ES row isn't testing the separator at all right now. Multiplying `totalItems` by ten (`2500000`, which is 25,000 pages) gets all five to show one. `currentPage={2500}` can stay as it is, so the FEC-1272 clipping is still in the picture.

packages/nimbus/src/components/combobox/combobox.stories.tsx
```diff
@@ -6198,3 +6272,89 @@ export const ContentAreaScrollBehavior: Story = {
...
+      </Stack>
+    );
+  },
+};
+
+/**
+ * Option: Label And Description
+ * The two-line option layout, composed from the `label` and `description` slots.
+ */
+export const OptionWithDescription: Story = {
```
tylermorrisford: `SmokeTest` is supposed to be the last story — the rule this PR restates in both docs — and this one comes after it, under the SMOKE TEST heading. Moving `OptionWithDescription` above that heading fixes both. Same thing over in `draggable-list.stories.tsx`, where `ColorPalettes` ends up after the new `SmokeTest`.

tylermorrisford (approved):
Same review as the one on #1875, since the work is the same — I'm carrying the five open comments over because they don't follow a closed PR.

Nice fix on the tree drag, and thank you for spelling out the mechanism in the comment. A module-global drag session that unmounting doesn't end is not something the next person would ever guess. Dropping the global `afterEach` in favor of it makes sense to me too: if the drag was the thing re-applying `inert`, then sweeping dialogs and listboxes was treating a symptom. `build-and-test` passing here is decent evidence that the narrow fix is enough.

All five below are suggestions, nothing that should hold this up. One caveat on the Tabs `waitFor` comment: this run went green, so treat it as a hunch about a busy CI machine rather than something that's actually failed.

-claudito

packages/nimbus/src/components/tab-nav/tab-nav.stories.tsx
valoriecarli: this was the cleanup i was going to go after in the final sweep, thanks! got a head start on it with https://github.com/commercetools/nimbus/pull/1888/commits/782705a5868307f2db01e79a62d8ec81cd72be92

packages/nimbus/src/components/tabs/tabs.stories.tsx
valoriecarli: same here

packages/nimbus/src/components/pagination/pagination.stories.tsx
valoriecarli: corrected in https://github.com/commercetools/nimbus/pull/1888/commits/782705a5868307f2db01e79a62d8ec81cd72be92

packages/nimbus/src/components/tabs/tabs.stories.tsx
valoriecarli: samesies

packages/nimbus/src/components/combobox/combobox.stories.tsx
valoriecarli: @tylermorrisford /claudito. this is true, but **_not_** included in this PR and will be addressed in one way or another during the final cleanup sweep next week. 
things happen.

PR #1900, opened 2026-08-10
test: stabilize flaky storybook interaction tests

ByronDWall:
## Summary

CI run [#31410284715](https://github.com/commercetools/nimbus/actions/runs/31410284715/job/93526422415) (PR #1897) failed 7 story tests across 5 files. That PR only touched a CI workflow file (`.github/workflows/renovate-changeset.yml`) — zero component/test code — confirming these are pre-existing flakes, not regressions.

All 7 failures trace back to the same root cause class: a `waitFor` assertion checking that a portal-rendered menu/dialog/popover has closed, timing out under CI resource contention (`browser.isolate: false` in `vitest.storybook.config.ts` runs many browser test files concurrently against one storybook dev server).

## Changes

- **rich-text-input**, **info-dialog**, **date-range-picker-field**: add explicit `{ timeout: 5000 }` to `waitFor` calls that had no timeout override (default 1000ms was too short).
- **split-button**: bump an existing `{ timeout: 3000 }` (added by #1808 for the same flake class) to `{ timeout: 8000 }` — 3s was no longer enough headroom.
- **splitter** ("Restore Defaults With Zero Size"): different symptom, same root cause class — an axe a11y violation (`scrollable-region-focusable`), because Zag's `ResizeObserver`-driven `tabIndex` update on the `ScrollArea` viewport hadn't settled before the `afterEach` a11y check ran. Added a `waitFor` asserting any overflowing viewport has `tabindex="0"` before the play function ends, rather than adjusting a timeout. `ScrollArea`'s own implementation is correct and untouched.

All new timeouts stay well under the project's 60s `testTimeout`.

## Testing

- Each file's play function fixed and verified individually.
- All 5 files run together: 83/83 tests passing.
- Splitter fix specifically re-run 4x to rule out residual flakiness (no a11y violations across any run).
- `prettier --check` and `eslint` clean on all 5 changed files.

### DIFF
--- a/packages/nimbus/src/components/rich-text-input/rich-text-input.stories.tsx
+++ b/packages/nimbus/src/components/rich-text-input/rich-text-input.stories.tsx
@@ -1045,9 +1045,12 @@ export const PendingMarksConsistency: Story = {
 
     // Close menu explicitly and wait for it to be gone to avoid toggle races
     await userEvent.keyboard("{Escape}");
-    await waitFor(() => {
-      expect(ui.queryByRole("menu")).not.toBeInTheDocument();
-    });
+    await waitFor(
+      () => {
+        expect(ui.queryByRole("menu")).not.toBeInTheDocument();
+      },
+      { timeout: 5000 }
+    );
 
     // Reopen and verify the Code mark stays checked (consistency)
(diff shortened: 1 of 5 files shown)


valoriecarli (approved):
thanks for doing this.

PR #1910, opened 2026-08-11
fix(nimbus): bound postinstall chakra typegen with a hard timeout

ByronDWall:
## Problem

`@commercetools/nimbus`'s `postinstall` script runs `chakra typegen` against the built theme entry point so consumers get typed recipe/theme props without a manual step:

```json
"postinstall": "test -f ./dist/index.es.js && chakra typegen ./dist/index.es.js || true"
```

In some Node/environment combinations (confirmed on a Vercel build, Node 22.22.2, `@chakra-ui/cli@3.36.1`), `chakra typegen` throws `ReferenceError: process is not defined` inside its VM-based bundle loader (`loadViaVm`) — and instead of exiting, the Node process then **hangs indefinitely**. The existing `|| true` guard only protects against a non-zero *exit code*, not a process that never terminates, so the whole install (and the consumer's build) stalls forever.

Observed live: the `mc-app-kit-playground` Vercel preview deploy for [commercetools/merchant-center-application-kit#4081](https://github.com/commercetools/merchant-center-application-kit/pull/4081) was stuck "pending" because of this, while a sibling deploy without this dependency path completed normally.

## Fix

Bound `chakra typegen` with a hard wall-clock timeout so `postinstall` can never hang again, regardless of what the upstream CLI does internally:

- **`scripts/lib/run-with-timeout.mjs`** — spawns a command in a detached process group with a timeout; on timeout, force-kills the whole subtree (`SIGKILL` to the process group, falling back to the immediate child). Never rejects/throws — always resolves `{ code, signal, timedOut, error }`.
- **`packages/nimbus/scripts/postinstall-typegen.mjs`** — thin CLI wrapper: exits immediately if the built entry point doesn't exist, otherwise runs `chakra typegen` through the timeout wrapper (30s) and always exits `0`, logging a one-line note to stderr on any non-success outcome (timeout / non-zero exit / spawn error).
- **`packages/nimbus/package.json`** — `postinstall` now calls the wrapper. `build-theme-typings` (the internal maintainer build step) is intentionally untouched — it should still fail loudly if broken.
- Changeset added (patch, `@commercetools/nimbus`).

## Testing

TDD — `scripts/lib/run-with-timeout.spec.mjs` covers:
- fast successful exit
- non-zero exit code
- a genuinely hanging child process + timeout (asserts bounded wall-clock time)
- a nonexistent binary (spawn error)

Also reproduced the real `chakra typegen` crash live against this repo's built `dist/index.es.js` and confirmed the wrapper catches it and exits `0` instead of hanging.

```
pnpm vitest --project scripts run
# Test Files  3 passed (3)
#      Tests  70 passed (70)
```

### DIFF
--- a/.changeset/postinstall-typegen-timeout.md
+++ b/.changeset/postinstall-typegen-timeout.md
@@ -1,0 +1,6 @@
+---
+"@commercetools/nimbus": patch
+---
+
+Fixed an issue where installing this package could hang indefinitely during
+`postinstall` type generation in some environments.
--- a/packages/nimbus/package.json
+++ b/packages/nimbus/package.json
@@ -152,7 +152,7 @@
     "typecheck": "tsc --noEmit -p tsconfig.prod.json",
     "typecheck:dev": "tsc --noEmit -p tsconfig.json",
     "build-theme-typings": "pnpm chakra typegen ./src/theme/index.ts",
-    "postinstall": "test -f ./dist/index.es.js && chakra typegen ./dist/index.es.js || true",
+    "postinstall": "node scripts/postinstall-typegen.mjs ./dist/index.es.js",
     "bundles:analyze": "ANALYZE_BUNDLE=true vite build",
     "storybook": "storybook dev -p 6006",
     "build-storybook": "storybook build",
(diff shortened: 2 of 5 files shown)


ByronDWall:
Closing in favor of #1911, which fixes the root cause (postinstall typegen was crashing on the full `dist/index.es.js` bundle) rather than just bounding the resulting hang with a timeout wrapper.

Issue #1509, opened 2026-05-15
Package shape: @commercetools/nimbus-tokens fails attw under modern Node resolution


misama-ct:
Surfaced by the package-shape gate added in #1498. `@commercetools/nimbus-tokens` is currently **excluded** from `scripts/check-package-shape.mjs` because the published tarball has two interacting shape issues that go beyond what the PR's framing wanted to absorb. Filing this so the gap is tracked.

## What attw / publint report

Against `packages/tokens` (`pnpm build:tokens && pnpm check:package-shape` with tokens re-added to `PACKAGES`):

| Resolution mode | Result |
|---|---|
| `node10` | 🟢 |
| `node16` from CJS | ⚠️ `ESM (dynamic import only)` + 🚭 `Unexpected module syntax` |
| `node16` from ESM | ❌ `No types` |
| `bundler` | ❌ `No types` |

publint corroborates:

> WARNING: `pkg.main` is `dist/commercetools-nimbus-tokens.cjs.js` and is written in CJS, but is interpreted as ESM. Consider using the `.cjs` extension.

## Root cause

`packages/tokens/package.json` declares `"type": "module"`, but preconstruct (2.8.13) emits the CJS bundle as `dist/commercetools-nimbus-tokens.cjs.js`. Because `"type": "module"` applies to all `.js` files in the package, Node interprets `.cjs.js` as ESM — so `require("@commercetools/nimbus-tokens")` from a CJS consumer dies before reading the file. Separately, the `exports` map has no `types` condition, so types only resolve under legacy `node10`.

## Fix paths

1. **Enable preconstruct's modern exports mode** (`"exports": true` in the package's preconstruct config). preconstruct emits a `dist/cjs/package.json` `{"type": "commonjs"}` marker and regenerates the exports map — should resolve both findings. Untested locally.
2. **Hand-write a `dist/cjs/` marker** as a postbuild step and update the exports map manually. Targeted, more code.
3. **Drop CJS** if no consumer needs it (unlikely — Jest defaults, classic Next pages dir).

## Done when

- `nimbus-tokens` is re-added to `PACKAGES` in `scripts/check-package-shape.mjs`.
- `pnpm check:package-shape` exits 0.
- The "What Gets Tracked" section in `docs/package-shape-verification.md` is updated to reflect 5 tracked packages.

PR #1911, opened 2026-08-11
fix(nimbus): repair postinstall typegen hook and broken CJS chunk requires

ByronDWall:
## What was broken

1. **`postinstall` crashed with `process is not defined`** — `chakra typegen` was pointed at `dist/index.es.js`, the entire component library. `@chakra-ui/cli@3.36.1` (latest) can't `require()` an ESM file, falls back to evaluating the bundle in a bare `vm` sandbox with no Node globals, and crashes on React's `process.env` check. Even patching that hole open, the bundle contains DOM-touching code at module scope (from a markdown dependency), which a headless sandbox can never satisfy — bundling the whole UI library for typegen was the wrong approach. The `|| true` in the script masked the crash from `pnpm install`'s exit code, so it silently never generated types.
2. **Found while verifying the fix: `require("@commercetools/nimbus")` was completely broken.** The Vite 8/Rolldown migration renames CJS chunks `.cjs.js` → `.cjs`, but only rewrote references from entry-point files, not chunk-to-chunk `require()` calls. Any CJS consumer hit `MODULE_NOT_FOUND`.

## What changed

- Added a lightweight `@commercetools/nimbus/theme` export (tokens/recipes/config only — no React, no DOM) and repointed `postinstall` + the docs' manual-generation command at `dist/theme.es.js`. Sidesteps both bugs above with no dependency patch needed.
- Extracted the chunk-rename logic into `scripts/lib/rewrite-cjs-chunk-extensions.mjs` (TDD: failing spec first, then the fix) so it rewrites chunk-to-chunk references too, not just entry points.
- Added a changeset (`minor`, since it adds a new export).

## Verification

- `pnpm run postinstall` succeeds cleanly end-to-end (no crash).
- `require('./dist/index.cjs')` no longer fails on chunk-name mismatches (a separate, already-tracked `@commercetools/nimbus-tokens` CJS bug — [#1509](https://github.com/commercetools/nimbus/issues/1509) — surfaces next; out of scope here).
- `pnpm typecheck:strict` clean.
- `pnpm check:package-shape` shows `@commercetools/nimbus/theme` green across node10/16(CJS)/16(ESM)/bundler.
- `pnpm check:bundle-size` flat (~0%).
- `pnpm test:unit` (1506 tests) and the new spec (5 tests) all pass.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/packages/nimbus/package.json
+++ b/packages/nimbus/package.json
@@ -17,6 +17,16 @@
       }
     },
     "./package.json": "./package.json",
+    "./theme": {
+      "import": {
+        "types": "./dist/theme.d.ts",
+        "default": "./dist/theme.es.js"
+      },
+      "require": {
+        "types": "./dist/theme.d.cts",
+        "default": "./dist/theme.cjs"
+      }
+    },
     "./setup-jsdom-polyfills": {
       "import": {
         "types": "./dist/setup-jsdom-polyfills.d.ts",
@@ -67,6 +77,7 @@
       "plugins/webpack": ["./dist/plugins/webpack.d.ts"],
       "plugins/vite": ["./dist/plugins/vite.d.ts"],
       "plugins/stub": ["./dist/plugins/stub.d.ts"],
+      "theme": ["./dist/theme.d.ts"],
       "*": ["./dist/index.d.ts"]
     }
   },
@@ -152,7 +163,7 @@
     "typecheck": "tsc --noEmit -p tsconfig.prod.json",
     "typecheck:dev": "tsc --noEmit -p tsconfig.json",
     "build-theme-typings": "pnpm chakra typegen ./src/theme/index.ts",
-    "postinstall": "test -f ./dist/index.es.js && chakra typegen ./dist/index.es.js || true",
+    "postinstall": "test -f ./dist/theme.es.js && chakra typegen ./dist/theme.es.js || true",
     "bundles:analyze": "ANALYZE_BUNDLE=true vite build",
     "storybook": "storybook dev -p 6006",
     "build-storybook": "storybook build",
--- a/packages/nimbus/scripts/postbuild-types.mjs
+++ b/packages/nimbus/scripts/postbuild-types.mjs
@@ -5,9 +5,12 @@
  *
  * Three tasks, in order:
  *
- *   1. Relocate setup-jsdom-polyfills.d.ts from dist/test/ to dist/ to match
- *      where the runtime files (.es.js / .cjs) land. vite-plugin-dts preserves
- *      src/ subdirs; vite emits the runtime flat for declared entry points.
+ *   1. Relocate setup-jsdom-polyfills.d.ts (from dist/test/) and theme.d.ts
+ *      (from dist/theme/index.d.ts) to dist/ to match where the runtime
+ *      files (.es.js / .cjs) land. vite-plugin-dts preserves src/ subdirs;
+ *      vite emits the runtime flat for declared entry points. dist/theme/
+ *      itself is kept — other components' .d.ts files still import from
+ *      ./theme/recipes, ./theme/tokens, etc.
  *
  *   2. Walk every .d.ts in dist/ and rewrite bare relative imports
  *      (`export * from './foo'`) to include explicit `.js` extensions
@@ -16,9 +19,10 @@
  *      moduleResolution: "nodenext". The shared rewriter lives at
  *      scripts/lib/rewrite-relative-imports.mjs.
  *
- *   3. Duplicate index.d.ts and setup-jsdom-polyfills.d.ts to .d.cts variants
- *      for the CJS exports path. Identical content; the extension flips how
- *      TypeScript interprets the module kind (CJS regardless of `"type"`).
+ *   3. Duplicate index.d.ts, theme.d.ts, and setup-jsdom-polyfills.d.ts to
+ *      .d.cts variants for the CJS exports path. Identical content; the
+ *      extension flips how TypeScript interprets the module kind (CJS
+ *      regardless of `"type"`).
  *
  * Steps 1 & 2 run before step 3 so the .d.cts copies inherit the rewritten
  * imports.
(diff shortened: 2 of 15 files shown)


packages/nimbus/package.json
```diff
@@ -152,7 +163,7 @@
     "typecheck": "tsc --noEmit -p tsconfig.prod.json",
     "typecheck:dev": "tsc --noEmit -p tsconfig.json",
     "build-theme-typings": "pnpm chakra typegen ./src/theme/index.ts",
-    "postinstall": "test -f ./dist/index.es.js && chakra typegen ./dist/index.es.js || true",
+    "postinstall": "test -f ./dist/theme.es.js && chakra typegen ./dist/theme.es.js || true",
```
tylermorrisford, line 166: Flag: `|| true` still swallows errors silently. Now that `typegen` actually succeeds it's benign — but if the target ever regresses we're back to a silent no-op with no error trail. Even a short inline comment (`# silent: postinstall must not fail consumer installs`) would make the intent explicit rather than ambiguous.

packages/nimbus/scripts/postbuild-types.mjs
```diff
@@ -52,6 +49,15 @@ const RELOCATIONS = [
     from: join(DIST, "test", "setup-jsdom-polyfills.d.ts"),
     to: join(DIST, "setup-jsdom-polyfills.d.ts"),
   },
+  {
+    // theme/index.d.ts only declares `system` typed against
+    // `@chakra-ui/react`'s public types — no relative imports to rewrite —
+    // so a plain copy is safe. Unlike dist/test/, dist/theme/ is NOT removed
```
tylermorrisford: Nit: "so a plain copy is safe" implies safety depends on `src/theme/index.ts` staying import-free. But `walkAndRewriteImports` runs over all `.d.ts` files in `dist/` after relocation anyway, so import-rewriting is handled regardless of what the source exports. Worth tightening — as written it could mislead the next reader into thinking the copy is only safe *because* there are currently no relative imports.

tylermorrisford (approved):
Beauty - one flag and one nit inline. Approval!

valoriecarli (approved):

packages/nimbus/package.json
ByronDWall: Addressed in 3e08b4aed: `packages/nimbus/package.json's postinstall `|| true` intent is now documented in `vite.config.ts`, next to the `theme` entry point comment it depends on (package.json is JSON and cannot hold an inline comment). The added note explains that theme typegen is a nice-to-have that must never fail a consumer's `pnpm install`, and points back at the theme entry point as the first place to check if it ever regresses.

packages/nimbus/scripts/postbuild-types.mjs
ByronDWall: Addressed in 1ad50379a: reworded the comment to say the plain copy is safe because `walkAndRewriteImports` rewrites relative imports across every `.d.ts` in `dist/` unconditionally after relocation — not because `src/theme/index.ts` happens to have no relative imports today. No longer implies the safety is contingent on the source file's current import shape.

ByronDWall:
Pushed three follow-up commits addressing review + reviewer-agent findings:

**e5c7dad0** `fix(tokens): repair broken CJS require() in nimbus-tokens`
`require("@commercetools/nimbus-tokens")` still threw `MODULE_NOT_FOUND` one layer deeper than the nimbus fix in this PR — same root cause (preconstruct emits `.cjs.js` under a `"type": "module"` package). Added `scripts/lib/rewrite-preconstruct-cjs-extensions.mjs` (+ spec) and `packages/tokens/scripts/postbuild-cjs-extensions.mjs`, wired into `build:tokens`. `exports["."].require` now points at the renamed `.cjs`/`.d.cts`. `main` is left as-is — preconstruct's own build validates that field against its filename convention, and modern resolvers (Node, TS node16/bundler) consult `exports` first anyway.

**1ad50379** `fix(nimbus): fix dangling .cjs.js references in per-component CJS entries`
`postbuild-types.mjs`'s `cjsEntryFiles` was derived from the curated `ENTRY_POINTS` list (6 entries) but `vite.config.ts` globs ~115 more per-component/pattern CJS entries under `dist/components/*.cjs`. None of those were passed to `fixCjsChunkExtensions`, so 114/115 still had dangling `.cjs.js` chunk references after a build. Replaced the curated list with `scripts/lib/list-cjs-entry-files.mjs`, which scans every entry `.cjs` file actually on disk (skipping `dist/chunks/`). Verified `grep -rl '.cjs.js' packages/nimbus/dist --include='*.cjs'` now returns zero matches, and `require("./dist/components/button.cjs")` succeeds.

**3e08b4ae** test coverage + doc/comment cleanup
- Added idempotency + bare-require regression tests to `rewrite-cjs-chunk-extensions.spec.mjs`.
- Rewrote the changeset to be consumer-focused per `docs/changeset-conventions.md`, and added a `@commercetools/nimbus-tokens` entry (the tokens fix above is consumer-visible for that package too).
- Replied inline to both open threads (postinstall `|| true` intent now documented in `vite.config.ts`; the theme.d.ts "plain copy is safe" comment reworded).

All verification passed: full rebuild from scratch, `require("./dist/theme.cjs")` and `require("./dist/components/*.cjs")` succeed, zero dangling `.cjs.js` references, `typecheck:strict`, `check:package-shape` (nimbus + nimbus-icons pass attw/publint cleanly), `lint`, and the full scripts test suite.

One thing surfaced during verification worth flagging explicitly: `require("./dist/index.cjs")` (the full component bundle) still throws — but with `ReferenceError: document is not defined`, not `MODULE_NOT_FOUND`. That's a different, pre-existing, unrelated issue: a vendored markdown dependency does `document.createElement(...)` at module scope, so the *full* eager bundle can't fully evaluate in headless Node without a DOM — which is exactly why this PR's `theme` entry point exists and is what `postinstall` targets instead of `index` (see the comment in `vite.config.ts`). `require("./dist/theme.cjs")` and every per-component entry (`require("./dist/components/button.cjs")`, etc.) work correctly with no crash. Not something this PR's scope covers.

`nimbus-tokens` remains excluded from `pnpm check:package-shape` (`scripts/check-package-shape.mjs`) — the runtime CJS bug (issue #1509) is fixed here, but that issue also covers a separate, still-open gap (no `types` condition on the `exports["."].import` branch, so ESM consumers under node16/bundler resolution get no types). Left the comment there updated to reflect that only the CJS half is now fixed.

PR #1936, opened 2026-08-19
feat(popover): export Popover as a public compound component

misama-ct:
## Summary

`Popover` already existed, but only as an internal single-element wrapper — its
own JSDoc said "this component is only used internally" — sitting behind
Combobox's dropdown and LocalizedField's hint box. This promotes it to a public
compound component (`Popover.Root`, `Popover.Trigger`, `Popover.Content`) and
moves both internal consumers off the old shape.

image

Ships as Beta. Two `minor` changesets: the new component, and a LocalizedField
accessibility fix that fell out of the migration.

## Motivation

There was no anchored-overlay primitive between `Tooltip` (plain hints) and
`Dialog` (flows that should block the page). Filter panels, short edit forms and
context-sensitive actions had to be hand-rolled from React Aria's
`DialogTrigger` + `Popover` + `Dialog` — which is exactly what LocalizedField
was doing, and it shipped an unnamed dialog as a result.

## Why not split

The recipe registration moves from `recipes` to `slot-recipes`: `nimbusPopover`
becomes a slot recipe the moment the compound lands. Both existing consumers
read the old plain recipe, so they change meaning in the same commit and have to
move with it. The docs set, spec rewrite and consumer migrations are all
downstream of that one switch.

## Map

| Area | Files | Purpose |
| --- | --- | --- |
| Component | `popover/components/popover.{root,trigger,content,context}.tsx` | The compound. Root is the single configuration surface and passes overlay props to Content through context |
| Styling | `popover.recipe.tsx`, `popover.slots.tsx`, `theme/{recipes,slot-recipes}/index.ts` | Plain recipe → slot recipe, re-registered |
| Public API | `popover.types.ts`, `components/index.ts` | Per-part prop contracts; barrel export |
| Docs | `popover.{mdx,dev.mdx,a11y.mdx,guidelines.mdx}` | Standard four-file set |
| Tests | `popover.stories.tsx`, `popover.docs.spec.tsx` | 19 stories, 8 consumer tests |
| Consumers | `combobox/**`, `localized-field/**` | Combobox moves to React Aria's `Popover` directly; LocalizedField adopts the public compound |
| Spec | `openspec/specs/nimbus-popover/`, `openspec/changes/archive/**` | Spec rewrite plus two archived proposals |

## How to review

1. `openspec/changes/archive/2026-08-20-popover-compound-export/design.md` — the
   API decisions, so the rest reads as consequences.
2. `popover.types.ts`, then the four `components/*.tsx` — contract, then
   implementation.
3. `popover.stories.tsx` — the behaviour, in play-function form.
4. The two consumer migrations. They are the only places existing behaviour
   could regress.

The docs and openspec files are prose — skim or skip.

## Test plan

- [x] `pnpm test:dev packages/nimbus/src/components/popover/` — 27 pass (19
      stories + 8 consumer tests)
- [x] `pnpm --filter @commercetools/nimbus typecheck:dev` clean
- [x] `build-and-test` and `chromatic` — still running at time of writing
- [x] **Chromatic baseline needs a human accept.** LocalizedField's `HintDialog`
      snapshot changes on purpose: one shadow instead of two stacked, no border
      or background tint of its own, fades in and out like other popovers, and
      no outline around itself when open (matching `Dialog` and `Drawer`).

## Risk

Medium. The new component is additive behind a new export, but two shipped
components changed rendering path. Combobox is the one to watch: it no longer
routes through the shared recipe, so `zIndex: 1` is now declared on its own
`popover` slot to keep stacking order unchanged.

Rollback is a plain revert — no migration, no infrastructure, no persisted
state. Popover is unreleased, so only the Combobox and LocalizedField changes
have consumer-visible surface.

FEC-1167


misama-ct:
@claude review this, gurrrrrl

misama-ct:
@claude have another look. I fixed the issues you mentioned but also refactored the API.

misama-ct:
@claude Review this branch.

packages/nimbus/src/components/popover/popover.a11y.mdx
```diff
@@ -0,0 +1,126 @@
...
+`isNonModal` does not lift that containment. What it changes is the page behind
+the popover: by default React Aria marks the rest of the page `inert` and locks
+page scroll, and `isNonModal` leaves both alone, so assistive technologies can
+reach the surrounding content. Focus still cycles inside the dialog and an
+outside press still dismisses it — containment comes from the dialog
+`Popover.Content` renders, not from modality. React Aria's own guidance is that
+most popovers should not use it, because screen readers can announce the
+surrounding page in ways that make the popover harder to follow. It exists for
+components like combobox, where the trigger must keep focus while the overlay is
+open.
```
ByronDWall: nit: this paragraph seems like a better fit for the 'implementation' section?

ByronDWall (approved):
One very small nit on the docs. The component behaves as expected in storybook, code looks good. approved, thanks for implementing this!

packages/nimbus/src/components/popover/popover.a11y.mdx
misama-ct: [Done](https://github.com/commercetools/nimbus/pull/1936/changes/2a3ab0a88d3c23d21d69338631a1571b633bae18)

PR #1574, opened 2026-06-08
feat(tree): add Tree component wrapping React Aria Tree (FEC-985)

misama-ct:
image

## Summary

Adds a new **`Tree`** compound component for displaying hierarchical data such as file trees and nested navigation structures, implementing [FEC-985](https://commercetools.atlassian.net/browse/FEC-985) (epic FEC-979 — Nimbus Chat Components).

It wraps React Aria's `Tree` / `TreeItem` / `TreeItemContent` and exposes a four-part compound API:

- **`Tree.Root`** — the tree container; owns the slot recipe + `size` variant and forwards `dragAndDropHooks` for opt-in drag-and-drop.
- **`Tree.Item`** — a single node (`role="row"` with `aria-level`).
- **`Tree.ItemContent`** — the content row; applies level-based indentation and auto-renders a selection checkbox in multiple-selection mode.
- **`Tree.Indicator`** — the expand/collapse chevron; visible only for items with children, rotates when expanded.

## What's included

- `nimbusTree` slot recipe (`root`, `item`, `itemContent`, `indicator`) with `sm`/`md` sizes, registered in the theme.
- Keyboard navigation (arrows, Home/End, type-ahead), expand/collapse, single & multiple selection, disabled items.
- Visual indentation by nesting level via React Aria's `--tree-item-level` custom property.
- Static and dynamic (`items` + recursive `Collection`) composition.
- Opt-in drag-and-drop: reorder, re-parent (drop onto a group), drop indicators, and keyboard DnD — documented with a worked example in `tree.dev.mdx`.
- Stories with play-function interaction tests, `tree.docs.spec.tsx`, developer docs, an OpenSpec change, and a changeset.

## Accessibility note

React Aria 1.17 implements the tree as an ARIA **`treegrid`** (`role="treegrid"` → `row` → `gridcell`, with `aria-level`/`aria-expanded`/`aria-selected`) — its screen-reader-tested pattern for interactive trees that support selection and per-row actions. The spec and docs reflect this rather than the bare `tree`/`treeitem`/`group` roles in the ticket text.

## Testing

- `pnpm --filter @commercetools/nimbus typecheck` ✅
- `pnpm exec eslint` on all new/changed files ✅
- `pnpm --filter @commercetools/nimbus build` ✅
- `pnpm test packages/nimbus/src/components/tree/tree.stories.tsx` — 9/9 play functions pass (built bundle) ✅
- `tree.docs.spec.tsx` — 4/4 pass ✅

[...]

PR #1940, opened 2026-08-21
fix(tree): reject moving a node into its own subtree and close Tree coverage gaps

misama-ct:
## Summary

FEC-1168 asks for a Tree component built on React Aria's `Tree` with
drag-and-drop. Nearly all of it is already shipped — the ticket was written
against in-flight branch commits about a month after the work merged, so several
of its acceptance criteria describe states that never reached `main`.

This PR verifies every criterion against `main`, then closes the three real gaps
that verification turned up: one defect, four missing regression guards, and two
inaccurate docs claims.

## Already in place

Tree shipped in `@commercetools/nimbus@3.3.0` via #1574 (FEC-985, Done) as a
net-new component — 19 files, ~2,500 lines. Verified criterion by criterion
against `main`:

| Criterion | Status | Evidence |
|---|---|---|
| Wraps React Aria Tree with Nimbus styling | ✅ In place | Every part delegates: `Tree`→`tree.root.tsx:24`, `TreeItem`→`tree.item.tsx:25`, `TreeItemContent`→`tree.item-content.tsx:37`, `Button slot="chevron"`→`tree.indicator.tsx:26`, `Collection`→`tree.sub-tree.tsx:30`. No hand-rolled collection, selection, focus or type-ahead logic; the five part files contain no React hooks at all. Styling is entirely the 4-slot Chakra recipe. |
| Checkbox selection without collision or misalignment | ✅ In place | Collision is structurally precluded, not merely absent: one flex row, uniform non-shrinking control boxes, no absolute positioning, no mode-conditional offsets. `--tree-indent-step` is *defined* as `control + gap`, which forces level-N's leading control onto level-(N−1)'s chevron column at both sizes. The two historical causes are fixed — the checkbox's label-width reservation is gated behind `:has([data-slot='label'])`, and chevron visibility is class-scoped so it can't hit the checkbox's own `data-slot="indicator"`. |
| Keyboard navigation (arrows expand/collapse, type-ahead) | ✅ In place | Fully delegated — `EXPANSION_KEYS` and `handleTreeExpansionKeys` live in react-aria's `useGridListItem`; `disallowTypeAhead` is never set, and `textValue` is a required prop, which is what type-ahead needs. |
| No forced `overflow:auto` on root | ✅ In place | No `overflow` in any slot, variant or state across the 240-line recipe; `tree.recipe.ts:32-35` records why. |
| Reorder-only story demonstrating DnD without selection | ✅ In place | `ReorderWithoutSelection` — DnD on, `selectionMode="none"`, asserts no checkboxes render. |
| Play tests for expand/collapse, selection, DnD | ✅ In place | Every story carries a play function. `DragAndDrop` performs a complete keyboard drag that re-parents a row and asserts the resulting order — covering the selection-order-vs-tree-order case too. |
| Existing stories updated to new API | ✅ In place | True within the branch: `425139332` migrated them off raw React Aria, swapping `<Collection>` for `<Tree.SubTree>` and a ~25-line inline `useTreeData`+`useDragAndDrop` block for one `useTree()` spread. Tree was net-new, so there were no stories on `main` to update. |
| DnD reordering with **token-based drag-handle optical offset** | ⚠️ Obsolete by design | Reordering and re-parenting are in place. The *offset* is not, and shouldn't be: it was built (`--tree-drag-offset`, a per-size `translateX`) and then deliberately replaced inside the same PR by the uniform leading-control-column layout, which removes the need for per-mode compensation. `--tree-drag-offset` has zero occurrences repo-wide; `tree.recipe.ts:20-27` explains the superseding approach. The criterion is a verbatim copy of the reverted sub-commit's title, and the canonical spec (`openspec/specs/tree/spec.md:148-187`) never mentions an offset. **No action taken — flagging for whoever refines the ticket.** |

### 1. A defect: `move` silently discarded a subtree

`useTree` returns React Stately's mutations as an imperative controller for
programmatic edits. Passing the moved node itself, or a descendant, as the target
parent detached the node and re-attached it under a key that no longer existed —
so the node and everything beneath it were dropped from the tree with no throw,
no rejection, no diagnostic. A 7-node fixture came back with 3.

`moveBefore`/`moveAfter` already threw for this exact condition, so the
re-parent path was the only mutation that lost data instead of reporting. `move`
now rejects it too. The guard is deliberately narrow — moving a node to its own
parent, an ancestor, or the root stays valid, and that's covered.

**Drag-and-drop was never affected.** React Aria's `getDropOperation` cancels any
internal drop targeting a dragged key or a descendant of one, pointer and
keyboard alike. Walking the real keyboard drop targets while dragging
`Documents` confirms only the untouched `Photos` subtree is ever offered.

### 2. Four regression guards for behavior that was already correct

- **Drop-target exclusion.** The guarantee above is load-bearing and lived
  entirely upstream. `DropTargetsExcludeDraggedSubtree` now fails if a custom
  `getDropOperation`, or a React Aria upgrade that relocates the guard, exposes
  the destructive path.
- **`ArrowUp`** was documented but never pressed in any test.
- **Sibling reorder order** — what `onMove` delegates to for before/after drops
  — had no assertion, including the selection-order case `keysInTreeOrder`
  exists for. Asserted at controller level, so it doesn't depend on which
  insertion point a keypress happens to land on.
- **Root `overflow`** was recorded only in a source comment; `FocusedRow` now
  asserts the computed value.

### 3. Two docs claims that were wrong

- Both docs said `Right`/`Left` "expand/collapse, or move to a child/parent".
  Left-to-parent is correct; `Right` never moves to a child — on an expanded row
  React Aria moves *within* the row to its own controls.
- `acceptedDragTypes` was documented as accepting external drag types. No
  external-drop handler is wired up and the move path requires an internal drag,
  so external drops always cancel; the option can only narrow an internal drag.

## Test plan

- [x] `pnpm test:storybook:dev packages/nimbus/src/components/tree/tree.stories.tsx` — 17/17 pass
- [x] `pnpm test:unit packages/nimbus/src/components/tree/` — 11/11 pass
- [x] `pnpm --filter @commercetools/nimbus typecheck:dev` — clean
- [x] `eslint` + `prettier` — clean
- [x] Changeset added (`patch`)

## Feedback wanted

1. **Throw vs. no-op.** I chose throwing, for consistency with
   `moveBefore`/`moveAfter`. A silent no-op would be more forgiving for a
   consumer computing moves from user input. Easy to switch.
2. **The obsolete criterion.** Worth confirming the offset really is dead by
   design before FEC-1168 is refined — the uniform-column layout looks like the
   better approach, so I've left it alone rather than reinstating it.

FEC-1168

### DIFF
--- a/.changeset/tree-move-subtree-guard.md
+++ b/.changeset/tree-move-subtree-guard.md
@@ -1,0 +1,10 @@
+---
+"@commercetools/nimbus": patch
+---
+
+`Tree`: `useTree`'s `move` no longer discards a node when it is moved into its
+own subtree. Passing the moved node itself, or any of its descendants, as the
+target parent previously removed the node and everything beneath it from the
+tree without reporting anything; it now throws, matching what `moveBefore` and
+`moveAfter` already did for the same case. Drag-and-drop was never affected —
+such a drop is refused before it reaches the tree.
--- a/packages/nimbus/src/components/tree/hooks/use-tree.spec.tsx
+++ b/packages/nimbus/src/components/tree/hooks/use-tree.spec.tsx
@@ -1,0 +1,125 @@
+import { describe, it, expect } from "vitest";
+import { act, renderHook } from "@testing-library/react";
+import type { Key } from "react-aria-components";
+import { useTree } from "./use-tree";
+import { fileTree, type TreeNode } from "../utils/tree.test-data";
+
+/**
+ * `useTree` exposes React Stately's tree mutations — including `move` — as an
+ * imperative controller for programmatic edits, alongside the state it spreads
+ * onto `Tree.Root`. Moving a node *into its own subtree* is not a meaningful
+ * operation, and `move` neither rejects nor reports it: the node is detached,
+ * then re-attached under a parent key that no longer exists, so the dragged node
+ * and everything beneath it are dropped from the tree with no error raised.
+ *
+ * Drag-and-drop does not expose this. React Aria's `getDropOperation` cancels
+ * any internal drop targeting a dragged key or a descendant of one, for pointer
+ * and keyboard alike (`react-stately`'s `useDroppableCollectionState`), so
+ * `onMove` is never handed such a target — pinned by the
+ * `DropTargetsExcludeDraggedSubtree` story. The reachable path is a consumer
+ * calling `move` directly.
+ *
+ * The two failing cases below are the defect; the first test is a control
+ * proving the harness itself moves nodes correctly.
+ */
+
+/** Structural view of a `useTreeData` node, enough to walk the hierarchy. */
+interface WalkableNode {
+  key: Key;
+  children?: WalkableNode[] | null;
+}
+
+/** Every key in the tree, in document order. */
+const flattenKeys = (nodes: readonly WalkableNode[]): Key[] =>
+  nodes.flatMap((node) => [node.key, ...flattenKeys(node.children ?? [])]);
+
+/** The 7 keys of `fileTree`, in document order, before any mutation. */
+const ALL_KEYS = [
+  "documents",
+  "project",
+  "report",
+  "budget",
+  "photos",
+  "image-1",
+  "image-2",
+];
+
+const setup = () =>
+  renderHook(() =>
+    useTree<TreeNode>({
+      initialItems: fileTree,
+      getKey: (item) => item.id,
+      getChildren: (item) => item.children ?? [],
+      dragAndDrop: true,
+    })
+  );
+
+/**
+ * Run a tree mutation, capturing a throw rather than failing on it, so each
+ * case can report whether the call threw *and* whether the data survived
+ * instead of stopping at the first symptom.
+ */
+const attempt = (fn: () => void): unknown => {
+  let caught: unknown;
+  act(() => {
+    try {
+      fn();
+    } catch (error) {
+      caught = error;
+    }
+  });
+  return caught;
+};
+
+describe("useTree — move into own subtree", () => {
+  it("moves a node into an unrelated group (control)", () => {
+    const { result } = setup();
+    expect(flattenKeys(result.current.items)).toEqual(ALL_KEYS);
+
+    const error = attempt(() => result.current.move("report", "photos", 2));
+
+    expect(error).toBeUndefined();
+    expect(result.current.getItem("report")).toBeDefined();
+    expect(flattenKeys(result.current.items)).toHaveLength(ALL_KEYS.length);
+    expect(
+      result.current.getItem("photos")?.children?.map((child) => child.key)
+    ).toContain("report");
+  });
+
+  it("keeps the node and its subtree when moved into itself", () => {
+    const { result } = setup();
+    const index = result.current.getItem("documents")?.children?.length ?? 0;
+
+    const error = attempt(() =>
+      result.current.move("documents", "documents", index)
+    );
+
+    expect
+      .soft(error, "move() neither rejected nor reported the no-op")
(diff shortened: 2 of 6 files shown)


misama-ct:
@claude roast this PR

misama-ct:
Dear @claude,

Confirmed and fixed in 1ec9d7229. Your framing was exactly right: `remove` covers its ledger-credited children *incidentally*, by tombstoning the key it was handed, and `update` has no such cover because the container survives.

| batch | before | after |
| --- | --- | --- |
| `append("project",draft)` → `update("project",{children:[report,budget]})` → `move("photos","draft",0)` | no throw, 3 lost | throws, all 7 intact |
| `move("report","photos",0)` → `update("photos",{children:[]})` → `move("documents","report",0)` | no throw, **one node left of seven** | throws |

You were right that the second is the shape a consumer is likelier to hit, and right that it passes the own-subtree check legitimately — the destination check was the only thing standing there and it said yes.

Both departure records now come from one helper collecting the snapshot's view of a container's contents together with whatever the ledger credits to it. Sweep before `recordInserted`, as you said, so a key reappearing in the replacement's children is re-credited.

## Writing it turned up an over-reach in the record `remove` already had

Found by a control, not a repro — which is the argument for writing controls at all:

```
move("report", "photos", 0);   // report escapes documents
update("documents", { children: [project] });
move("image-1", "report", 0);  // refused — but report is reachable under photos
```

The snapshot still shows `report` under `project`, so tombstoning its subtree buried a node that had already left. Filtering candidates by effective containment fixes it.

**And that filter must not be applied to `remove`** — the two disagree about what is still inside, and the disagreement is upstream's. `update` composes with the queue, so an escapee has genuinely escaped. `remove` computes its next state eagerly from the last render and discards the queue, so the relocation never happens and the escapee leaves with the container after all:

```
move("report", "photos", 0);
remove("documents");           // discards the queued move — report goes too
move("image-1", "report", 0);  // must be refused
```

Sparing `report` as an escapee there admits a move onto a node upstream is about to take. So the flag is load-bearing in both directions, and there's a mutation for each: making `remove` respect the queue admits the discarded relocation, making `update` ignore it refuses a reachable escapee.

This is also the first place the 84-of-600 class becomes something a wrapper can be honest about. I can't stop `remove` discarding the queue, but I can stop the guard from being fooled by it. Spec'd and documented rather than inferred, since it's consumer-observable: a relocation issued before a `remove` in the same tick is silently discarded.

## Your two notes

The early-return observation was load-bearing for the fix, not just an aside — "the walk is fine; it's being told the truth about `remove` and not about `update`" is the sentence that told me where to look, and it's why I didn't touch `willExist` this round.

The root-control note about `while (cursor !== undefined)` is a real hole in my test and I've left the control in place for exactly that reason. I didn't add a separate exit-condition test; the root control catches it, which is what you predicted.

Spec clause extended — a key can now leave because something above it was *emptied*, not only removed. As you noted last round, the property is checkable against the code rather than against the cases that motivated it, and this round it was: the clause as written already implied the emptied-container case, and the code didn't do it.

52 unit, 17 story, typecheck, `openspec --strict`, lint all green.

Five rounds, twelve findings, eleven real. Two of the corrections I made to your repros were about which call reaches the bug rather than whether it exists, and one of the holes this round I found myself only because you'd argued for writing controls alongside repros. The unexecuted reasoning has been the more reliable half of this exchange, which I'd not have predicted at the start.

ByronDWall (approved):
Not only do I approve this PR, I took the liberty of opening [an issue in react-aria](https://github.com/adobe/react-spectrum/issues/10539) with the findings.

I really appreciate the work, and the thoroughness.

PR #1941, opened 2026-08-21
fix(bundle): reduce consumer bundle size by 57% (377→163 KB gz)

ByronDWall:
## Summary

Reduces the consumer production bundle size from **377 KB gz to 163 KB gz** (−57%) by fixing icon tree-shaking, inlining internal packages, and eliminating dead token imports.

### 1. Fix nimbus-icons tree-shaking (biggest win)
Nested `dist/esm/package.json` files were missing `sideEffects: false`, shadowing the root declaration. Bundlers included all 2,127 icons even when only ~50 are used. Fixed by adding `sideEffects: false` to the postbuild marker files.

### 2. Bundle nimbus-icons into nimbus (drop peer dep)
Nimbus uses only ~50 icons internally. These are now bundled into nimbus's dist instead of being externalized. Consumers who import icons directly for their own UI install `nimbus-icons` separately — tree-shaking now works correctly for them too.

### 3. Bundle nimbus-tokens into nimbus (drop peer dep)
Token values are consumed at build time to create the Chakra theme. They're now inlined into nimbus's theme chunk. No runtime dependency on `nimbus-tokens`.

### 4. Eliminate designTokens import
Two files (`money-input.tsx`, `multiline-text-input.recipe.ts`) imported the 93 KB `designTokens` module for a single value each. Migrated to `themeTokens` (which is already in the dependency graph via the theme chunk).

### 5. Bundle visualizer for blank-app
Added opt-in `rollup-plugin-visualizer` behind `ANALYZE_BUNDLE` env var. Run `pnpm --filter blank-app build:analyze` to generate `bundle-report.html`.

## Peer dependency changes

| Before | After |
|--------|-------|
| `@commercetools/nimbus-icons` | _(removed — bundled internally)_ |
| `@commercetools/nimbus-tokens` | _(removed — bundled at build time)_ |
| `@chakra-ui/react` | `@chakra-ui/react` |
| `react` | `react` |
| `slate`, `slate-react`, `slate-history`, `slate-hyperscript` | _(unchanged)_ |

Consumers who only had `nimbus-icons` and `nimbus-tokens` installed because nimbus required them can now remove both packages.

## Verification

- ✅ `pnpm build:packages` — builds successfully
- ✅ `pnpm --filter @commercetools/nimbus typecheck:dev` — clean
- ✅ `pnpm test:dev` — 3,225 tests pass (250 files)
- ✅ `pnpm check:package-shape` — all packages pass attw + publint
- ✅ `pnpm --filter blank-app build` — 588 KB raw / 163 KB gz (down from 377 KB gz)
- ✅ No `@commercetools/nimbus-tokens` references in dist output
- ✅ nimbus-icons tree-shaking verified (only ~50 used icons in bundle)

## Future work (not in this PR)

- **Toast code-splitting** (~18 KB gz): `NimbusProvider` unconditionally imports `ToastOutlet`, pulling in `@zag-js/*`. Needs a separate entry point (`@commercetools/nimbus/toast`) — breaking API change requiring OpenSpec proposal.
- **Slate isolation**: Move rich text editor to a separate entry point to make Slate deps optional.
- **Theme chunk optimization** (~38 KB gz): All 70 recipes load via `createSystem()` regardless of which components are used.

Closes FEC-1293

### DIFF
--- a/apps/blank-app/package.json
+++ b/apps/blank-app/package.json
@@ -6,6 +6,7 @@
   "scripts": {
     "dev": "vite",
     "build": "tsc -b && vite build",
+    "build:analyze": "ANALYZE_BUNDLE=true tsc -b && ANALYZE_BUNDLE=true vite build",
     "build:webpack": "webpack --mode production",
     "lint": "eslint .",
     "preview": "vite preview"
@@ -26,6 +27,7 @@
     "globals": "catalog:tooling",
     "typescript": "catalog:tooling",
     "typescript-eslint": "catalog:tooling",
+    "rollup-plugin-visualizer": "^5.14.0",
     "vite": "catalog:tooling",
     "webpack": "catalog:tooling",
     "webpack-cli": "catalog:tooling"
--- a/.changeset/bundle-optimization-icons-tokens.md
+++ b/.changeset/bundle-optimization-icons-tokens.md
@@ -1,0 +1,18 @@
+---
+"@commercetools/nimbus": minor
+"@commercetools/nimbus-icons": patch
+---
+
+`@commercetools/nimbus-icons` and `@commercetools/nimbus-tokens` are no longer
+required peer dependencies. Nimbus now bundles the icons and tokens it uses
+internally, so consumers only need to install `@chakra-ui/react` and `react` as
+peers.
+
+- Consumers who import icons directly from `@commercetools/nimbus-icons` for
+  their own UI should keep it as a regular dependency — it continues to work
+  independently and now tree-shakes correctly.
+- Consumers who only had `nimbus-icons` and `nimbus-tokens` installed because
+  nimbus required them can remove both packages.
+
+`@commercetools/nimbus-icons`: fixed tree-shaking — bundlers now correctly
+eliminate unused icons instead of including all 2,000+.
--- a/apps/blank-app/.gitignore
+++ b/apps/blank-app/.gitignore
@@ -9,6 +9,7 @@ lerna-debug.log*
 
 node_modules
 dist
+bundle-report.html
 dist-ssr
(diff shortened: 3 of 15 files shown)


ByronDWall:
> Package 	Format 	Current 	Baseline 	Delta 	Status
> @commercetools/nimbus 	dist 	19762.0 KB 	18950.9 KB 	+4.3% 	✅ ok

This bump is because we are now bundling icons and tokens directly instead of requiring peer deps, the up side is that now we aren't importing the icons or tokens packages directly, so large chunks of them are tree-shaken out leading to a leaner overall bundle for the consumer.

apps/blank-app/package.json
```diff
@@ -26,6 +27,7 @@
     "globals": "catalog:tooling",
     "typescript": "catalog:tooling",
     "typescript-eslint": "catalog:tooling",
+    "rollup-plugin-visualizer": "^5.14.0",
```
misama-ct: Anything speaking against adding this to the tooling group like the other deps?

packages/nimbus/src/components/toast/components/toast.outlet.lazy.tsx
misama-ct, line 1: Is this something we could make a convention out of? Maybe even use `.lazy.` instead of `.impl.`.

apps/blank-app/package.json
ByronDWall: there sure isn't, good catch

packages/nimbus/src/components/toast/components/toast.outlet.lazy.tsx
ByronDWall: I like this suggestion and will implement it

